PR #133, opened 2021-04-18
WIP: Add ability to connect to a Gnosis Safe as a Safe app

TomAFrench:
~~Draft PR as this depends on https://github.com/snapshot-labs/lock/pull/13 being merged/published.~~

This PR adds a "Gnosis Safe" wallet option if the app is loaded within an iframe which will attempt to connect to a Safe using the [Safe Apps SDK](https://github.com/gnosis/safe-apps-sdk). This allows the app to be used on https://gnosis-safe.io without needing to maintain a separate version.


TomAFrench:
Due to some quirks of Gnosis' Safe provider, notify.js does not correctly report when a transaction has confirmed for 1 of n multisigs leading to us getting stuck in an infinite loading scenario. These issues do not occur for m of n (m > 1).

Putting this back into WIP as the issues with notify.js are either going to need updates on Gnosis' side or for us to handle the translation between safeTxHashes and the real transaction hashes.

TomAFrench:
Closing this as we've gone through some changes to the wallet code so this is out of date. Much of the changes were made in other packages so we can  revive this relatively quickly if needed.

garethfuller:
@TomAFrench Sorry this didn't get merged at the time. It was right around V2 launch and I didn't have the time or awareness to review properly. If you want to revive it with the latest code then feel free.

PR #310, opened 2021-05-10
John/UI 415 0x renders with a times symbol if

johngrantuk:
Fixes # .

- Add `font-variant-ligatures: no-contextual;` to addresses to prevent auto multiplaction symbol formatting (https://github.com/rsms/inter/issues/222)
- 
-


TomAFrench:
I've noticed this issue popping up again for token addresses where they're not included in our tokenlists.

> I don't think we want that type of text transformation anywhere right? If so, lets just set this as a global css property here: https://github.com/balancer-labs/frontend-v2/blob/develop/src/assets/css/global/all.css#L5-L9

@garethfuller did you come to a decision not to go down this route or should I make that change?

garethfuller:
@TomAFrench so the solution from this PR was only applied in the place it was seen, not globally. What we could do is add a class to `assets/css/global/typography.css` like:
```css
.eth-address {
  font-variant-ligatures: no-contextual;
}
```
And then wherever you are seeing this problem we can just apply this class.

PR #342, opened 2021-05-13
update: table stickiness and width calculation

1saf:
Fixes table issues on mobile

**Testing Criteria**
Make sure tables work as intended on desktop - should be equivalent behaviour to the production app.
Make sure that tables work well on mobile devices, are there any artefacts or weird behaviour?
Try to test different screen sizes.

- Adjust table width to be a flat number provided in the column config, this seems the way to go to avoid most dramas I've observed with other table implementations and libraries
- The table is comprised of a header table and a body table - the header scroll is controlled by JS as the body scrolls - this is to avoid the sliding issue if the table header is made to be sticky. As a sticky element nested within another sticky element just doesn't work.
- Moved the loading and no results block out of the table tag as they will be affected by colgroup.


garethfuller (approved):
Table used to be like this:
Screenshot 2021-05-13 at 10 21 27
Now it's like this, notice the scroll bar placeholder.
Screenshot 2021-05-13 at 10 21 35
Wondering if we should try to avoid that for non fixed height tables?

On mobile it's looking mint!

PR #343, opened 2021-05-13
Load modals conditionally

evgenyboxer:
By loading the modals with `v-if` we are not doing any computation for for these modals (since they are hidden at first).
Another benefit is the cleanup logic - not needed anymore because its destroyed and created each time.

Hopefully this can speed up load time by not rendering components that are initially hidden.


src/composables/useTokenLists.ts
```diff
@@ -75,13 +75,16 @@ export default function useTokenLists(request?: TokenListRequest) {
   const queryKey = QUERY_KEYS.TokenLists;
   const queryFn = loadAllTokenLists;
 
-  const { data: lists, isLoading } = useQuery<TokenList[]>(queryKey, queryFn);
+  const { data: lists, isLoading } = useQuery<TokenList[]>(queryKey, queryFn, {
+    refetchOnMount: false,
+    refetchOnWindowFocus: false
```
evgenyboxer, line 80: To further optimize loading, I was thinking its best to avoid refetching when a component mounts / window refocus for this specific query.

Token lists don't change that often and I think we could think of a better mechanism to update them throughout the user session.

src/composables/useTokenLists.ts
```diff
@@ -75,13 +75,16 @@ export default function useTokenLists(request?: TokenListRequest) {
...
+  });
 
   const listDictionary = computed(() => keyBy(lists.value, 'name'));
   const injectedTokens = store.getters['registry/getInjected'];
 
   const tokens = computed(() => {
-    const _tokens = uniqBy(
+    const _tokens = uniqBy<TokenListItem>(
```
evgenyboxer, line 87: Added some missing types here

src/components/modals/AccountModal.vue
```diff
@@ -92,15 +92,6 @@ export default defineComponent({
       );
     });
 
-    // WATCHERS
-    watch(
-      () => props.open,
-      () => {
-        data.step = null;
-        data.connectorsLimit = 3;
```
1saf, line 100: is this important?

src/composables/useTokenLists.ts
1saf: thanks for this

src/components/modals/AccountModal.vue
evgenyboxer: This is the cleanup logic - its just trying to restore it to the initial state which is:
```
    // DATA
    const data = reactive({
      connectorsLimit: 3,
      step: null,
      path:
        'https://raw.githubusercontent.com/snapshot-labs/lock/master/connectors/assets'
    });
```

1saf (approved):

garethfuller (approved):
Nice work!

PR #346, opened 2021-05-13
Handle pool loading error

garethfuller:
If the poolQuery fails we should redirect back to pools list. Ideally we should show a notification that the pool failed to load, but before we implement that we should get Pon to design some notifications (where they should appear and how they should look).


evgenyboxer:
Yeah it would be nice to have some kind of a loading error notification with a retry button... guess its ok for now (to redirect)

evgenyboxer (approved):

PR #348, opened 2021-05-14
Persist pool filters

evgenyboxer:
Persist pool filters (user session) for the home (invest page) and individual pool page.

It resets when the user navigates to the Trade page.


src/components/inputs/TokenSearchInput.vue
```diff
@@ -45,7 +45,7 @@
         </span>
       </div>
       <div v-else class="text-gray-400 flex flex-wrap">
-        <span class="mr-6">Popular Bases:</span>
+        <span class="mr-6">{{ $t('popularBases') }}</span>
```
evgenyboxer, line 48: ^ some missing translations in this file

src/store/modules/app.ts
```diff
@@ -83,9 +85,16 @@ const mutations = {
...
+  setTradeLiquidity(state: AppState, tradeLiquidity: LiquiditySelection) {
     state.tradeLiquidity = tradeLiquidity;
     lsSet('tradeLiquidity', state.tradeLiquidity);
+  },
+
+  setSelectedPoolTokens(
+    state: AppState,
+    selectedPoolTokens: AppState['selectedPoolTokens']
+  ) {
+    state.selectedPoolTokens = selectedPoolTokens;
```
evgenyboxer, line 97: I thought its better to just reuse the already defined type in `AppState` for all the params (also removed `void` - since its already inferred in TS)

garethfuller (approved):

PR #349, opened 2021-05-14
fix: add guards to undefined keys being accessed

1saf:
Fixes # .
https://sentry.io/organizations/balancer-labs/issues/2393335684/?project=5725878&query=is%3Aunresolved
https://sentry.io/organizations/balancer-labs/issues/2393284015/?project=5725878&query=is%3Aunresolved&sort=freq&statsPeriod=14d

Didn't manage to re-create the issues in local somehow, so I've added guards to the potentially undefined keys being accessed.

### DIFF
--- a/src/components/cards/TradeCard/TradeRoute.vue
+++ b/src/components/cards/TradeCard/TradeRoute.vue
@@ -77,7 +77,7 @@
           <div class="relative z-10">
             <div
               v-for="route in routes"
-              :key="route.hops[0].pool.address"
+              :key="route.hops[0]?.pool?.address"
               class="mt-9 first:mt-0 flex justify-between"
             >
               <div class="w-4 ml-4 flex items-center">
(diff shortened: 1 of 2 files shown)


src/components/cards/TradeCard/TradeRoute.vue
```diff
@@ -393,7 +393,8 @@ export default defineComponent({
       routes,
 
       formatShare,
-      getPoolLink
+      getPoolLink,
+      console
```
evgenyboxer, line 397: Do we need console here?

evgenyboxer (approved):

garethfuller (changes requested):
@1saf this is worth doing as a hotfix. Instead of mixing it in with a bunch of other changes in the next release we'll be better able to see if it fixes any problems if we push it as a standalone release.

Could you branch off master and make these same changes, then push another PR?

PR #358, opened 2021-05-16
Show all swap fee possibilities (Gauntlet, fixed, owner)

EndymionJkb:
- There are three swap fee options - show all of them
- owner-controlled: 0x03cd191f589d12b0582a99808cf19851e468e6b500020000000000000000002b
- truly fixed: 0xf461f2240b66d55dcf9059e26c022160c06863bf000100000000000000000007


src/constants/pools.ts
```diff
@@ -34,7 +36,8 @@ export const POOLS = {
       '0x5c6ee304399dbdb9c8ef030ab642b10820db8f56000200000000000000000014',
       '0xf4c0dd9b82da36c07605df83c8a416f11724d88b000200000000000000000026',
       '0xf8a0623ab66f985effc1c69d05f1af4badb01b0000020000000000000000001d',
-      '0x021c343c6180f03ce9e48fae3ff432309b9af19900020000000000000000000b'
+      '0x021c343c6180f03ce9e48fae3ff432309b9af19900020000000000000000000b',
+      '0x87165b659ba7746907a48763063efa3b323c2b0700020000000000000000002d'
```
markusbkoch, line 40: This list is meant to contain only pools that are actively managed by Gauntlet

EndymionJkb: Am I wrong that the current state is it manages all of them with that owner, but the new Authorizer will make it fine-grained?

markusbkoch: They can manage all but will only manage the ones in the original list

EndymionJkb: Ok then; restoring original list

src/services/balancer/contracts/contracts/vault.ts
```diff
@@ -30,6 +30,7 @@ export default class Vault {
     result = await this.multiCaller.execute(result);
     this.service.multiCaller.call('totalSupply', poolAddress, 'totalSupply');
     this.service.multiCaller.call('decimals', poolAddress, 'decimals');
+    this.service.multiCaller.call('owner', poolAddress, 'getOwner');
```
garethfuller, line 33: Is the owner likely to change often? Can we instead get this from the subgraph? Reason being, ideally we only fetch the pool attributes that need to be dynamic with onchain calls.

src/pages/Pool.vue
```diff
@@ -159,13 +155,32 @@ export default defineComponent({
...
+      () => pool.value?.onchain.owner == POOLS.ZeroAddress
+    );
+    const swapFeeToolTip = computed(() =>
+      feesManagedByGauntlet.value
+        ? t('feesManagedByGauntlet')
+        : feesFixed.value
+        ? t('fixedFeesTooltip')
+        : communityManagedFees.value
+        ? t('delegateFeesTooltip')
+        : t('ownerFeesTooltip')
```
garethfuller, line 181: Using the ternary operator is quite unreadable here. 
```suggestion
      if (feesManagedByGauntlet.value) {
        t('feesManagedByGauntlet')
      } else if (communityManagedFees.value)
        t('delegateFeesTooltip')
      } else if (feesFixed.value)
        t('fixedFeesTooltip')
      } else {
        t('ownerFeesTooltip')
      }
```

garethfuller (changes requested):
Thanks for this Jeff. Just some minor comments. Also, let's try to keep PRs single purpose.

EndymionJkb:
The owner can't change (with core pools, anyway). Don't think it's on the
Subgraph but will check. I also want to differentiate between oracle and
non-oracle weighted pools (and show the state of the oracle). That is a
subgraph change as well.

Meant to be single purpose, but running my script to populate the new tx
keys pulled in all the other new keys as well. Sorry! Will do a tx-only one
first next time.

src/services/balancer/contracts/contracts/vault.ts
EndymionJkb: It was in the subgraph after all. (And is immutable for core pools.) So now I get it from there rather than on-chain.

garethfuller (approved):

PR #360, opened 2021-05-17
Allow Baltable to support link behaviour on rows (i.e. open in new tab)

1saf:
- Allow Baltable to support link behaviour on rows (i.e. open in new tab)

**Testing Criteria**
- Pools table has been updated to use link behaviour. Make sure that you can open the pool in a new tab, can you use the browser back button too?
- Styles should remain the same across mobile and desktop, is anything different because of this new wrapped element?
- Test open in new tab on mobile too


src/components/_global/BalTable/BalTable.vue
```diff
@@ -106,31 +106,61 @@
...
+                    column.align === 'right' ? 'text-right' : 'text-left'
+                  ]"
+                >
+                  {{
+                    typeof column.accessor === 'string'
+                      ? dataItem[column.accessor]
+                      : column.accessor(dataItem)
+                  }}
+                </div>
+              </template>
```
1saf, line 163: I couldn't find an elegant way to have link and non-link behaviour without duplication - 
I tried using the v-slot behaviour for custom router-links but that didn't work.
This is because for non-linkable rows we don't want users to see that you can 'right click -> open url options' in the browser but for link ones too.

Do you guys have any better ideas?

src/components/_global/BalTable/BalTable.vue
```diff
@@ -106,31 +106,61 @@
           >
             <td
               v-for="(column, columnIndex) in filteredColumns"
+              @click="navigate"
+              :href="href"
```
garethfuller, line 110: Where are `navigate` and `href` coming from? I don't see any reference to them throughout the component? Also, why is `href` on a td tag?

src/components/_global/BalTable/BalTable.vue
garethfuller: Can we not maybe use `<component :is="tag">...` in a similar way to how we use it for `BalBtn`?

1saf: Ill try that

src/components/_global/BalTable/BalTable.vue
1saf: Forgot to commit the change on this, will update in a moment

src/components/_global/BalTable/BalTable.vue
1saf: The recommended way to change the HTML element for router-links is to use the `tag` prop but that does not seem to work.

garethfuller: Does this help: https://next.router.vuejs.org/guide/advanced/extending-router-link.html#extending-routerlink
Note make sure you're viewing vue-router docs @4x. A lot of google links are still pointing to @3x.

garethfuller: And this: https://next.router.vuejs.org/api/#router-link-props

1saf: Had the same problems with the new API as well. The `custom` prop you saw earlier was from the `router-link` v-slot preperation.

markusbkoch:
The vercel preview exhibits a weird behavior on my tests: the link opens in the same tab, but a new tab opens with the list of pools

1saf:
> The vercel preview exhibits a weird behavior on my tests: the link opens in the same tab, but a new tab opens with the list of pools

Resolved :D

PR #364, opened 2021-05-17
When changing the account via metamask, the navbar does not reflect this change

1saf:
- Commit the web3 loading state to true when changing chain or account so dependents can change without undefined errors

**Testing Criteria**
- Change between accounts and chains on metamask, the navbar should update to reflect changes (does the address change?)

### DIFF
--- a/src/store/modules/web3.ts
+++ b/src/store/modules/web3.ts
@@ -83,20 +83,26 @@ const actions = {
         auth.provider.value.removeAllListeners();
       if (auth.provider.value.on) {
         auth.provider.value.on('chainChanged', async chainId => {
+          commit('setLoading', true);
+
           await dispatch('account/resetAccount', null, { root: true });
           auth.web3 = new Web3Provider(auth.provider.value);
+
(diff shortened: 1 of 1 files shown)


src/store/modules/web3.ts
```diff
@@ -83,20 +83,26 @@ const actions = {
...
+          commit('setLoading', false);
         });
         auth.provider.value.on('accountsChanged', async accounts => {
           if (accounts.length !== 0) {
+            commit('setLoading', true);
+
             await dispatch('account/resetAccount', null, { root: true });
             auth.web3 = new Web3Provider(auth.provider.value);
             commit('setAccount', accounts[0]);
-            await dispatch('loadProvider');
```
garethfuller, line 97: You're right to try and remove this, but are you sure it won't have any unintended consequences?

garethfuller (approved):

src/store/modules/web3.ts
1saf: It will simply re-register everything, there will be none

PR #367, opened 2021-05-18
Fix various TS types

evgenyboxer:
Fix various types

### DIFF
--- a/src/services/balancer/contracts/contracts/vault.ts
+++ b/src/services/balancer/contracts/contracts/vault.ts
@@ -4,7 +4,7 @@ import { Multicaller } from '@/lib/utils/balancer/contract';
 import { getAddress } from '@ethersproject/address';
 import { formatUnits } from '@ethersproject/units';
 import { BigNumber } from '@ethersproject/bignumber';
-import { Token } from '@/types';
+import { TokenMap } from '@/types';
 import { OnchainPoolData, PoolType } from '../../subgraph/types';
 
 const NETWORK = process.env.VUE_APP_NETWORK || '1';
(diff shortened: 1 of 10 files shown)


src/services/balancer/contracts/contracts/vault.ts
```diff
@@ -96,7 +100,8 @@ export default class Vault {
           100
       );
     } else if (type === 'Stable') {
-      return tokens.map(() => 100 / tokens.length);
+      const tokensList = Object.values(tokens);
+      return tokensList.map(() => 100 / tokensList.length);
```
evgenyboxer, line 104: @garethfuller is this the intended behaviour here?

src/services/pool/calculator/index.ts
```diff
@@ -81,7 +84,7 @@ export default class Calculator {
       amounts.send.forEach((amount, amountIndex) => {
         const greaterThanBalance =
           Number(amount) >
-          this.allTokens[this.tokenOf(type, amountIndex)].balance;
+          Number(this.allTokens[this.tokenOf(type, amountIndex)].balance);
```
evgenyboxer, line 87: This coercion in JavaScript would work fine, but TypeScript doesn't like it... so gotta convert it to a number first before doing the math operation.

src/types/index.ts
```diff
@@ -8,7 +8,7 @@ export interface FormRef {
 
 export interface Token {
   address: string;
-  balance: number;
+  balance: string;
```
evgenyboxer, line 11: Have console logged in - and its a string

src/services/balancer/contracts/contracts/vault.ts
```diff
@@ -96,7 +100,8 @@ export default class Vault {
           100
       );
     } else if (type === 'Stable') {
-      return tokens.map(() => 100 / tokens.length);
```
evgenyboxer, line 99: Tokens was never a an array, it was actually a map.

garethfuller (approved):

PR #368, opened 2021-05-18
Liquidity Mining V2 Tooltip

evgenyboxer:
Adds a tooltip for liquidity mining info for the pools / individual pool pages.


src/lib/utils/liquidityMining/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  BALPrice: string,
+  totalLiquidity: string
+) {
+  return bnum(rewards)
+    .div(7)
+    .times(BALPrice)
+    .times(365)
+    .div(totalLiquidity)
+    .div(100)
+    .toNumber();
```
evgenyboxer, line 44: @markusbkoch does this look right?

src/lib/utils/liquidityMining/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  start_timestamp: number;
+  end_timestamp: number;
+  tiers: Record<LiquidityMiningTierLevel, LiquidityMiningTier>;
+};
+
+type LiquidityMiningRewards = Record<string, number>;
+
+// Liquidity mining started on June 1, 2020
+function getCurrentLiquidityMiningWeek() {
+  return differenceInWeeks(new Date(), new Date(2020, 5, 1)) + 1;
```
evgenyboxer, line 26: @markusbkoch It seems I need to always add `1` to the diff here right? because `differenceInWeeks(new Date(), new Date(2020, 5, 1))` currently gives me `50`.

src/components/tables/PoolsTable.vue
```diff
@@ -51,6 +51,51 @@
...
       </template>
+      <template v-slot:apyCell="pool">
+        <div class="px-6 py-4 -mt-1 flex justify-end">
+          {{
+            Number(pool.dynamic.apy) > 10000
+              ? '-'
+              : fNum(
+                  Number(pool.dynamic.apy) + pool.liquidityMiningAPY,
+                  'percent'
+                )
```
garethfuller, line 62: Let's move this into a computed property to keep the template a bit cleaner.

src/components/tables/PoolsTable.vue
```diff
@@ -51,6 +51,51 @@
...
+            </template>
+            <div class="text-sm">
+              <div class="mb-2">
+                <div>Total APY</div>
+                <div>
+                  {{
+                    fNum(
+                      Number(pool.dynamic.apy) + pool.liquidityMiningAPY,
+                      'percent'
+                    )
```
garethfuller, line 79: Again, this could be moved to the js with the above.

src/composables/queries/usePoolsQuery.ts
```diff
@@ -61,8 +67,28 @@ export default function usePoolsQuery(
       await store.dispatch('registry/injectTokens', uninjectedTokens);
     }
 
+    const poolsWithLiquidityMining = pools.map(pool => {
```
garethfuller, line 70: I would suggest that this should actually be part of the serialisation of the subgraph request here: https://github.com/balancer-labs/frontend-v2/blob/develop/src/services/balancer/subgraph/entities/pools/index.ts#L47-L64

So that liquidityMiningAPY is just an additional attribute of the decorated Pool type.

garethfuller (commented):
And someone will be required to update `LiquidityMiningV2.json` weekly?

src/lib/utils/liquidityMining/index.ts
markusbkoch: I was checking the numbers on the UI and they all seem to be 1/100 of what they should. I think you need to remove line 43 here.

src/lib/utils/liquidityMining/index.ts
markusbkoch: Correct

src/services/balancer/subgraph/entities/pools/index.ts
```diff
@@ -10,6 +10,13 @@ import {
 } from '../../types';
 import { Prices } from '@/services/coingecko';
 import { getAddress } from '@ethersproject/address';
+import {
+  currentLiquidityMiningRewards,
+  computeAPYForPool
+} from '@/lib/utils/liquidityMining';
+import { TOKENS } from '@/constants/tokens';
+
+const IS_LIQUIDITY_MINING_ENABLED = true;
```
evgenyboxer, line 19: @garethfuller once the PR is reviewed and approved - I'll set it to this value to `false` so that it won't block any releases until we have the correct JSON file in place.

evgenyboxer:
@markusbkoch @garethfuller its now ready for review (have addressed the previous code review comments)

evgenyboxer:
Update: just noticed the tooltip has extra padding... ignore this.

src/components/tooltips/LiquidityMiningTooltip.vue
```diff
@@ -0,0 +1,66 @@
...
+
+    return {
+      fNum
+    };
+  }
+});
+</script>
+<style>
+.liquidity-mining-tooltip {
+  @apply p-0;
```
evgenyboxer, line 64: @garethfuller any ideas why this is not enough to override the padding? (needed to make the divider)

src/services/balancer/subgraph/entities/pools/index.ts
```diff
@@ -56,10 +63,41 @@ export default class Pools {
...
+        ? !!liquidityMiningRewards
+        : false;
+
+      if (hasLiquidityMiningRewards) {
+        liquidityMiningAPY = computeAPYForPool(
+          liquidityMiningRewards,
+          prices[TOKENS.AddressMap.BAL].price || 0,
+          pool.totalLiquidity
+        );
+      }
```
garethfuller, line 81: I'd move this into it's own function like `calcAPY` to help keep this serialize function as simple as possible to understand.

src/components/tooltips/LiquidityMiningTooltip.vue
garethfuller: For whatever reason this class is obviously not overriding the default `p-3` set in the BalTooltip component. I would take the opportunity to make the BalTooltip component a bit more flexible and add a boolean `noPad` prop like we have for the BalCard, if it's set to true then we don't apply this `p-3` class.

src/lib/utils/liquidityMining/LiquidityMiningV2.json
```diff
@@ -0,0 +1,27 @@
+{
+    "week_51": {
+      "start_timestamp": 1621220400,
```
markusbkoch, line 3: Although I initially suggested the `start_timestamp` and `end_timestamp` keys, I'd like to propose we remove them. They are not used in the frontend and I don't expect we'll have a need for them in the off-chain scripts either.

src/locales/default.json
```diff
@@ -97,9 +97,11 @@
     "learnMore": "Learn more",
     "light": "Light",
     "loading": "Loading...",
+    "liquidityMiningAPY": "Liquidity Mining APY",
     "loadMore": "Load more",
     "logOut": "Log out",
     "lpsEarnFee": "<span>LPs earn <b>{0}</b> in fees</span>",
+    "lpFeeAPY": "LP fees APY",
```
markusbkoch, line 104: We should change this to "Swap fees APY" - we use "swap fees" everywhere else in the app. Already cleared this with Pon.

evgenyboxer:
Merged after verifying with @markusbkoch that numbers are ok.

PR #370, opened 2021-05-18
[hotfix] require-valid-default-prop compilation

Ulydev:
Fixes the following compilation error:
> frontend-v2/src/components/_global/BalTable/BalTable.vue
>  237:16  error  Type of the default value for 'initialState' prop must be a function  vue/require-valid-default-prop


evgenyboxer (approved):
Thanks for this!

PR #371, opened 2021-05-18
WIP: Ensure that we pull token icons from larger TokenLists

TomAFrench:
(WIP as the new icons aren't picked up until we re-open the token menu atm)

Zerion and Coingecko both don't use checksummed addresses in their tokenlists and this is causing issues when trying to search for token icons as we miss these entries and fall back to checking trustwallet. This is an issue as the tokens which aren't listed on trust wallet are the exact tokens which we rely on these two lists for.

I've updated `loadTokenlist` to automatically convert all addresses to checksum addresses to avoid this.


TomAFrench:
Closing as this need to be part of a larger refactor of the registry store

PR #374, opened 2021-05-20
BAL Claim button popover

evgenyboxer:
Refactor BAL Claim button

This PR is mostly done, just need to tweak the claim logic a bit.


src/services/claim.ts
```diff
@@ -0,0 +1,182 @@
...
+        const lastEntry = liquidityProviders[liquidityProviders.length - 1];
+
+        return {
+          rewards: liquidityProviders
+            .reduce(
+              (total, { current_estimate }) => total.plus(current_estimate),
+              bnum(0)
+            )
+            .toString(),
+          velocity: lastEntry.velocity,
```
markusbkoch, line 140: We can't count on the lastEntry being the current week, but we can count on one of the entries always having a `velocity` of zero. So IIUC we can sum the `velocity`s like you did with the `current_estimate`s

src/components/navs/AppNav/AppNavClaimBtn.vue
```diff
@@ -0,0 +1,201 @@
...
+
+    useIntervalFn(async () => {
+      if (userClaims.value != null && userClaims.value.currentRewardsEstimate) {
+        const diffInSeconds = differenceInSeconds(
+          new Date(),
+          new Date(userClaims.value.currentRewardsEstimate.timestamp)
+        );
+        rewardsEstimateSinceTimestamp.value = bnum(diffInSeconds)
+          .times(userClaims.value.currentRewardsEstimate.velocity)
+          .toString();
```
markusbkoch, line 148: Can we make the number of decimal places constant at eg 4? It looks weird when the estimate goes from eg `1.9999` to `2` and the bubble decreases in width by a lot, only to go back to the original width one second later at `2.0001`

src/components/navs/AppNav/AppNavClaimBtn.vue
```diff
@@ -0,0 +1,201 @@
...
+      userClaims.value != null
+        ? bnum(userClaims.value.totalRewards)
+            .plus(rewardsEstimateSinceTimestamp.value)
+            .toString()
+        : null
+    );
+
+    useIntervalFn(async () => {
+      if (userClaims.value != null && userClaims.value.currentRewardsEstimate) {
+        const diffInSeconds = differenceInSeconds(
```
markusbkoch, line 142: How about using `differenceInMilliseconds` instead and a higher frequency (eg updates every 30 milliseconds) for a more fluid look? Or would that be too memory intensive?

src/components/navs/AppNav/AppNavClaimBtn.vue
```diff
@@ -0,0 +1,203 @@
...
+          const tx = await claimRewards(
+            appNetwork.id,
+            auth.web3,
+            account.value,
+            userClaims.value.pendingClaims,
+            userClaims.value.pendingClaimsReports
+          );
+          txListener(tx.hash);
+
+          await tx.wait();
```
evgenyboxer, line 175: Blocknative updates too fast... I need to actually verify with Infura that the transaction was mined so that `userClaimsQuery` would actually return `0` when claimed.

1saf (approved):

1saf:
Lgtm, might need @garethfuller to weigh in on it too

src/services/claim.ts
```diff
@@ -0,0 +1,183 @@
...
+  try {
+    const claims = pendingClaims.map(week => {
+      const claimBalance = week.amount;
+      const merkleTree = loadTree(reports[week.id]);
+
+      const proof = merkleTree.getHexProof(
+        soliditySha3(account, toWei(claimBalance))
+      );
+      return [parseInt(week.id), toWei(claimBalance), proof];
+    });
```
evgenyboxer, line 170: This function is pretty slow and can even cause the browser to freeze if you have a lot to claim... this is a pretty heavy computation... I do have some experience with web workers which might be the solution in this case. (it will definitely help the UI/UX and not "freeze" the browser)

1saf: I think it would be better to implement it under a web worker then right?

src/components/_global/BalSelectInput/BalSelectInput.vue
```diff
@@ -99,7 +99,7 @@ export default defineComponent({
         case 'lg':
           return 'h-16';
         default:
-          return 'h-14';
+          return 'h-12';
```
evgenyboxer, line 102: seems like this is the default now (from Pon design)

src/composables/useBreakpoints.ts
```diff
@@ -17,5 +17,9 @@ export default function useBreakpoints() {
 
   const width = computed(() => windowWidth.value);
 
-  return { width, bp };
+  const upToLargeBreakpoint = computed(() =>
+    ['sm', 'md', 'lg'].includes(bp.value)
```
evgenyboxer, line 21: this was used quite often, was thinking having some handy shortcuts

src/composables/useNotify.ts
```diff
@@ -30,9 +26,9 @@ export default function useNotify() {
       TransactionEventCode,
       TxCallback | undefined
     >> = {
-      txConfirmed: onTxConfirmed,
-      txCancel: onTxCancel,
-      txFailed: onTxFailed
+      txConfirmed: callbacks?.onTxConfirmed,
+      txCancel: callbacks?.onTxCancel,
+      txFailed: callbacks?.onTxFailed
```
evgenyboxer, line 31: this is just so I don't have to pass the second argument

src/services/claim.ts
1saf: I think it would be better to instil a better trust and confidence in the UI if we don't have issues with lag or freezes from the get go

src/composables/useNumbers.ts
```diff
@@ -5,6 +5,7 @@ import { useStore } from 'vuex';
 export type Preset =
   | 'default'
   | 'token'
+  | 'token_4_decimals'
```
evgenyboxer, line 8: wasn't sure if this is the best name for this format... essentially a "fixed" 4 decimal format.

src/components/navs/AppNav/AppNavClaimBtn.vue
evgenyboxer: Yep! fixed.

src/components/navs/AppNav/AppNavClaimBtn.vue
evgenyboxer: I actually tried it - but it didn't look great. I think I can definitely improve on this... but perhaps in a later PR?

src/components/_global/BalBtn/BalBtn.vue
```diff
@@ -51,6 +51,7 @@ export default defineComponent({
           'primary',
           'gradient',
           'gradient-reverse',
+          'gradient-pink-yellow',
```
evgenyboxer, line 54: new gradient color

src/components/navs/AppNav/AppNavClaimBtn.vue
```diff
@@ -0,0 +1,203 @@
+<template>
+  <BalPopover no-pad v-if="userClaims != null">
```
evgenyboxer, line 2: Currently there is no "loading state" for the claim button, so I'm just hiding it until I can show it. Pinged Pon about this.

src/locales/default.json
```diff
@@ -160,6 +163,7 @@
     "ownerFeesTooltip": "Liquidity providers earn dynamic swap fees on every trade utilizing the Liquidity in this pool. Dynamic swap fees are controlled by the pool owner.",
     "payload": "Payload",
     "pendingClaims": "Pending claims",
+    "pendingCurrentWeek": "Pending: current week (estimate)",
```
markusbkoch, line 166: I suggest we use `Pending (estimate)` as the label, since the number will include both the current week and the previous one until the previous week results have been made official. So:
```
    "pendingCurrentWeek": "Pending (estimate)",
```

src/composables/useNumbers.ts
garethfuller:
```suggestion
  | 'token_fixed'
```

src/services/claim.ts
garethfuller: If we have time to try a web worker for this, then lets at least experiment with it. Or we could try a serverless function somewhere, but that complicates decentralising.

garethfuller (changes requested):

garethfuller (approved):

PR #378, opened 2021-05-22
Liquidity Mining - Switch to using UTC time for time diff

evgenyboxer:
As discussed with @markusbkoch its better to switch to UTC time when computing the diff.


src/lib/utils/liquidityMining/index.ts
```diff
@@ -46,7 +50,7 @@ export function getLiquidityMiningRewards(
 ) {
   const miningWeek =
     week === 'current' ? getCurrentLiquidityMiningWeek() : week;
-
+  console.log(miningWeek);
```
1saf, line 53: leave this log here?

1saf (approved):

PR #387, opened 2021-05-25
Error states for trade UI

Destiner:
Adds `TradeError` component to show errors instead of using button. Also, replaces "high price impact" checkbox.


src/composables/trade/useValidation.ts
```diff
@@ -1,6 +1,17 @@
...
+const MIN_ETH_REQUIRED = 0.0001;
+
+export enum TradeValidation {
+  VALID,
+  NO_ACCOUNT,
+  EMPTY,
+  NO_ETHER,
+  NO_BALANCE,
+  NO_LIQUIDITY
+}
```
Destiner, line 14: Added enum for better separation of concerns (composable doesn't have to know about localization) and type safety.

src/composables/trade/useValidation.ts
```diff
@@ -12,7 +23,7 @@ export default function useValidation(
   const auth = useAuth();
 
   const validationStatus = computed(() => {
-    if (!auth.isAuthenticated) return '';
+    if (!auth.isAuthenticated.value) return TradeValidation.NO_ACCOUNT;
```
Destiner, line 26: Standartizing "not authenticated" error + fixed a bug where the condition was always false.

garethfuller (changes requested):
Thanks for this Timur. I have a more general comment about the TradeError component. Really we should be using the BalAlert component for these types of error messages. However, it's not currently built to support exactly what you're trying to do here so I'll update it and commit to this PR if that's ok.

garethfuller (approved):

src/components/cards/TradeCard/TradeCard.vue
```diff
@@ -207,8 +207,48 @@ export default defineComponent({
...
+        };
+      }
+      if (isHighPriceImpact.value) {
+        return {
+          header: t('highPriceImpact'),
+          body: t('highPriceImpactDetailed'),
+          label: t('accept')
+        };
+      }
+      return undefined;
```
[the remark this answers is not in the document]
Destiner: Done

PR #389, opened 2021-05-26
WIP: add support for Polygon deployment (awaiting contract addresses)

TomAFrench:
In preparation for deployments to other networks I've gone through the frontend to handle any issues that may crop up.

Changelog:
- network configs now contain some information on the native asset on that chain. e.g. MATIC
- Localisation strings which we use which reference ETH now use this native asset info.
- Renamed InfuraService to ~~Ws~~RpcProviderService as we'll need to use non-infura providers (can revert if this is unwanted)
- Made sure that provider urls are always pulled from config.
- ~~Added [EIP3085](https://eips.ethereum.org/EIPS/eip-3085) support to prompt user to switch to app's network (silently fails for default networks in Metamask)~~

~~I've also a button to the Trade and Invest components such that when they are connected to the wrong network they are prompted to connect to the correct one. Clicking this button will do nothing if the app is set up for Mainnet or Kovan but for L2s it will bring up a metamask prompt to switch to this network.~~

fixes UI-510


src/components/cards/TradeCard/TradeCard.vue
```diff
@@ -48,6 +48,12 @@
         block
         @click.prevent="connectWallet"
       />
+      <BalBtn
+        v-else-if="networkMismatch"
+        :label="$t('networkMismatch', [appNetwork.name])"
+        block
+        @click.prevent="changeNetwork"
+      />
```
Destiner, line 56: It probably makes sense to use `BalAlert` here the same way we use it to show other errors (#387).

TomAFrench: I'm not sure that this would be a great UX change. 

I feel that the "Trade" button should clearly signal what users need to do next and make it as easy as possible.
By using `BalAlert`, we're moving the main action we're looking to push users towards off of the main focal point of the trade UI.

By the same token, I don't think we should have removed the "Connect Wallet" behaviour from that button.

TomAFrench:
Updating to reflect a conversation I had with @garethfuller before my week off. I've removed all of the UI changing elements so that we can just use the improvements to the network configs and additional services created.

src/App.vue
```diff
@@ -39,7 +39,7 @@ export default defineComponent({
     const store = useStore();
 
     // SERVICES
-    const infuraService = new InfuraService();
+    const rpcProvider = new RpcProviderService();
```
garethfuller, line 42:
```suggestion
    const providerService = new RpcProviderService();
```

src/App.vue
```diff
@@ -58,7 +58,7 @@ export default defineComponent({
     // CALLBACKS
     onBeforeMount(() => {
       store.dispatch('app/init');
-      infuraService.initBlockListener(setBlockNumber);
+      rpcProvider.initBlockListener(setBlockNumber);
```
garethfuller, line 61:
```suggestion
      providerService.initBlockListener(setBlockNumber);
```

src/lib/config/index.ts
```diff
@@ -1,6 +1,7 @@
 import homestead from './homestead.json';
 import kovan from './kovan.json';
 import rinkeby from './rinkeby.json';
+import matic from './matic.json';
```
garethfuller, line 4: Should it be called matic or polygon?

src/lib/config/matic.json
```diff
@@ -0,0 +1,45 @@
...
+  "chainId": 137,
+  "name": "Matic Mainnet",
+  "shortName": "Matic",
+  "network": "matic",
+  "unknown": false,
+  "nativeAsset": "MATIC",
+  "nativeAssetLong": "Matic",
+  "rpc": "https://polygon-mainnet.infura.io/v3/daaa68ec242643719749dd1caba2fc66",
+  "ws": "wss://rpc-mainnet.maticvigil.com/ws",
+  "explorer": "https://explorer-mainnet.maticvigil.com/",
```
garethfuller, line 12: Think we can replace this now?
```suggestion
  "explorer": "https://polygonscan.com/",
```

garethfuller (changes requested):
Just had another scan of this and have some minor comments.

src/lib/config/index.ts
TomAFrench: Ack, force of habit. Polygon team seems to split the usage across both terms (using matic in more technical writing) but it makes sense to just use Polygon.

garethfuller (changes requested):
Final thing, the previews seem to be broken by a bug that existed on the develop branch in the past. Could you merge in develop so we can see everything is fine here before we merge?

TomAFrench:
Sure thing, I've merged it in now

garethfuller (approved):

PR #394, opened 2021-05-27
refactor: replace some "any"s with concrete types

TomAFrench:
No changes to behaviour, just passing back proper types from multicall and sending transactions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "@ensdomains/content-hash": "^2.5.3",
     "@ethersproject/abi": "^5.0.1",
     "@ethersproject/address": "^5.0.1",
+    "@ethersproject/bignumber": "^5.2.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
     "@ensdomains/content-hash": "^2.5.3",
     "@ethersproject/abi": "^5.0.1",
     "@ethersproject/address": "^5.0.1",
+    "@ethersproject/bignumber": "^5.2.0",
```
garethfuller, line 29: I don't understand why this didn't already exist? We are importing from `@ethersproject/bignumber` already in multiple places...

garethfuller (approved):

package.json
TomAFrench: I saw some instances of BigNumber coming from the SOR but if I used those then I had errors coming up from presumably clashing versions.

PR #395, opened 2021-05-27
Week 53 LM

jnapier-code:
This reflects the second week of LM on v2 as proposed by the Ballers.

These changes would go into effect Monday 31st 00:00 UTC.


garethfuller:
Thanks for this. This PR is superseded by https://github.com/balancer-labs/frontend-v2/pull/397. I had to bump the version. I talked with @markusbkoch and in the future we will change the way this LM update process works so it doesn't require hotfix PRs.

PR #398, opened 2021-05-28
Handle no init liquidity on pool page

garethfuller:
- Display warning message in pool page title
- Don't render invest/withdraw card


mikemcdonald:
Since the invest/withdraw widget still does the "fade loading" it feels like something is supposed to eventually load there. Can we remove that entirely? I'd also update the text to something like `This pool has no initial liquidity. Until the pool creation UI is ready a INIT join type is required at the smart contract before invests are supported`

garethfuller:
@mikemcdonald done.

mikemcdonald (approved):

PR #400, opened 2021-05-31
update: persist invest/withdraw form state across tab changes

1saf:
Fixes # .

- Persist persist/withdraw form state across tab changes

**Testing Criteria**
- When adjusting the inputs on the invest and withdraw form, switch between tabs. Upon returning to each respective tab the inputs that were previously modified should persist.

**Questions**
- I've adjusted the behaviour to also persist the input changes when switching between sub-types in each form. (proportional vs custom), is this something that we want or not?

### DIFF
--- a/src/components/forms/pool_actions/InvestForm.vue
+++ b/src/components/forms/pool_actions/InvestForm.vue
@@ -583,7 +583,7 @@ export default defineComponent({
     watch(allTokens, newTokens => poolCalculator.setAllTokens(newTokens));
 
     watch(data, newData => {
-      emit('update:modelValue', newData);
+      // emit('update:modelValue', newData);
     });
 
     watch(
--- a/src/components/forms/pool_actions/InvestForm.vue
+++ b/src/components/forms/pool_actions/InvestForm.vue
@@ -120,20 +120,13 @@
           </div>
         </template>
         <template v-slot:info>
-          <div
-            class="cursor-pointer"
-            @click.prevent="amounts[i] = tokenBalance(i).toString()"
-          >
+          <div class="cursor-pointer" @click.prevent="setMaxInput">
             {{ $t('balance') }}: {{ formatBalance(i) }}
           </div>
         </template>
         <template v-slot:append>
           <div class="p-2">
-            <BalBtn
-              size="xs"
-              color="white"
-              @click.prevent="amounts[i] = tokenBalance(i).toString()"
-            >
+            <BalBtn size="xs" color="white" @click.prevent="setMaxInput">
               {{ $t('max') }}
             </BalBtn>
           </div>
@@ -271,6 +264,7 @@ import { FullPool } from '@/services/balancer/subgraph/types';
 import useFathom from '@/composables/useFathom';
 
 import { TOKENS } from '@/constants/tokens';
+import { reset } from 'numeral';
 
 export enum FormTypes {
   proportional = 'proportional',
@@ -561,6 +555,11 @@ export default defineComponent({
       }
     }
 
+    function setMaxInput(index: number) {
+      amounts[index] = tokenBalance(index).toString();
+      resetSlider();
+    }
+
     async function submit(): Promise<void> {
       if (!data.investForm.validate()) return;
       try {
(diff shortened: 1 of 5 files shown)


garethfuller (commented):
When I try to access a pool page the app crashes. Not sure if it's related to these changes?

1saf:
@garethfuller Crashing should not happen now

src/components/forms/pool_actions/InvestForm.vue
```diff
@@ -295,23 +297,29 @@ export default defineComponent({
...
-  emits: ['success'],
+  emits: ['success', 'update:modelValue'],
 
   props: {
     pool: { type: Object as PropType<FullPool>, required: true },
-    missingPrices: { type: Boolean, default: false }
+    missingPrices: { type: Boolean, default: false },
+    initialState: {
+      type: Object as PropType<
+        Pick<DataProps, 'investType' | 'range' | 'amounts'>
```
garethfuller, line 307: Is the `DataProps` definition missing from this component?

src/components/forms/pool_actions/WithdrawForm.vue
```diff
@@ -214,29 +215,49 @@ export enum FormTypes {
...
+    initialState: {
+      type: Object as PropType<
+        Pick<DataProps, 'bptIn' | 'withdrawType' | 'range' | 'amounts'>
+      >,
+      default: () => ({})
+    }
   },
 
-  setup(props: { pool: FullPool }, { emit }) {
+  setup(props: { pool: FullPool; initialState }, { emit }) {
```
garethfuller, line 249: Same comment here as for the InvestForm

garethfuller:
Something messed up seems to be happening with the balances on the withdraw tab. Not sure if it's caused by this PR or was already a bug.

Screenshot 2021-06-02 at 11 09 27

src/components/forms/pool_actions/InvestForm.vue
```diff
@@ -295,23 +297,29 @@ export default defineComponent({
...
+      >,
+      default: () => ({})
+    }
   },
 
-  setup(props: { pool: FullPool }, { emit }) {
+  setup(props: { pool: FullPool; initialState }, { emit }) {
```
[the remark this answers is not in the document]
1saf: yep that is a much cleaner impl

1saf:
> Something messed up seems to be happening with the balances on the withdraw tab. Not sure if it's caused by this PR or was already a bug.
> 
> Screenshot 2021-06-02 at 11 09 27

hmm ill look at it

garethfuller:
Another small bug: https://www.loom.com/share/6a8f33cb48c7411ab0a95d8bee52a220

1saf:
> Another small bug:
> 
> https://www.loom.com/share/6a8f33cb48c7411ab0a95d8bee52a220

Ah this is what I meant by persisting the behaviour between changing between proportional and custom inputs. So in this case we want to persist in tab changes but reset if switching between the type of invest/withdrawal?

garethfuller:
> > Another small bug:
> > https://www.loom.com/share/6a8f33cb48c7411ab0a95d8bee52a220
> 
> Ah this is what I meant by persisting the behaviour between changing between proportional and custom inputs. So in this case we want to persist in tab changes but reset if switching between the type of invest/withdrawal?

Well it's not so much that it's the wrong behaviour, but when I switch back to proportional, notice how the percentage says 100% but the slider is still at like 50%. Happy to have persistent state between types.

One other thing to check before we merge this, test if when a large investment/withdrawal in the pool happens, the proportional amounts change. To do this you need to invest in the pool via another window.

garethfuller:
small bug demo: https://www.loom.com/share/99b42f74debd44d3a9e3b43d5e2d485a

I don't think when you go to custom amounts, enter some values and switch back to proportional that it should maintain the custom numbers in the proportional case in anyway. There will be too many edge cases if we do that. Best just to reset the proportional case on type switch or maintain the proportional state prior to switching.

In general lets try not to overcomplicate this logic. When we get into a redesign of this flow, it shouldn't be needed.

PR #403, opened 2021-06-02
feature: blocknative onboard proof of concept

1saf:
Fixes # .

- 
- 
-


src/composables/useOnboard.ts
```diff
@@ -0,0 +1,27 @@
+import { inject, Ref } from 'vue';
+// @ts-ignore
```
1saf, line 2: remember to remove

garethfuller:
Weird z-index issues to fix. Top nav, and table column should both be below the modal overlay.

Screenshot 2021-06-02 at 20 06 28

src/composables/useOnboard.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export type Web3State = {
+  onboardInstance: Ref<null | API>;
+  web3Instance: Ref<null | Web3>;
+};
+
+export const Web3ProviderSymbol = Symbol('WEB3_PROVIDER');
+
+export default function useOnboard() {
+  const web3State = inject<Web3State>(Web3ProviderSymbol);
```
garethfuller, line 14: Maybe we could just add this to `useBlocknative` since we're already doing a similar thing there for notify.js? It's also where we have access to the blocknative sdk so it kind of is a wrapper for all the blocknative plugins.

src/providers/Web3Provider.vue
```diff
@@ -0,0 +1,41 @@
...
+    });
+
+    provide(Web3ProviderSymbol, toRefs(readonly(web3State)));
+  },
+  render() {
+    // @ts-ignore
+    return this.$slots.default();
+  }
+});
+</script>
```
garethfuller, line 41: I might be wrong but I feel like this logic (e.g. setting up the instance) could be included in the [blocknative Vue plugin](https://github.com/balancer-labs/frontend-v2/blob/develop/src/plugins/blocknative.ts) I built, rather than being a wrapper component. Where did you get this pattern from?

garethfuller (commented):
Looks like the flow will be a good fit, but I think there is going to be a lot of logic to adjust as a result. This will probably turn out to be a big PR. Once you've fixed the overlay issues lets share with Pon/Mike and get their feedback on wether to proceed.

garethfuller:
@1saf I think we can close this now since it looks like we'll go with the native vue option. We can always re-open if we decide otherwise.

PR #406, opened 2021-06-03
Call handleSwitchTokens if tokenIn/Out === tokenIn/Out on selection.

johngrantuk:
Fixes #UI-469.

- On selection check if tokenIn/Out === tokenIn/Out (i.e. same token>token trade)
- If it does call handleSwitchTokens to flip tokenIn/Out
-

### DIFF
--- a/src/components/cards/TradeCard/TradePair.vue
+++ b/src/components/cards/TradeCard/TradePair.vue
@@ -263,9 +263,15 @@ export default defineComponent({
 
     function handleSelectToken(address: string): void {
       if (modalSelectTokenType.value === 'input') {
-        emit('tokenInAddressChange', address);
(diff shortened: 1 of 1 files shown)


garethfuller (approved):
Nice, seems to work well! One question, why do you not want to emit the address change event if it's a token switch case? The token address is still changing.

johngrantuk:
> Nice, seems to work well! One question, why do you not want to emit the address change event if it's a token switch case? The token address is still changing.

The handleSwitchTokens function actually emits the events: https://github.com/balancer-labs/frontend-v2/blob/22e2fb786680ea4734a158330fdf0aed747ab00d/src/components/cards/TradeCard/TradePair.vue#L259

PR #416, opened 2021-06-07
single subdomain level for all networks

mikemcdonald:
AFAIK this is the only place where the domain needs to be updated. If this looks good I can update the Vercel config to use `kovan.balancer.fi` so that moving forward all networks will use a single subdomain instead of `kovan.app.balancer.fi`

### DIFF
--- a/src/components/cards/TradeCard/TradeRoute.vue
+++ b/src/components/cards/TradeCard/TradeRoute.vue
@@ -427,7 +427,7 @@ export default defineComponent({
     function getPoolLink(id: string): string {
       const { chainId } = getConfig();
       const prefixMap = {
-        1: '',
+        1: 'app.',
         42: 'kovan.'
       };
       const prefix = prefixMap[chainId] || '';
--- a/src/components/cards/TradeCard/TradeRoute.vue
+++ b/src/components/cards/TradeCard/TradeRoute.vue
@@ -433,7 +433,7 @@ export default defineComponent({
       const prefix = prefixMap[chainId] || '';
       return props.sorReturn.isV1swap
         ? `https://${prefix}pools.balancer.exchange/#/pool/${id}`
-        : `https://${prefix}app.balancer.fi/#pool/${id}`;
+        : `https://${prefix}balancer.fi/#pool/${id}`;
     }
 
     return {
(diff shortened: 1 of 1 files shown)


src/components/cards/TradeCard/TradeRoute.vue
```diff
@@ -433,7 +433,7 @@ export default defineComponent({
       const prefix = prefixMap[chainId] || '';
       return props.sorReturn.isV1swap
         ? `https://${prefix}pools.balancer.exchange/#/pool/${id}`
-        : `https://${prefix}app.balancer.fi/#pool/${id}`;
+        : `https://${prefix}balancer.fi/#pool/${id}`;
```
garethfuller, line 436: I think if we remove the `app.` here it will break for mainnet, so you will need to update the `prefixMap` [just above](https://github.com/balancer-labs/frontend-v2/blob/7a432cb0d06bdb9526e6b949ebf7745382e6ba3d/src/components/cards/TradeCard/TradeRoute.vue#L430). So that mainnet returns `app.`

garethfuller (changes requested):

garethfuller:
@mikemcdonald  Yeah sounds good, we just need to make sure to update the `VUE_APP_DOMAIN` env var in Vercel too.

src/components/cards/TradeCard/TradeRoute.vue
mikemcdonald: Added a separate check for v1 mainnet. Since adding `app.` to the prefix would break v1 url

garethfuller (approved):

src/components/cards/TradeCard/TradeRoute.vue
```diff
@@ -427,13 +427,17 @@ export default defineComponent({
...
       const prefixMap = {
-        1: '',
+        1: 'app.',
         42: 'kovan.'
       };
       const prefix = prefixMap[chainId] || '';
-      return props.sorReturn.isV1swap
-        ? `https://${prefix}pools.balancer.exchange/#/pool/${id}`
-        : `https://${prefix}app.balancer.fi/#pool/${id}`;
+      if (props.sorReturn.isV1swap && chainId === 1) {
```
evgenyboxer, line 434: We can now do something like `chainId === Network.MAINNET` (imported from `network.ts`).

mikemcdonald:
Ah sorry missed this before, yes this is good to merge

PR #419, opened 2021-06-09
feature: Vue Native wallet connection

1saf:
Fixes # .

- 
- 
-


src/components/cards/TradeCard/TradePair.vue
```diff
@@ -149,7 +149,7 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, toRefs, computed, ref } from 'vue';
+import { defineComponent, toRefs, computed, ref, watch } from 'vue';
```
evgenyboxer, line 152: doesn't seem to be used?

src/components/cards/TradeCard/TradePair.vue
```diff
@@ -158,6 +158,7 @@ import { ETHER } from '@/constants/tokenlists';
 
 import TradePairToggle from '@/components/cards/TradeCard/TradePairToggle.vue';
 import SelectTokenModal from '@/components/modals/SelectTokenModal/SelectTokenModal.vue';
+import useTokenLists from '@/composables/useTokenLists';
```
evgenyboxer, line 161: This one as well (not used)

src/services/web3/connectors/connector.ts
```diff
@@ -0,0 +1,69 @@
...
+    this.selectedAccount = selectedAccount || '';
+  }
+
+  // must return the provider
+  abstract connect(): Promise<ConnectorPayload>;
+
+  handleAccountsChanged = accounts => {
+    if (this.selectedAccount !== '') {
+      const account = accounts.find(
+        a => getAddress(a) === getAddress(this.selectedAccount)
```
evgenyboxer, line 28:
```suggestion
        account => getAddress(account) === getAddress(this.selectedAccount)
```

src/services/web3/connectors/metamask/metamask.connector.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (provider) {
+      this.provider = provider;
+      this.active.value = true;
+
+      try {
+        const accounts = await provider.request({
+          method: 'eth_requestAccounts'
+        });
+
+        const chainId = await provider.request({ method: 'eth_chainId' });
```
evgenyboxer, line 16: Seems like `accounts` and `chainId` can be combined to a `promise.all`

src/services/web3/useVueWeb3.ts
```diff
@@ -0,0 +1,94 @@
...
+      enabled: canLoadProfile
+    })
+  );
+
+  // TODO seperate this out?
+  // load up a list of all the evm chains
+  const { data: evmChains, isLoading: isLoadingEvmChains } = useQuery(
+    'EVM_CHAINS',
+    async () =>
+      await (await axios.get('https://chainid.network/chains.json')).data
```
evgenyboxer, line 62: double `await` ^

src/services/web3/connectors/connector.ts
```diff
@@ -0,0 +1,69 @@
...
+  account: Ref<string | null>;
+  chainId: Ref<number | null>;
+};
+export abstract class Connector {
+  provider: any = null;
+  account: Ref<string | null> = ref(null);
+  chainId: Ref<number | null> = ref(null);
+  active: Ref<boolean> = ref(false);
+  selectedAccount = '';
+  id = 'unknown';
```
evgenyboxer, line 16: I think we should add a `signer` here as well so that it's easily available when you need it. (its essentially calling `getSigner()` on the provider)

src/App.vue
```diff
@@ -1,17 +1,16 @@
 <template>
+  <div id="modal" />
```
evgenyboxer, line 2: Is this a fix for the modal behaviour?

src/components/forms/pool_actions/WithdrawForm.vue
```diff
@@ -289,6 +294,7 @@ export default defineComponent({
     });
 
     const propMaxUSD = computed(() => {
+      console.log('ropmax', data.propMax);
```
evgenyboxer, line 297: remove `console.log`^

src/components/forms/pool_actions/WithdrawForm.vue
```diff
@@ -313,7 +319,11 @@ export default defineComponent({
     });
 
     const bptBalance = computed(() => {
-      return allTokens.value[props.pool.address].balance;
+      console.log(
+        'tokendict',
+        tokenDictionary.value[props.pool.address].balance
+      );
```
evgenyboxer, line 325: remove `console.log`^

src/composables/useAllowances.ts
```diff
@@ -0,0 +1,95 @@
...
+  );
+
+  const isQueryEnabled = computed(() => account && tokens.value.length > 0);
+  const {
+    data: allowances,
+    isLoading,
+    isFetching,
+    refetch: refetchAllowances
+  } = useQuery(
+    reactive(['ALLOWANCES', { chainId, account, dstList, tokens }]),
```
evgenyboxer, line 45: can `ALLOWANCES` be part of `queryKeys`

evgenyboxer:
I've played a bit with the wallet and tried to make some account switches / account lock / etc, and I've had some errors thrown when trying to do it.

Screen Shot 2021-06-23 at 12 16 28 pm
Screen Shot 2021-06-23 at 12 18 52 pm
Screen Shot 2021-06-23 at 12 19 01 pm
Screen Shot 2021-06-23 at 12 11 57 pm

src/services/web3/useVueWeb3.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  const canLoadProfile = computed(
+    () => account.value !== '' && chainId.value !== 0
+  );
+
+  const getProvider = () => new Web3Provider(provider.value as any);
+
+  // TODO separate this out?
+  const { isLoading: isLoadingProfile, data: profile } = useQuery(
+    ['WEB3_PROFILE', { account, chainId }],
```
evgenyboxer, line 50: Can we have this in queryKeys?

src/services/web3/useVueWeb3.ts
```diff
@@ -0,0 +1,94 @@
...
+    () => getProfile(account.value, String(chainId.value)),
+    reactive({
+      enabled: canLoadProfile
+    })
+  );
+
+  // TODO seperate this out?
+  // load up a list of all the evm chains
+  const { data: evmChains, isLoading: isLoadingEvmChains } = useQuery(
+    'EVM_CHAINS',
```
evgenyboxer, line 60: This one as well

src/services/web3/connectors/metamask/metamask.connector.ts
1saf: I could do this, but I prefer the readability of having them separate

src/services/web3/connectors/connector.ts
1saf: I believe most provider consumers will extract the signer for you if you simply provide a signer capable provider? I could be wrong

src/App.vue
1saf: Yes

1saf:
> I've played a bit with the wallet and tried to make some account switches / account lock / etc, and I've had some errors thrown when trying to do it.
> 
> Screen Shot 2021-06-23 at 12 16 28 pm
> 
> Screen Shot 2021-06-23 at 12 18 52 pm
> 
> Screen Shot 2021-06-23 at 12 19 01 pm
> 
> Screen Shot 2021-06-23 at 12 11 57 pm
> 
> Other than those errors it seems to connect way faster!

Resolved this :)

src/services/web3/connectors/connector.ts
evgenyboxer: That's an awesome idea, I wish it worked this way - but it seems you have to grab the signer from the provider to do wallet stuff (see `sendTransaction` function in `web3.ts`)

johngrantuk (approved):
Look good. Mentioned Trade balance issue to Ansaf directly.

src/services/web3/useVueWeb3.ts
```diff
@@ -0,0 +1,104 @@
...
+  const {
+    connectWallet,
+    account,
+    chainId,
+    disconnectWallet,
+    connector,
+    provider,
+    walletState
+  } = inject(Web3ProviderSymbol) as Web3Plugin;
+  const appConfig = new ConfigService();
```
garethfuller, line 22:
```suggestion
  const configService = new ConfigService();
```

src/services/web3/useVueWeb3.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  const userNetworkConfig = computed(() =>
+    appConfig.getNetworkConfig(String(chainId.value))
```
garethfuller, line 25:
```suggestion
    configService.getNetworkConfig(String(chainId.value))
```

src/services/web3/useVueWeb3.ts
```diff
@@ -0,0 +1,104 @@
...
+  );
+  const appNetworkConfig = appConfig.network;
```
garethfuller, line 27:
```suggestion
  const appNetworkConfig = configService.network;
```

src/services/web3/useVueWeb3.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  const networkName = computed(() => {
+    if (!isLoadingEvmChains.value) {
+      const chain = evmChains.value.find(
+        chain => chain.networkId === userNetworkConfig.value.chainId
+      );
+      return chain?.name || 'Unknown Network';
+    }
+    return 'Loading...';
+  });
```
garethfuller, line 76: Why do we need this? The networkName should come from the configs we already have. https://github.com/balancer-labs/frontend-v2/blob/develop/src/lib/config/homestead.json#L4

src/services/web3/useVueWeb3.ts
```diff
@@ -0,0 +1,104 @@
...
+  const getProvider = () => new Web3Provider(provider.value as any);
+
+  // TODO separate this out?
+  const { isLoading: isLoadingProfile, data: profile } = useQuery(
+    QUERY_KEYS.Account.Profile(account, userNetworkConfig),
+    () => getProfile(account.value, String(userNetworkConfig.value.chainId)),
+    reactive({
+      enabled: canLoadProfile
+    })
+  );
```
garethfuller, line 59: I'd quite like to see this separated out. In the format we have the other queries. I know it's a really simple one, but if you're not familiar with reactQuery it's actually quite hard to parse and understand what's happening.

Also, could you adjust `getProfile` to not fetch the 3box data. We basically only want to fetch the ens name now.

src/services/web3/web3.plugin.ts
```diff
@@ -0,0 +1,148 @@
...
+import { Web3Provider } from '@ethersproject/providers';
+import { WalletLinkConnector } from './connectors/walletlink/walletlink.connector';
+
+export type Wallet = 'metamask' | 'walletconnect' | 'walletlink';
+export const SupportedWallets = [
+  'metamask',
+  'walletconnect',
+  'walletlink'
+] as Wallet[];
+export const WalletNameDictionary: Record<Wallet, string> = {
```
garethfuller, line 16: What do you think about using 'map' terminology instead of dictionary throughout the app? Dictionary is such a long word! 😅 
```suggestion
export const WalletNameMap: Record<Wallet, string> = {
```

src/services/web3/web3.plugin.ts
```diff
@@ -0,0 +1,148 @@
...
+type WalletState = 'connecting' | 'connected' | 'disconnected';
+type PluginState = {
+  connector: any;
+  walletState: WalletState;
+};
+
+export default {
+  install: async app => {
+    const alreadyConnectedAccount = ref(lsGet('connectedWallet', null));
+    const alreadyConnectedProvider = ref(lsGet('connectedProvider', null));
```
garethfuller, line 49: Nice! So this allows us to not have to reload the connection all the time right?

src/services/web3/web3.plugin.ts
```diff
@@ -0,0 +1,148 @@
...
+      disconnectWallet,
+      ...toRefs(pluginState),
+      account,
+      chainId,
+      provider
+    };
+
+    app.provide(Web3ProviderSymbol, payload);
+  }
+};
```
garethfuller, line 148: I'm ok with the logic here instead of abstracting elsewhere actually.

garethfuller (changes requested):
Nice work, looks great! I've left some redundant comments (from our call) in the review just for the record.

src/services/web3/components/WalletSelectModal.vue
```diff
@@ -0,0 +1,28 @@
+<template>
```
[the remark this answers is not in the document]
1saf: This has been resolved, thanks for raising. Components are now part of the bal component library (non-global)

src/components/forms/pool_actions/InvestForm.vue
```diff
@@ -561,16 +570,23 @@ export default defineComponent({
...
+          onTxConfirmed: async (tx: TransactionData) => {
             emit('success', tx);
             data.amounts = [];
             data.loading = false;
+            await refetchBalances.value();
+            setPropMax();
+            if (hasZeroBalance.value) {
+              data.investType = FormTypes.custom;
+              resetSlider();
+            }
```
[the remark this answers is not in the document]
1saf: I'll re-test and make sure we do not need this.

src/services/web3/useVueWeb3.ts
1saf: I originally had this because if the network is unsupported - it will not show anything as part of the navbar settings modal. I can change it to use from the config.

src/services/web3/useVueWeb3.ts
1saf: No worries.

src/services/web3/web3.plugin.ts
1saf: Done deal

src/services/web3/web3.plugin.ts
1saf: That is correct, it's mainly for when a user comes back to the app, it will load the provider + account that was previously used.

src/services/web3/useVueWeb3.ts
garethfuller: Yeah I don't think we need to add another network request for this. Especially if we go with a full-page modal for unsupported networks. We can also just use the term 'Unsupported network' in that modal rather than fetching the network's name.

src/services/web3/useVueWeb3.ts
1saf: Leaving this here actually, conversation around too much abstraction. It is a pretty simply query with not much ancillary compute happening, so will leave here. We can revisit to separate it out if it becomes a big boy

src/services/web3/useVueWeb3.ts
1saf: All fixed

src/services/web3/connectors/connector.ts
1saf: added in a signer which gets returned from the composable :)

garethfuller (approved):
Should we not turn this off now: https://github.com/balancer-labs/frontend-v2/blob/378d67d838e39353989626e9dad02a0b302c63dc/src/store/modules/app.ts#L44-L46

1saf:
> Should we not turn this off now:
> 
> https://github.com/balancer-labs/frontend-v2/blob/378d67d838e39353989626e9dad02a0b302c63dc/src/store/modules/app.ts#L44-L46

yeos

PR #426, opened 2021-06-14
Missing past pools

markusbkoch:
Fixes an issue where the Nth to Mth pools in the time travel query are not the same Nth to Mth in the current state query, in which case 24h volume is incorrectly returned.

### DIFF
--- a/src/services/balancer/subgraph/entities/pools/index.ts
+++ b/src/services/balancer/subgraph/entities/pools/index.ts
@@ -49,7 +49,7 @@ export default class Pools {
     const block = { number: await this.timeTravelBlock(period) };
     const isCurrentPool = { id_in: currentPools.map(pool => pool.id) };
     const pastPoolsQuery = this.query(
-      { ...args, where: isCurrentPool, block },
+      { where: isCurrentPool, block },
       attrs
     );
     const { pools: pastPools } = await this.service.client.get(pastPoolsQuery);
--- a/src/services/balancer/subgraph/entities/pools/index.ts
+++ b/src/services/balancer/subgraph/entities/pools/index.ts
@@ -48,10 +48,7 @@ export default class Pools {
     // Get past state of current pools
     const block = { number: await this.timeTravelBlock(period) };
     const isCurrentPool = { id_in: currentPools.map(pool => pool.id) };
-    const pastPoolsQuery = this.query(
-      { where: isCurrentPool, block },
-      attrs
-    );
+    const pastPoolsQuery = this.query({ where: isCurrentPool, block }, attrs);
     const { pools: pastPools } = await this.service.client.get(pastPoolsQuery);
 
     return this.serialize(currentPools, pastPools, period, prices);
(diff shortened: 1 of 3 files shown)


src/services/balancer/subgraph/entities/pools/index.ts
```diff
@@ -40,13 +40,17 @@ export default class Pools {
...
-      currentPools: this.query(args, attrs).pools,
-      pastPools: this.query({ ...args, block }, attrs).pools
+    const currentPoolsQuery = {
+      currentPools: this.query(args, attrs).pools
     };
+    const { currentPools } = await this.service.client.get(currentPoolsQuery);
 
-    const { currentPools, pastPools } = await this.service.client.get(query);
+    const block = { number: await this.timeTravelBlock(period) };
+    const pools_of_interest = { id_in: currentPools.map(pool => pool.id) };
```
evgenyboxer, line 49:
```suggestion
    const poolsOfInterest = { id_in: currentPools.map(pool => pool.id) };
```

src/services/balancer/subgraph/entities/pools/index.ts
```diff
@@ -40,13 +40,17 @@ export default class Pools {
...
+    const pastPoolsQuery = {
+      pastPools: this.query({ where: pools_of_interest, block }, attrs).pools
```
evgenyboxer, line 51:
```suggestion
      pastPools: this.query({ ...args, where: poolsOfInterest, block }, attrs).pools
```

src/services/balancer/subgraph/entities/pools/index.ts
```diff
@@ -40,13 +40,17 @@ export default class Pools {
...
-    const query = {
-      currentPools: this.query(args, attrs).pools,
-      pastPools: this.query({ ...args, block }, attrs).pools
+    const currentPoolsQuery = {
+      currentPools: this.query(args, attrs).pools
     };
+    const { currentPools } = await this.service.client.get(currentPoolsQuery);
 
-    const { currentPools, pastPools } = await this.service.client.get(query);
+    const block = { number: await this.timeTravelBlock(period) };
```
evgenyboxer, line 48: `currentPools` and `block` can be part of a `promise.all` so we could benefit from parallelism

evgenyboxer:
In terms of having both of the queries in 1 request - I think, since these are now dependent queries, your solution looks good to me.

garethfuller:
@markusbkoch thanks for spotting and fixing this 🙌 

I just [refactored a little](https://github.com/balancer-labs/frontend-v2/pull/426/commits/986b10b40faac9f91739b63a0148f3db86b3d315), since we can no longer combine the queries in one request the graphql aliases are no longer needed.

Could you do one final sense check and let me know if everything is ok? Then I can do the hotfix release.

src/services/balancer/subgraph/entities/pools/index.ts
markusbkoch: Thanks Ev. On my tests, reintroducing args leads to the bug coming back for the GTC/WETH

garethfuller: I initially thought Ev was right here but since it's just for the pastPools query and we're fetching the exact pools by Id, the passed args should be irrelevant.

PR #430, opened 2021-06-15
Fix potential race condition

garethfuller:
Fixes #UI-555

- Help prevent race condition between blocknative txConfirmation and allowances update onchain

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer-labs/frontend-v2",
-  "version": "1.4.1",
+  "version": "1.4.2",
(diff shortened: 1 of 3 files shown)


evgenyboxer:
@garethfuller I've actually encountered this type of issue before (The claim button uses both blocknative / infura to deal with the race issue). So essentially after your fire the listener you also wait for the transaction to be mined via `await tx.wait()`

I kinda think the long-term solution would actually be not to use Blocknative for transactions and roll our own transaction handler. (Gnosis for example needs a custom solution anyway and cannot use Blocknative)

PR #435, opened 2021-06-17
refactor: store subgraph urls in network config json files

TomAFrench:
I've consolidated the subgraph urls so that they live in the relevant network config. The subgraph service can then pull them through the `ConfigService`.

As an aside, do we want to make the methods on the `ConfigService` static?


garethfuller (approved):
Nice work! Thanks 🙌

PR #437, opened 2021-06-18
Polygon MVP

TomAFrench:
I'm trying to flush out any possible bugs and get a rough version of the frontend which allows trading on Polygon. We have contracts + subgraph deployed so all the tasks from here is frontend related.

Todo:
- [x] deploy a set of initial pools and set `poolsUrlV2` in `polygon.json` so that we can test trading.
- [x] Add Polygon assets to a Balancer tokenlist. (We can use the Quickswap tokenlist as a starting point)
- [x] ~~add mappings for Polygon assets to translate to mainnet addresses to query CoinGecko with. (We can query the Matic bridge for these rather than maintaining a huge mapping)~~ Luckily, CoinGecko allows us to query with the Polygon addresses

Bugs:
- [x] Balancer tokenlist doesn't include Polygon assets and we can't easily override the default tokenlist for Polygon which causes `GasReimbursement` to crash as it doesn't receive addreses.

### DIFF
--- a/src/constants/initialTokens.json
+++ b/src/constants/initialTokens.json
@@ -6,5 +6,9 @@
   "42": {
     "input": "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE",
     "output": "0x04DF6e4121c27713ED22341E7c7Df330F56f289B"
+  },
+  "137": {
+    "input": "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE",
+    "output": "0x8f3cf7ad23cd3cadbd9735aff958023239c6a063"
   }
 }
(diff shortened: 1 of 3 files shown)


TomAFrench:
This default tokenlist differing from what is defined in the registry is causing the issues with `GasReimbursement`: https://github.com/balancer-labs/frontend-v2/blob/be384c9e5679c118afe16581d438d6652b79fd51/src/composables/useTokenLists.ts#L70

garethfuller:
> This default tokenlist differing from what is defined in the registry is causing the issues with `GasReimbursement`:
> 
> https://github.com/balancer-labs/frontend-v2/blob/be384c9e5679c118afe16581d438d6652b79fd51/src/composables/useTokenLists.ts#L70

Did you fix this? If not maybe @1saf can take a look since he built this composable.

TomAFrench:
> Did you fix this? If not maybe @1saf can take a look since he built this composable.

Yeah, I did. Turns out that my reasoning was off in that comment and it's purely because we didn't fail gracefully when we didn't handle a given network in `eligible.json`. This was fixed in #448

garethfuller (approved):

PR #438, opened 2021-06-18
refactor: prep tokenlists definitions for multichain

TomAFrench:
I've added a mapping in `tokenLists.ts` to allow us to choose a set of token lists depending on the network so that we can have a separate choice of tokenlists for Polygon, etc.

I've also added a small function to allow us to deduplicate some of the definitions in there.

### DIFF
--- a/src/constants/tokenlists.ts
+++ b/src/constants/tokenlists.ts
@@ -43,7 +43,9 @@ type TokenLists = {
   All: string[];
   Balancer: {
     All: string[];
+    // Compliant list for exchange
     Default: string;
+    // Extended list to include LBP tokens
     Vetted: string;
   };
   Approved: string[];
@@ -71,13 +73,6 @@ const processTokenLists = (config: TokenListConfig): TokenLists => {
   };
 };
 
-const FULL_TOKEN_LISTS = processTokenLists(TOKEN_LISTS_MAP[env.NETWORK]);
+const TOKEN_LISTS = processTokenLists(TOKEN_LISTS_MAP[env.NETWORK]);
 
-// Compliant list for exchange
-export const TOKEN_LIST_DEFAULT = FULL_TOKEN_LISTS.Balancer.Default;
-// Extended list to include LBP tokens
-export const VETTED_TOKEN_LIST = FULL_TOKEN_LISTS.Balancer.Vetted;
-
-export const TOKEN_LISTS = FULL_TOKEN_LISTS.Approved;
-
-export default FULL_TOKEN_LISTS;
+export default TOKEN_LISTS;
--- a/src/constants/tokenlists.ts
+++ b/src/constants/tokenlists.ts
@@ -2,6 +2,8 @@ import useConfig from '@/composables/useConfig';
 
 const { env, networkConfig } = useConfig();
 
+// TODO replace imports of ETHER throughout the app so we can remove it from here.
+// It doesn't make sense to have this ETHER definition here, this file should be for tokenlist URIs only.
 export const ETHER = networkConfig.nativeAsset;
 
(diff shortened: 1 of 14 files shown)


src/constants/tokenlists.ts
```diff
@@ -1,49 +1,78 @@
...
   id: 'ether',
-  name: 'Ether',
+  name: networkConfig.nativeAssetLong,
   address: '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE',
-  symbol: 'ETH',
+  symbol: networkConfig.nativeAsset,
   decimals: 18,
   logoURI:
     'https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2/logo.png'
 };
```
garethfuller, line 12: Instead of mixing in the network config here, maybe we should move the native asset definition into the configs, wdyt?

TomAFrench: Yeah, I think that's sensible. Gimme a couple mins.

src/constants/tokenlists.ts
```diff
@@ -1,49 +1,74 @@
...
 export const ETHER = {
-  id: 'ether',
-  name: 'Ether',
-  address: '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE',
-  symbol: 'ETH',
-  decimals: 18,
-  logoURI:
-    'https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2/logo.png'
+  ...networkConfig.nativeAsset,
+  id: 'ether'
```
garethfuller, line 7: Is this ID used anywhere? If not maybe we can get rid of the ETHER definition in this file altogether? I feel like it doesn't belong here at all.

TomAFrench: image

It looks like we use it to allow having app.balancer.fi/ether/0x123... redirect to app.balancer.fi/0xee...ee/0x123... This behaviour seems a little odd to me (it's kinda a weird halfway step to readable deeplinks) so I think it's safe to remove but didn't want to start carving out behaviour.

garethfuller: Yeah this is kind of weird behaviour but it would be good to maintain those deeplinks in case they're being used. Perhaps we can add a user friendly `deeplinkId` to the native asset config and then just reference that here in `populateInitialTokens`?

TomAFrench: Sure thing, I've made this change.

I don't think it would be right to remove `ETHER` entirely in this PR as it's spread out over the codebase so should probably be handled separately.

src/constants/tokenlists.ts
```diff
@@ -1,49 +1,98 @@
...
 
-export const extLists = [
-  'ipns://tokens.uniswap.org',
-  'tokenlist.zerion.eth',
-  'tokens.1inch.eth',
-  'tokenlist.aave.eth',
-  'https://tokens.coingecko.com/uniswap/all.json',
-  'https://umaproject.org/uma.tokenlist.json'
-];
+export const ETHER = networkConfig.nativeAsset;
```
garethfuller, line 5:
```suggestion
// TODO replace imports of ETHER throughout the app so we can remove it from here.
// It doesn't make sense to have this ETHER definition here, this file should be for tokenlist URIs only.
export const ETHER = networkConfig.nativeAsset;
```

garethfuller (approved):

PR #439, opened 2021-06-19
feat: gnosis tx deadline + slippage + wrap/unwrap ETH

evgenyboxer:
The PR has a couple of features - 

1. Transaction deadline (which appears only when Gnosis interface is selected).
2. Slippage is now working properly.
3. Gnosis fee is now correctly being deducted from the sell side.
4. Quotes refresh on every block.
5. Remove price impact (isn't really used)
6. Wrap/Unwrap ETH support.

Note - there won't be any visible changes, it's only when Gnosis Integration is turned on.


src/components/modals/TradePreviewModalGP.vue
```diff
@@ -18,6 +22,10 @@
         </div>
       </div>
       <div>
+        <div class="mt-6 mb-3 text-sm">
+          Fee: {{ fNum(formattedFeeAmount, 'token') }}
```
evgenyboxer, line 26: The trade preview will contain the fee info, so I've added it here - but it isn't the final design.

src/components/cards/TradeCardGP/TradeCardGP.vue
```diff
@@ -12,13 +12,13 @@
         :token-in-address-input="tokenInAddress"
         :token-out-amount-input="tokenOutAmount"
         :token-out-address-input="tokenOutAddress"
-        :is-sell="isSell"
+        :exact-in="exactIn"
```
evgenyboxer, line 15: Switched to `exactIn` instead of `isSell` to match what we have on Balancer.

src/components/cards/TradeCardGP/TradeCardGP.vue
```diff
@@ -381,14 +397,24 @@ export default defineComponent({
...
-      updateQuotes();
+      if (tokenOutAmount.value !== '') {
+        updateQuotes();
+      } else {
+        tokenInAmount.value = '';
+      }
     });
 
     watch(blockNumber, async () => {
+      updateQuotes();
```
evgenyboxer, line 416: Quotes update on every block change (might optimize later)

src/constants/options.ts
```diff
@@ -0,0 +1,16 @@
...
+    label: option,
+    value: option
+  }));
+
+export const tradeInterfaceOptions = Object.values(TradeInterface)
+  .filter(v => typeof v === 'string')
+  .map(option => ({
+    label: option,
+    value: option
+  }));
```
evgenyboxer, line 16: Since we need these options both in the app settings / trade settings I thought its better to move them into a shared file.

src/components/popovers/TradeSettingsPopover.vue
```diff
@@ -24,7 +24,10 @@
       </div>
       <AppSlippageForm class="mt-1" />
     </div>
-    <div v-if="!hideLiquidity" class="mt-6">
+    <div
+      v-if="appTradeInterface === TradeInterface.BALANCER && !hideLiquidity"
```
evgenyboxer, line 28: Enable only for Balancer interface

src/components/cards/TradeCardGP/TradeCardGP.vue
```diff
@@ -121,11 +123,12 @@ export default defineComponent({
   setup() {
     const highPiAccepted = ref(false);
     const priceImpact = ref(0);
-    const isSell = ref(true);
+    const exactIn = ref(true);
     const feeQuote = ref<FeeInformation | null>(null);
     const feeExceedsPrice = ref(false);
     const orderId = ref('');
     const orderMetadata = ref<OrderMetaData | null>(null);
+    const updatingQuotes = ref(false);
```
evgenyboxer, line 131: This isn't used for now in the UI - but it will show that a quote is being updated (mostly for initial load/switching pairs)

src/components/cards/TradeCardGP/TradeCardGP.vue
```diff
@@ -119,34 +126,41 @@ export default defineComponent({
...
     const tradeSuccess = ref(false);
     const trading = ref(false);
-    const txHash = ref('');
     const modalTradePreviewIsOpen = ref(false);
 
+    // COMPUTED
+    const getTokens = (params = {}) =>
+      store.getters['registry/getTokens'](params);
+    const getConfig = () => store.getters['web3/getConfig']();
+    const tokens = computed(() => getTokens({ includeEther: true }));
```
garethfuller, line 159:
```suggestion
    const { allTokensIncludeEth: tokens } = useTokens();
```
This composable will be different after the registry is fully refactored and implemented but for now we have this legacy composable that works instead of using the getter.

src/components/cards/TradeCardGP/TradeCardGP.vue
```diff
@@ -119,34 +126,41 @@ export default defineComponent({
...
     const tokenOutAddress = ref('');
     const tokenOutAmount = ref('');
     const tradeSuccess = ref(false);
     const trading = ref(false);
-    const txHash = ref('');
     const modalTradePreviewIsOpen = ref(false);
 
+    // COMPUTED
+    const getTokens = (params = {}) =>
+      store.getters['registry/getTokens'](params);
```
garethfuller, line 157:
```suggestion
```

src/components/cards/TradeCardGP/TradeCardGP.vue
```diff
@@ -119,34 +126,41 @@ export default defineComponent({
...
+    const getConfig = () => store.getters['web3/getConfig']();
```
garethfuller, line 158: TODO @1saf when you merge your work with the new web3 composable can you replace this.
```suggestion
    // TODO use web3 composable
    const getConfig = () => store.getters['web3/getConfig']();
```

src/components/cards/TradeCardGP/TradeCardGP.vue
```diff
@@ -264,25 +275,83 @@ export default defineComponent({
...
+              .integerValue(BigNumber.ROUND_DOWN)
+              .toString(),
+            tokenOutDecimals.value
+          ).toString();
+        } else {
+          sellAmount = parseUnits(
+            tokenInAmountScaled
+              .times(1 + slippageBufferRate.value)
+              .integerValue(BigNumber.ROUND_DOWN)
+              .toString(),
```
garethfuller, line 338: I'm not a fan of these massive functions. Could we not break it down a little so it's more readable? e.g.
```ts
async function trade() {
  if (isWrap.value) {
    await wrap();
  } else if (isUnwrap.value) {
    await unwrap();
  } else {
    await executeTrade();
  }
}
```
Doesn't have to be like this exactly or named like this, I just think it could be broken down a bit.

garethfuller (commented):
Looks good, just a few comments to discuss.

src/components/cards/TradeCardGP/TradeCardGP.vue
evgenyboxer: Yeah totally agree - I split the functions.

The code is likely to keep changing a lot tho, so this isn't the final one! Will try to keep functions smaller and more readable.

PR #448, opened 2021-06-22
Prevent GasReimbursement component crashing trading page on unreimbursed networks

TomAFrench:
Currently on any network where we aren't running the BAL4Gas program the trading page will crash.

I've added graceful failure for getting eligible assets to prevent this.

### DIFF
--- a/src/components/cards/TradeCard/GasReimbursement.vue
+++ b/src/components/cards/TradeCard/GasReimbursement.vue
@@ -51,7 +51,7 @@ export default defineComponent({
     const { t } = useI18n();
     const BAL = '0xba100000625a3754423978a60c9317c58a424e3d';
 
-    const eligibleAssetMeta = eligibleAssetList[appNetwork.networkName];
+    const eligibleAssetMeta = eligibleAssetList[appNetwork.networkName] ?? {};
     const eligibleAssets = Object.fromEntries(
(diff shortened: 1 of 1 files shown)


src/components/cards/TradeCard/GasReimbursement.vue
```diff
@@ -51,7 +51,7 @@ export default defineComponent({
     const { t } = useI18n();
     const BAL = '0xba100000625a3754423978a60c9317c58a424e3d';
 
-    const eligibleAssetMeta = eligibleAssetList[appNetwork.networkName];
+    const eligibleAssetMeta = eligibleAssetList[appNetwork.networkName] ?? {};
     const eligibleAssets = Object.fromEntries(
```
evgenyboxer, line 55: minor - what if we do a condition on the `eligibleAssets` instead? (just to save us from doing the computation which we know won't happen)

```
const eligibleAssets = eligibleAssetMeta == null ? {} : Object.fromEntries(...)`
```

evgenyboxer (approved):

src/components/cards/TradeCard/GasReimbursement.vue
TomAFrench: I think this effect would be so small to be unnoticeable tbh. In any case it increases computation on the happy path which will be much more common.

PR #449, opened 2021-06-22
Allow pulling token prices from chains other than Ethereum from CoinGecko

TomAFrench:
We now query prices for tokens on Coingecko for whichever network we are connected to.

### DIFF
--- a/src/services/coingecko.ts
+++ b/src/services/coingecko.ts
@@ -1,5 +1,5 @@
 import useConfig from '@/composables/useConfig';
-import { getNativeAssetId, getPlatformId } from './coingecko/api/price.service';
+import { getNativeAssetId, getPlatformId } from './coingecko/coingecko.service';
 
 function getChainAddress(chainId: number, address: string) {
   if (!address) {
--- a/src/services/coingecko.ts
+++ b/src/services/coingecko.ts
@@ -71,8 +71,8 @@ export type Prices = Record<string, Price>;
 export type HistoricalPrices = Record<string, Prices>;
 
 export async function getEtherPrice(): Promise<Price> {
-  const { networkConfig } = useConfig();
-  const nativeAssetId = getNativeAssetId(networkConfig.chainId);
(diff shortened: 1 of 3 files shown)


src/services/coingecko/api/price.service.ts
```diff
@@ -10,11 +10,25 @@ export type Price = { [fiat: string]: number };
...
+  '1': 'ethereum',
+  '42': 'ethereum',
+  '137': 'matic-network'
+};
+
+const platformIdMap = {
+  '1': 'ethereum',
+  '42': 'ethereum',
+  '137': 'polygon-pos'
+};
```
evgenyboxer, line 23: It seems that `nativeAssetIdMap` and `platformIdMap` is defined twice? could we reuse it?

TomAFrench: Yeah, we could reuse this. The duplication is a little bit of laziness from me as I assume we're going to be deleting one of these services as we finish the migration.

src/services/coingecko.ts
```diff
@@ -62,32 +65,48 @@ interface Price {
...
+const nativeAssetIdMap = {
+  1: 'ethereum',
+  42: 'ethereum',
+  137: 'matic-network'
+};
+
+const platformIdMap = {
+  1: 'ethereum',
+  42: 'ethereum',
+  137: 'polygon-pos'
```
evgenyboxer, line 77: How about if we have a function to resolve these mappings?

`function resolveNativeId(networkId: string | number) { ... }` // returns `ethereum`, `matic-network`, etc... (with a fallback to `ethereum` if the network id does not match)

`function resolvePlatformId(networkId: string | number) { ... }` // returns `ethereum`, `polygon-pos`, etc.. (with a fallback to `ethereum` if the network id does not match)

This way, if we are on Rinkeby it wouldn't fail.

evgenyboxer (approved):

PR #450, opened 2021-06-23
Remove useTokenLists and replace it with useTokenLists2

TomAFrench:
I've switched over the remaining places we use `useTokenLists` to alternatives allowing us to replace it with what was `useTokenLists2`.

I've pulled out the search functionality of `useTokenLists` and popped it in `SelectTokenModal` for now. I'd be happy to move this back into the new `useTokenLists` if desired.


garethfuller:
@TomAFrench I could be wrong but I don't think we're ready for this change. useTokenLists2 was specifically designed to work with a new way of fetching prices, balances and allowances that hasn't been fully implemented yet.

TomAFrench:
> I could be wrong but I don't think we're ready for this change. useTokenLists2 was specifically designed to work with a new way of fetching prices, balances and allowances that hasn't been fully implemented yet.

Ah ok, sorry I wasn't aware. Everything _looks_ ok though? What issues are we waiting to be resolved?

garethfuller:
Closing since registry refactor covers this.

PR #451, opened 2021-06-24
Bump SOR. Added default to V2 liquidity for non V1 case.

johngrantuk:
Fixes # .

- Bump SOR to Polygon supported version.
- Add option to set MaxPools and SwapCost via .env
- Set Liquidity selection to V2 by default if user currently has V1 manually set but V1 not supported.


TomAFrench:
LGTM. I could perform trades on this so all seems good.

TomAFrench (approved):

garethfuller (approved):

PR #456, opened 2021-06-25
Multi-network gas estimates

garethfuller:
- Add a GasPriceService which handles gas prices for multiple networks
- Remove old gas price utility function


garethfuller:
@evgenyboxer I'm going to merge this so that it gets tested today and over the weekend. But it would be good if you could have a scan at some point just to make sure you're ok with these changes since you implemented the original function.

PR #461, opened 2021-06-26
Disable/remove claim button on non Mainnet networks

evgenyboxer:
Claiming only works on Mainnet for now and so other networks shouldn't show it.

For now, it's completely removed - but might get a nicer design from Pon.

### DIFF
--- a/src/composables/useWeb3.ts
+++ b/src/composables/useWeb3.ts
@@ -43,7 +43,7 @@ export default function useWeb3() {
   });
 
   const isMainnet = computed(() => {
-    return userNetwork.value.name === 'Mainnet';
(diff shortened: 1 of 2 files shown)


src/composables/useWeb3.ts
```diff
@@ -43,7 +43,7 @@ export default function useWeb3() {
   });
 
   const isMainnet = computed(() => {
-    return userNetwork.value.name === 'Mainnet';
+    return appNetwork.name === 'Mainnet';
```
evgenyboxer, line 46: This function wasn't actually used so this change doesn't have much effect... I do think it should be checking for `appNetwork` instead of `userNetwork`.

garethfuller (approved):

PR #465, opened 2021-06-28
fix/update: fix circular dependency in useTokenLists. Seperate some o…

1saf:
There were some major issues with how I designed `useTokenLists` and how it was consumed.

One of the biggest problems was that `useAccountBalances` was consuming tokens from the vuex store, so I didn't notice this problem. After updating it to use `useTokenLists` to retrieve a list of all tokens so it could perform its query my shit decision making was out in full force. 

A circular dependency was born where `useAccountBalances` needed `useTokenLists` which also needed the former. Yikes.
Given this notion, I had to separate some concerns, I've updated `useTokenLists` to become `useTokenStore` whos responsibility is loading a list of tokens from the active token lists, injected tokens from the vuex store and ether. 

The `useTokens` composable is now responsible for all the balance and other metadata which was being applied to the map of tokens returned by what is now `useTokenStore`. 

This change means that composables and other behaviour which just need a list of tokens without all the balance metadata can do so without any risk of circular deps.

**What about the registry refactor?**
This doesn't affect the registry refactor, since it is just an issue with the existing composables. If you look at the places I've updated the usages of `tokenDictionary` you'll see that all I've had to do is update the composable it is using and the variable being accessed.

With the registry refactor, it will be pretty much the same, where the usages will get updated and that'll be it.


garethfuller (approved):

garethfuller:
Great that you noticed before it went into `develop`! 🙌

evgenyboxer (approved):

PR #472, opened 2021-06-28
feat: add fallback method of determining whether a tx has confirmed

TomAFrench:
This is a potential stopgap solution to deal with the lack for Blocknative support on Polygon. In parallel with using Blocknative we manually listen for the provided transaction hashes being confirmed using ethers, if we see this then we can fire the callback. It doesn't handle failures/cancellations nicely but these would be much rarer in comparison.

I've added a 2 second delay to begin looking for the tx hashes as otherwise the RPC often hasn't seen the transaction yet and returns null.


TomAFrench:
Closed in favour of another planned implementation.

PR #478, opened 2021-06-29
Handle tx events on polygon

garethfuller:
- Add a useEthers composable with simple tx listener function similar to our blocknative equivalent.
- If network is polygon handle tx confirmations with ethers tx listener.

This is a stop-gap solution until we role our own tx handler which is also needed to support the Gnosis integration.

### DIFF
--- a/src/components/forms/pool_actions/InvestForm.vue
+++ b/src/components/forms/pool_actions/InvestForm.vue
@@ -320,10 +320,10 @@ export default defineComponent({
     // COMPOSABLES
     const store = useStore();
     const { isAuthenticated } = useAuth();
-    const { account, appNetwork, userNetwork, isPolygon } = useWeb3();
+    const { account, appNetwork, userNetwork } = useWeb3();
     const { fNum, toFiat } = useNumbers();
     const { t } = useI18n();
-    const { txListener } = useNotify();
+    const { txListener, supportsBlocknative } = useNotify();
     const { minusSlippage } = useSlippage();
     const { allTokens } = useTokens();
     const { trackGoal, Goals } = useFathom();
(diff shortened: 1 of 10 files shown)


evgenyboxer (approved):

src/components/forms/pool_actions/InvestForm.vue
```diff
@@ -566,19 +600,11 @@ export default defineComponent({
...
-            data.loading = false;
-          },
-          onTxCancel: () => {
-            data.loading = false;
-          },
-          onTxFailed: () => {
-            data.loading = false;
-          }
-        });
+        if (isPolygon.value) {
```
TomAFrench, line 603:
```suggestion
        if (!supportsBlocknative.value) {
```
Can we make this more generic? We'll possibly be on other networks which don't have block stove support soon. I'd probably swap the two branches as well if we make this change.

TomAFrench (commented):
Looks good. Only one change I would make but we can trust that up later if we just want to get this in asap.

src/components/forms/pool_actions/InvestForm.vue
garethfuller: Good point! https://github.com/balancer-labs/frontend-v2/pull/478/commits/6911dbe65b6f2244ca3569e84c5b991d66e89664

src/composables/trade/useTokenApproval.ts
```diff
@@ -90,14 +93,22 @@ export default function useTokenApproval(tokenInAddress, amount, tokens) {
...
-  function approvalTxListener(hash: string) {
+  function txHandler(tx: TransactionResponse): void {
+    if (supportsBlocknative.value) {
+      blocknativeTxHandler(tx.hash);
+    } else {
+      ethersTxHandler(tx);
+    }
+  }
+
+  function blocknativeTxHandler(hash: string): void {
```
TomAFrench, line 111: This is inconsistent with what we have in the `InvestForm` where both handlers receive a `TransactionResponse` and BN pulls out the hash internally.

src/composables/useBlocknative.ts
```diff
@@ -1,12 +1,23 @@
+import { computed } from 'vue';
```
TomAFrench, line 1:
```suggestion
import { computed, injected } from 'vue';
```

Can merge imports here

PR #509, opened 2021-07-02
[Gnosis] - New preview screen and integrated Balancer trading

evgenyboxer:
This PR introduces the new preview screen and integrates Balancer trading into the interface.

I've created a new `useTrading.ts` composable which under the hood uses `useSor.ts` and `useGnosis.ts` composables (this could allow us to integrate more trading sources in the future)

Trading works for both gnosis/balancer (wrapping and unwrapping too). However, there is no tracking of transactions - so that's the next task.

I tried to keep the changes to `useSor.ts` minimal so that it could work well with `useTrading.ts`.

To test locally:

```
VUE_APP_NETWORK=1 VUE_APP_GNOSIS_INTEGRATION=true npm run serve
```

Or here - https://gnosis-git-evgeny-ui-580-implement-new-preview-screen-balancer.vercel.app/

What to expect:
ETH -> ERC20 (Balancer trade) - sending a tx (approvals are not required)
ERC20 -> ERC20 (Gnosis trade) - signing a message (approvals required)


src/App.vue
```diff
@@ -18,13 +18,18 @@
...
 import AppNav from '@/components/navs/AppNav/AppNav.vue';
 import AppHero from '@/components/heros/AppHero.vue';
 import WalletSelectModal from '@/components/web3/WalletSelectModal.vue';
-import useVueWeb3 from './services/web3/useVueWeb3';
+import useVueWeb3 from '@/services/web3/useVueWeb3';
 import RpcProviderService from '@/services/rpc-provider/rpc-provider.service';
-import { useRoute } from 'vue-router';
+import { DEFAULT_TOKEN_DECIMALS } from './constants/tokens';
+
+BigNumber.config({ DECIMAL_PLACES: DEFAULT_TOKEN_DECIMALS });
```
evgenyboxer, line 32: This was done to mitigate any underflow issues when using bignumber and then trying to pass the result to ethers BigNumber

src/components/_global/BalAlert/BalAlert.vue
```diff
@@ -65,9 +65,9 @@ export default defineComponent({
     const bgColorClass = computed(() => {
       switch (props.type) {
         case 'warning':
-          return 'bg-yellow-100 border border-yellow-500 text-black';
+          return 'bg-yellow-50 text-black';
         case 'error':
-          return 'bg-red-100 border border-red-500 text-black';
+          return 'bg-red-50 text-black';
```
evgenyboxer, line 70: ^ to match new design

src/components/cards/TradeCard/TradeCard.vue
```diff
@@ -288,6 +290,11 @@ export default defineComponent({
       handleAmountChange();
     });
 
+    watch(liquiditySelection, () => {
+      // When the liquidity type is changed we need to update the trade to use that selection
+      handleAmountChange();
+    });
```
evgenyboxer, line 296: ported from `useSor`

src/components/cards/TradeCard/TradeCard.vue
```diff
@@ -192,22 +195,21 @@ export default defineComponent({
...
+      tokenInAddressInput: tokenInAddress,
+      tokenInAmountInput: tokenInAmount,
+      tokenOutAddressInput: tokenOutAddress,
+      tokenOutAmountInput: tokenOutAmount,
       tokens,
-      allowanceState,
       isWrap,
       isUnwrap
-    );
+    });
```
evgenyboxer, line 212: these are minimal changes to allow passing an object since there were too many params.

src/components/cards/TradeCard/TradeCard.vue
```diff
@@ -181,7 +184,7 @@ export default defineComponent({
     });
 
     // COMPOSABLES
-    const { allowanceState, isLoading: isLoadingApprovals } = useTokenApproval(
+    const { isLoading: isLoadingApprovals } = useTokenApproval(
```
evgenyboxer, line 187: allowanceState wasn't used

src/components/popovers/TradeSettingsPopover.vue
```diff
@@ -24,10 +24,7 @@
       </div>
       <AppSlippageForm class="mt-1" />
     </div>
-    <div
-      v-if="appTradeInterface === TradeInterface.BALANCER && !hideLiquidity"
-      class="mt-6"
-    >
+    <div v-if="!hideLiquidity" class="mt-6">
```
evgenyboxer, line 27: this is just re-enabling previously disabled behaviour... it should be possible to select liquidity source when using the new trade interface (since it integrates Balancer too)

src/services/config/config.service.ts
```diff
@@ -31,3 +31,5 @@ export default class ConfigService {
     return configs[key];
   }
 }
+
+export const configService = new ConfigService();
```
evgenyboxer, line 35: I don't think we ever need more than one instance of `ConfigService` right? so I just init it once here and export it for usage.

src/components/_global/icons/ArrowDownIcon.vue
```diff
@@ -0,0 +1,14 @@
+<template>
+  <svg
```
1saf, line 2: Can we use the feather icons compared to this? or are we moving away from feather @garethfuller

src/components/modals/TradePreviewModalGP.vue
```diff
@@ -1,70 +1,189 @@
...
-          :size="30"
-          :width="55"
-        />
-        <div class="flex flex-col">
-          <div class="font-bold">
-            {{ fNum(amountIn, 'token') }} {{ symbolIn }} ->
-            {{ fNum(amountOut, 'token') }} {{ symbolOut }}
+      <BalCard noPad class="relative mb-6">
+        <template v-slot:header>
+          <div class="w-full p-3 border-b bg-gray-100 rounded-t-lg text-sm">
```
garethfuller, line 6: There is this weird gap between the card border and the contents:
Screenshot 2021-07-05 at 19 06 36
One option is to remove the internal rounded border and hide any overflow.
```suggestion
      <BalCard noPad class="relative mb-6 overflow-hidden">
        <template v-slot:header>
          <div class="w-full p-3 border-b bg-gray-100 text-sm">
```

src/components/modals/TradePreviewModalGP.vue
```diff
@@ -1,70 +1,189 @@
...
+            {{ $t('effectivePrice') }}
+            {{
+              trading.exactIn.value
+                ? trading.effectivePriceMessage.value.tokenIn
+                : trading.effectivePriceMessage.value.tokenOut
+            }}
+          </div>
```
garethfuller, line 13: I feel like this effective price header could be a bit more subtle and the label could use a bit of visual distinction from the amounts. From this:
Screenshot 2021-07-05 at 19 11 32
to this:
Screenshot 2021-07-05 at 19 11 14
```suggestion
      <BalCard noPad class="relative mb-6">
        <template v-slot:header>
          <div class="w-full p-3 border-b bg-gray-50 rounded-t-lg text-xs">
            <span class="text-gray-500">
              {{ $t('effectivePrice') }}
            </span>
            {{
              trading.exactIn.value
                ? trading.effectivePriceMessage.value.tokenIn
                : trading.effectivePriceMessage.value.tokenOut
            }}
          </div>
```

src/components/modals/TradePreviewModalGP.vue
```diff
@@ -1,70 +1,189 @@
...
+                {{ $t('tokens') }}
+              </div>
+              <div
+                :class="showSummaryInFiat && 'text-blue-600 font-medium'"
+                class="pl-1 cursor-pointer"
+                @click="showSummaryInFiat = true"
+              >
+                {{ FiatCurrency.USD }}
+              </div>
+            </div>
```
garethfuller, line 89: These are essential text btns (could potentially be a Bal component) and so it kind of makes sense to uppercase them both and have font-medium in all states, so they don't look like regular text at all. The other reason for keeping font-medium in both states is so you don't get the small shift when you select one or the other.

Finally, I'm really not a fan of having multiple class attributes, so I've cleaned that up.
```suggestion
            <div class="flex divide-x text-xs uppercase">
              <div
                :class="[
                  'pr-1 cursor-pointer font-medium',
                  { 'text-blue-600': !showSummaryInFiat }
                ]"
                @click="showSummaryInFiat = false"
              >
                {{ $t('tokens') }}
              </div>
              <div
                :class="[
                  'pl-1 cursor-pointer font-medium',
                  { 'text-blue-600': showSummaryInFiat }
                ]"
                @click="showSummaryInFiat = true"
              >
                {{ FiatCurrency.USD }}
              </div>
            </div>
```

src/components/modals/TradePreviewModalGP.vue
```diff
@@ -1,70 +1,189 @@
...
+          Requires {{ trading.requiresApproval.value ? 2 : 1 }}
+          {{ trading.requiresApproval.value ? 'transactions' : 'transaction' }}:
         </div>
-        <div>
+        <div class="text-sm">
           <div
-            v-if="requiresApproval"
-            class="p-3 flex items-center border rounded-lg"
+            v-if="trading.requiresApproval.value"
+            class="flex items-center mb-2"
```
garethfuller, line 153: I think the requires transactions section should be the same style as the summary, no colon and a line break.
Screenshot 2021-07-05 at 19 24 07

garethfuller (changes requested):

src/components/_global/icons/ArrowDownIcon.vue
evgenyboxer: I took the one Pon used in his designs - it's more "pointy" than the feathers one... I do think we should stick to whatever Pon uses in his designs.

src/composables/trade/useGnosis.ts
```diff
@@ -0,0 +1,241 @@
...
+  tokenOutAmountInput,
+  tokenInAmountScaled,
+  tokenOutAmountScaled,
+  tokenIn,
+  tokenOut
+}: Props) {
+  // COMPOSABLES
+  const store = useStore();
+  const { account, getSigner } = useVueWeb3();
+  const signer = getSigner();
```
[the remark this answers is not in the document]
evgenyboxer: You are right! I have moved it to be called on the trade function (similar to how we call to get the signer before sending a transaction)

src/composables/trade/useGnosis.ts
```diff
@@ -0,0 +1,241 @@
...
+    errors.value = {
+      feeExceedsPrice: false
+    };
+  }
+
+  function resetFees() {
+    feeQuote.value = null;
+  }
+
+  function getQuote() {
```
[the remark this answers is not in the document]
evgenyboxer: Yeah, good call. Added.

evgenyboxer:
@garethfuller Fixed all issues. Thanks for the review.

Screen Shot 2021-07-06 at 12 11 48 pm

garethfuller (approved):

PR #521, opened 2021-07-04
Fix isV1Supported

acryptosx:
It was doing the opposite of what it was supposed to do.


garethfuller (approved):

garethfuller:
@acryptosx Good spot, thanks for this! 🙌

PR #523, opened 2021-07-05
feature: EIP 3085 Add + switch network

1saf:
Fixes # .

- Add prompt to switch to polygon when connecting wallet on polygon.


src/services/web3/utils/helpers.ts
```diff
@@ -0,0 +1,44 @@
+import { ExternalProvider, Web3Provider } from '@ethersproject/providers';
+import ConfigService from '@/services/config/config.service';
+
+export async function importPolygonDetailsToWallet(provider: ExternalProvider) {
+  const configService = new ConfigService();
+  const polygonNetworkConfig = configService.getNetworkConfig('137');
```
1saf, line 6: should I pull this chainId '137' out of the configs constant?

src/services/web3/web3.plugin.ts
```diff
@@ -122,6 +125,21 @@ export default {
           lsSet('connectedWallet', account.value);
           lsSet('connectedProvider', wallet);
           pluginState.walletState = 'connected';
+
+          // special case for polygon
+          if (
+            configService.env.NETWORK === '137' &&
```
1saf, line 131: same configs constant comment as above. 
also im having second thoughts about pulling this if statement logic out of the connect wallet function, thoughts?

src/services/web3/utils/helpers.ts
evgenyboxer: There is actually a `Network.POLYGON` (`network.ts`) which you can use. I guess I should change the name to something like `ChainId.POLYGON` so it will be clearer that its a chain id.

garethfuller (commented):
Looks good! Can we have it prompt you to switch networks for the other networks too? e.g. If you try to connect on Mainnet but your current network is kovan?

src/services/web3/web3.plugin.ts
```diff
@@ -13,6 +13,8 @@ import { WalletLinkConnector } from './connectors/walletlink/walletlink.connecto
 import { PortisConnector } from './connectors/portis/portis.connector';
 import useFathom from '@/composables/useFathom';
 import getProvider from '@/lib/utils/provider';
+import ConfigService from '../config/config.service';
```
garethfuller, line 16: If you merge in `develop` you should be able to do this:
```suggestion
import { configService } from '../config/config.service';
```
And then not have to initiate it with `new` below.

garethfuller (changes requested):

src/services/web3/utils/helpers.ts
```diff
@@ -0,0 +1,69 @@
+import { ExternalProvider } from '@ethersproject/providers';
+import ConfigService from '@/services/config/config.service';
+
+const configService = new ConfigService();
```
garethfuller, line 4: Sorry, missed this one.

garethfuller (approved):

PR #537, opened 2021-07-08
Migrate composables to a provider pattern, so watchers and composables register once.

1saf:
# Description
The current issue with a lot of token list lag was the fact that our composables were called many times. Given the fact that these composable instances are isolated, they would register their own watchers and computed properties multiple times as well. 
Some of of these watchers and composables are quite hefty in the compute they perform, so it would essentially become a very expensive render every tim.
This PR aims to refactor that logic and move the more used composable contents to a provider pattern, so they are provided once (watchers and computables) registered once and a much needed speed improvement is introduced.

Fixes #TokenListsLagging

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] Other

## How should this be tested?
Entire app, make sure invest works, withdraw works (balances update when so). Trading works (balances update)

## Checklist:

- [ WIP] I have performed a self-review of my own code
- [X ] I have commented my code where relevant, particularly in hard-to-understand areas
- [ X] My changes generate no new console warnings

### DIFF
--- a/src/AppProviders.vue
+++ b/src/AppProviders.vue
@@ -1,0 +1,42 @@
+<script lang="ts">
+import { defineComponent, h } from 'vue';
+
+import Provider from '@/plugins/providers/Provider.vue';
+import provideTokenStore from '@/plugins/providers/tokenstore.provider';
+import provideTokens from '@/plugins/providers/tokens.provider';
+import provideAccountBalances from '@/plugins/providers/balances.provider';
+import provideAccountAllowances from '@/plugins/providers/allowances.provider';
+
+const providers = [
+  provideTokenStore,
+  provideAccountBalances,
+  provideTokens,
+  provideAccountAllowances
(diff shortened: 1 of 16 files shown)


src/AppProviders.vue
```diff
@@ -0,0 +1,42 @@
+<script lang="ts">
+import { defineComponent, h } from 'vue';
+
+import Provider from '@/plugins/providers/Provider.vue';
+import provideTokenStore from '@/plugins/providers/tokenstore.provider';
+import provideTokens from '@/plugins/providers/tokens.provider';
+import provideAccountBalances from '@/plugins/providers/balances.provider';
+import provideAccountAllowances from '@/plugins/providers/allowances.provider';
```
1saf, line 8: These probably will move to a better home that isn't inside plugins

PR #558, opened 2021-07-13
Pk UI

pkattera:
# Description

- Typography updates to enable the variable Inter font to work across more devices using `font-variation-settings`
- Fluid tables that stretch full width on smaller screens.
- Updated the pool composition icons in the Invest tables so that the z-index decreases for any additional pool token (resulting in the first token appearing on top instead of at the bottom).
- Updating the external link icon to be consistent across the site
- Replaced the `+` and `-` icons in the pool transactions table to be cleaner and not feel like actionable buttons.
- Updated the column order of the pool transaction table, so that the `details` appears before the value.
- Minor UI tweaks to the filter section above the investment pools to make the padding more consistent.


src/components/_global/BalAsset/BalAssetSet.vue
```diff
@@ -55,3 +56,16 @@ export default defineComponent({
   }
 });
 </script>
+
+<style scoped>
+.token-set {
+  @apply absolute;
+  box-shadow: 0 0 0 1.5px #ffffff;
```
garethfuller, line 63: imo it looks better without these borders, it's already quite intricate visually, I feel like the borders over complicate it.

src/components/_global/BalAsset/BalAssetSet.vue
```diff
@@ -55,3 +56,16 @@ export default defineComponent({
...
+  border-radius: 50%;
```
garethfuller, line 64: I don't think this is needed?

src/components/_global/BalAsset/BalAssetSet.vue
```diff
@@ -55,3 +56,16 @@ export default defineComponent({
...
+}
+
+.dark .token-set {
+  @apply bg-gray-850;
+  box-shadow: 0 0 0 1px #162031;
```
garethfuller, line 69: Same comment about borders above.

src/components/_global/BalTable/BalTable.vue
```diff
@@ -36,7 +36,7 @@
               ></slot>
               <div v-else>
                 <h5
-                  class="text-base font-semibold text-gray-800 dark:text-gray-100"
+                  class="text-base text-gray-800 dark:text-gray-100"
```
garethfuller, line 39: Please fix lint issue.

src/components/tables/PoolsTable/PoolsTable.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <BalCard shadow="lg" class="mt-4" no-pad>
+  <BalCard shadow="lg" class="mt-4" no-pad square>
```
garethfuller, line 2: This should be dependent on the breakpoint. I don't think the table should be max width and square on screen sizes above mobile. For example on my 15" macbook there should be padding around the tables as it was before.

This is what it looks like full screen for me now.
Screenshot 2021-07-13 at 14 42 13

src/pages/Home.vue
```diff
@@ -1,29 +1,34 @@
 <template>
-  <div class="container mx-auto px-4 lg:px-0 pt-10 md:pt-12">
+  <div class="invest-container mx-auto pt-10 md:pt-12">
```
garethfuller, line 2: I really think we should stick to the tailwind `container` class here as we're using it on the pool page too. Keeps everything consistent.

src/pages/Pool.vue
```diff
@@ -1,11 +1,11 @@
 <template>
-  <div class="container mx-auto px-4 lg:px-0 pt-8">
+  <div class="pool-container lg:px-4 mx-auto pt-8">
```
garethfuller, line 2: Again, I think we should stick to the default `container` class. Can you explain the reasoning behind this change?

garethfuller (changes requested):

src/components/_global/BalAsset/BalAssetSet.vue
pkattera: Creating separation like this is a fairly common design pattern when overlapping images. It helps people to visually distinguish the different tokens. 

For example, see how Instagram does this within their mobile app on overlapping images which are even smaller than our token logos.

IMG_B3DACB92D53B-1


IMG_EF8D69887255-1 2

src/components/_global/BalAsset/BalAssetSet.vue
pkattera: The border-radius is only really needed for the unknown token symbol which actually appears to be a square (which also randomly appears to be larger than the other tokens...). Because it's a square, the border-radius is required if we want to do the border separation technique.

src/components/_global/BalAsset/BalAssetSet.vue
pkattera: As above.

src/components/tables/PoolsTable/PoolsTable.vue
pkattera: This was intentional but I think we can do better. Basically, if the table can be scrolled horizontally, it definitely needs to extend to the right hand side edge of the screen. Otherwise, people may not realize that there is additional content, and scrolling horizontally within a container feels weird. 

I'll try adding left padding to the tables, but I definitely want to extend the table to the right hand edge until it reaches it's full width.

src/pages/Home.vue
pkattera: The Tailwind container class has a bunch of max-widths, like this

Screen Shot 2021-07-13 at 10 46 09 AM

This ends up creating a lot of weird layouts like this:

Screen Shot 2021-07-13 at 10 45 49 AM

This is a really bad use of screen real estate, especially when we have a horizontally scrolling table which really should extend to the right edge of the screen. We're much better off having true responsive, fluid layouts with side padding which use the full screen space instead of having arbitrary max-widths.

src/pages/Pool.vue
pkattera: As above, I'd prefer fluid layouts with one final max-width than one with arbitrary max-widths across different breakpoints.

src/components/_global/BalAsset/BalAssetSet.vue
garethfuller: Ahh yeah this is makes more sense now, but we should be using the tailwind colors and also adjusting those colors on row hover so it matches the hover background color. If you need help with that lmk.

src/components/_global/BalAsset/BalAssetSet.vue
garethfuller: Ahh ok, yeah the child components handle this as far as I'm aware with the `rounded-full` class but if we need it at this level too then let's use that.

src/components/_global/BalTable/BalTable.vue
garethfuller: If you scroll threw the [files view ](https://github.com/balancer-labs/frontend-v2/pull/558/files) of the PR you'll see this big warning:
Screenshot 2021-07-13 at 19 39 23

src/components/tables/PoolsTable/PoolsTable.vue
garethfuller: I don't think we're on the same page here. On my laptop the table extends the full width of the page when there is no need. Currently on production it's like this for me (full screen):

Screenshot 2021-07-13 at 19 42 43

Which is correct, the table should only go full-width when the screen size prevents the whole table from being seen. We just need to use the breakpoints correctly here. I can help with this.

src/pages/Home.vue
garethfuller: Ok, I don't think this layout is necessarily to do with the container class, let me play around with this too.

src/components/_global/BalAsset/BalAssetSet.vue
pkattera: Done! I initially preferred to use a box-shadow instead of a border to retain the with and height of the icons (instead of losing a couple of pixels to an inner border... Anyway, it should be good now.

src/components/_global/BalAsset/BalAssetSet.vue
pkattera: Done

src/components/_global/BalTable/BalTable.vue
pkattera: This should be fixed

src/components/tables/PoolsTable/PoolsTable.vue
pkattera: I've done another version which I think works better. The tables start full width of mobile, then add left padding on small screens upwards. Once it hits `lg`, I place padding on both sides. Feels pretty fluid now.

src/pages/Home.vue
```diff
@@ -134,6 +134,6 @@ export default defineComponent({
 
 <style scoped>
 .invest-container {
-  @apply w-full max-w-screen-2xl;
+  @apply w-full max-w-screen-xl;
```
pkattera, line 137: How do you feel about reducing the max-width of the content sections on the home page? The tables feel too wide at the moment and I think this smaller layout keeps things compact and tidy.

src/components/_global/BalAsset/BalAssetSet.vue
garethfuller: hmm I think this is only useful if you don't use shadows under the icons. See below:

Screenshot 2021-07-14 at 10 03 55

With shadows it doesn't look like they're cut out of each other like it does on instagram, I think this is what I was noticing. I would suggest we either we have shadows with no borders or borders with no shadows, and if we go for borders then I'd probabl increase the width to make it more distinctive:

Screenshot 2021-07-14 at 10 08 32

src/components/tables/PoolsTable/PoolsTable.vue
garethfuller: Hey, so Tailwind gives a suggestion of how to use the container class with a breakpoint so it feels fluid [here](https://tailwindcss.com/docs/container#responsi). I've updated to use that in this [commit](https://github.com/balancer-labs/frontend-v2/pull/558/commits/395c3dea2e399bec879d7d955ef247b71d8c0d81).

I also don't think we need to go to absolute full width and square for the tables on mobile. I think we can maintain the border radius and just have a much smaller padding. I've included that in the above commit too, let me know what you think. I feel like if we want to have absolute full width for tables on mobile we need to remove border radius, border colors, and shadows and actually design a full screen experience that feels more like a mobile app.

src/pages/Home.vue
garethfuller: This is a duplicate discussion about containers, will resolve in favour of above thread.

src/pages/Home.vue
garethfuller: I think we should stick to the container class as I've done in the last commit, and mentioned above. That way the home and pool pages are consistent. I guess we could change the container dimensions across the board if needed. However, the tables don't feel that wide to me on the home page tbh and it gives us room to add more context/data later if needed.

src/components/_global/BalAsset/BalAssetSet.vue
pkattera: Happy to remove the box-shadows.

I agree with increasing the borders, but the problem is that increasing the border size reduces the token image size since css borders are inner borders. This is why I was initially using box-shadows to act as outer borders but then we lose tailwind's color system... Maybe we could increase the size of the token images here by 2px, and then have `border-2` set on them?

garethfuller (approved):

PR #565, opened 2021-07-14
Cleanup old web3

1saf:
# Description
Now that the web3 refactor is complete, we should fix up all the naming and any remaining dangling references. This PR removes the web3 vuex module and the old composable files. Also cleaned up any dangling references to the old code.

Fixes # (issue)

## Type of change

- [X] Other

## How should this be tested?

Just give a general sense check to the app. Only names have changed here, and files.

## Checklist:


garethfuller (changes requested):
Thanks for this! Couple of things still need to be done I think:

- Do a code search for `.web3` and you'll see that there are still some references to the web3 vuex module.
- Also, the `account` vuex module relies on it quite a lot, can we get rid of that yet?
- Can we get rid of `components/modals/AccountModal.vue`?
- Can we get rid of `constants/connectors.json`?
- Can we get rid of `plugins/authOptions.ts` and all the images imported in that file?

1saf:
> Thanks for this! Couple of things still need to be done I think:
> 
> * Do a code search for `.web3` and you'll see that there are still some references to the web3 vuex module.
> * Also, the `account` vuex module relies on it quite a lot, can we get rid of that yet?
> * Can we get rid of `components/modals/AccountModal.vue`?
> * Can we get rid of `constants/connectors.json`?
> * Can we get rid of `plugins/authOptions.ts` and all the images imported in that file?

`authOptions.ts` is actually used for logos and names
`connectors.json` is kill
`AccountModal` is kill
`account.ts` is kill

garethfuller (commented):
@1saf If we're still using the functions in authOptions to get the logos I think we should probably refactor that so it's part of the web3 service right?

1saf:
All resolved @garethfuller

garethfuller (approved):

PR #569, opened 2021-07-14
[WIP] Display liquidity mining amounts

markusbkoch:
# Description

* Add a column to the pools table to display the amount of tokens received in liquidity mining
* Filters the subgraph query so as to only fetch incentivized pools

This is just a rough draft, mostly to help check the weekly JSON and automate production of a graphic for announcements. But at some point we might want to have something like this on its own page within the app, or maybe as a filter option?

## Know issues / TODO

- [ ] The first time the page loads, the table comes up empty. Select a token in the filter section, then unselect it, and the results are as expected.
- [ ] Prettify the new column (smaller font size? one token per line?)
- [ ] Parameterize so as to move the feature to a new page or hide it behind a filter option

## Type of change

- [ ] Other

## Checklist:


TomAFrench:
Would this work better if it showed up inside the liquidity mining hover popup rather than be in the main table?

markusbkoch:
> Would this work better if it showed up inside the liquidity mining hover popup rather than be in the main table?

For the main "invest" page, yes, I agree.

This version with the additional column could be its own page (eg `/liquidity-mining` or `/incentivized-pools`) if we think it's worth the effort; or just forever in a vercel preview that we use to:
1. double check the weekly JSON proposals; and 
1. screenshot to use in the weekly announcements (once the new column's design is nicer). 

@1saf is helping with (2) 🙏

markusbkoch:
Handled #615

PR #583, opened 2021-07-20
fix: undefined errors and unused vuex store

1saf:
# Description
Fixes the undefined errors from development

Fixes # (issue)

## Type of change

- [ ] Other

## How should this be tested?
Sense check the app

## Checklist:

### DIFF
--- a/src/store/modules/registry.ts
+++ b/src/store/modules/registry.ts
@@ -85,7 +85,7 @@ const getters = {
       return token;
     });
 
-    if (rootState.web3.account) {
+    if (rootState?.web3?.account) {
       tokens = tokens.map(token => {
         const address = token.address.toLowerCase();
         token.balanceDenorm = rootState.account.balances[address] || '0';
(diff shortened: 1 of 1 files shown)


src/store/modules/registry.ts
```diff
@@ -209,8 +209,6 @@ const actions = {
       if (meta) injected.push({ ...meta, injected: true });
     });
     commit('setInjected', injected);
-    await dispatch('account/getBalances', null, { root: true });
-    await dispatch('account/getAllowances', { tokens }, { root: true });
```
garethfuller, line 213: If we're still using this function, will we not still need to fetch balances/allowances when new tokens are added? Or is that happening elsewhere?

1saf: that is happening elsewhere and is reactive

1saf: as even if we fetch balances and allowances via the account store, the values aren't actually used from the store

garethfuller (approved):

PR #588, opened 2021-07-20
Add Gnosis Safe connector

TomAFrench:
# Description

This PR adds a new Gnosis Safe connector which allows loading the interface through the Gnosis website to directly connect to a Safe.

## Type of change

- [ ] Other

## How should this be tested?

The main issue which needs to be tested is that we correctly pick up when a transaction confirms when using a Gnosis safe with only one signer. This is because we don't talk directly with the RPC but through a Gnosis service which sometimes can not know about a transaction that's in flight for some time.

Testing that you can do a single transaction and not have the app get stuck in an infinite load should be sufficient but we should probably test each transaction type just to be thorough.

You can load the app on either [mainnet](https://gnosis-safe.io) or [polygon](https://polygon.gnosis-safe.io)

To open as a safe app:
- Load a Safe on https://polygon.gnosis-safe.io and go to the Apps tab.
- Click the first tile on that page which allows loading a custom safe app.
- Set the most recent preview URL as the URL and the name+icon should autofill. Click add.
- You'll be taken to the app running in an iframe (It's possible that your wallet will autoload at this point, just log out if so)
- Connect again using the new option with the name "Gnosis Safe".

**If you're trying to run this on mainnet then you need to run the app locally as passwords on Vercel prevent it from loading properly**

Known issues:
- ~~Transactions in the "recent activity" menu use the "safeTxHash" rather than the real hash so the links are broken.~~ Addressed
- ~~It's possible to connect to metamask on the Gnosis interface so we should detect this case and prevent connecting to other wallets.~~ Addressed

## Checklist:


garethfuller (changes requested):
Thanks for taking this on again Tom! A few comments:

- Could we move `public/logo512.png` into `public/images/logos/` just to keep the public folder organised going forward.
- I don't think it's possible to test on mainnet atm via the preview URLs due to password protection. Not sure if we need remove that temporarily?
- Transactions stayed pending indefinitely in the notification system, the UI however seems to recognise confirmation and updates. cc @evgenyboxer

TomAFrench:
@evgenyboxer I've integrated your changes to correct the transaction hash when using Gnosis while disabling the link on pending transactions. I feel this is at the point where it's ready to be merged so we can do some initial community testing, could you review?

evgenyboxer (approved):

src/components/navs/AppNav/AppNavSettings.vue
```diff
@@ -221,6 +221,7 @@ export default defineComponent({
     const connectorName = computed(() => getConnectorName(connector.value.id));
 
     const connectorLogo = computed(() => getConnectorLogo(connector.value.id));
+    const hideDisconnect = computed(() => connector.value.id == 'gnosis');
```
evgenyboxer, line 224: I've seen that in other instances you have used
```
connector.value?.id === 'gnosis'
```

should this be the case here too?

TomAFrench: It's not really necessary as I think these components don't get rendered until we have a connector (In other instances I was getting crashes due to trying to read from a null connector.) It doesn't hurt though.

TomAFrench:
@garethfuller can you sign off on this? I can't merge without a rereview

garethfuller (approved):
Looks good, sorry for the delay.

TomAFrench:
No worries, this got pushed down the priority list so I didn't even notice it was still locked until this morning.

PR #592, opened 2021-07-21
Hotfix/1.10.2 providers

1saf:
# Description

Based off from https://github.com/balancer-labs/frontend-v2/pull/537


src/providers/Provider.vue
```diff
@@ -0,0 +1,19 @@
+<template>
```
garethfuller, line 1: Let's delete this file now.

garethfuller (changes requested):
Could you convert this to a regular PR into develop? It's quite a big change that will need some proper testing.

Also there seems to be a linting issue that needs to be fixed.

PR #629, opened 2021-08-04
Cleanup dependencies

garethfuller:
# Description

There were quite a few unused packages so I've removed them.

## Type of change

- [x] Other

## How should this be tested?

- [ ] Generally test app functionality.

## Checklist:


package.json
```diff
@@ -24,18 +24,6 @@
...
-    "@ethersproject/address": "^5.0.1",
-    "@ethersproject/bignumber": "^5.2.0",
-    "@ethersproject/bytes": "^5.0.5",
-    "@ethersproject/constants": "^5.0.1",
-    "@ethersproject/contracts": "^5.0.1",
-    "@ethersproject/hash": "^5.0.6",
-    "@ethersproject/providers": "^5.0.12",
-    "@ethersproject/strings": "^5.0.5",
-    "@ethersproject/units": "^5.0.1",
-    "@ethersproject/wallet": "^5.0.1",
```
garethfuller, line 38: These are not needed if you import the main package. Any references to them in the code still works.

evgenyboxer: Awesome... I think we can just import everything from `ethers` then right?
So like instead of:
`import { formatUnits } from '@ethersproject/units';`
You would write:
`import { formatUnits } from 'ethers/lib/utils';`

evgenyboxer (approved):

package.json
garethfuller: Yes but it seems you can still use the @ethersproject syntax with just the main `ethers` package installed. What would be the benefit in going through the project and replacing all the imports?

PR #644, opened 2021-08-09
Steth wrapping

TomAFrench:
# Description

This PR adds support for us to allow wrapping/unwrapping of stETH in the trade interfaces similar to how we do for WETH. This is important for us to have as we're currently planning on requiring users to wrap their stETH before adding it as liquidity to a pool.

@evgenyboxer I've left in your overrides to always require the batch relayer although these need to be removed to test wrapping.

## Type of change

- [ ] Other

## How should this be tested?

This can be tested locally by removing the lines which enforce all trades to use the batch relayer and then

* attempt a swap from stETH to wsETH and vice versa.
* attempt a non-wrapping swap.
* ensure that attempting to trade from ETH to wsETH is not attempted as a wrap.

## Checklist:


evgenyboxer:
Thanks, @TomAFrench!

PR #660, opened 2021-08-11
Add support for EIP1559

evgenyboxer:
# Description

Adds support for EIP1559 type transactions.

Here is an example of a tx I made with the change - https://etherscan.io/tx/0xf2d62b204da7f2c0500fb0844821c627ca177a3d8c9cb15e1402be2041c38dd3

Fixes # (issue)

## Type of change

- [x] Other

## How should this be tested?

- [ ] Make a trade

## Checklist:

### DIFF
--- a/src/lib/utils/balancer/web3.ts
+++ b/src/lib/utils/balancer/web3.ts
@@ -42,10 +42,23 @@ export async function sendTransaction(
     const gasLimit = gasLimitNumber.toNumber();
     overrides.gasLimit = Math.floor(gasLimit * (1 + GAS_LIMIT_BUFFER));
 
-    if (USE_BLOCKNATIVE_GAS_PLATFORM && overrides.gasPrice == null) {
+    if (
+      USE_BLOCKNATIVE_GAS_PLATFORM &&
+      overrides.gasPrice == null &&
+      overrides.maxFeePerGas == null &&
+      overrides.maxPriorityFeePerGas == null
+    ) {
       const gasPrice = await gasPriceService.getLatest();
       if (gasPrice != null) {
-        overrides.gasPrice = gasPrice;
+        if (
+          gasPrice.maxFeePerGas != null &&
+          gasPrice.maxPriorityFeePerGas != null
+        ) {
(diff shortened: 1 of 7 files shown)


src/lib/utils/balancer/web3.ts
```diff
@@ -42,10 +42,23 @@ export async function sendTransaction(
     const gasLimit = gasLimitNumber.toNumber();
     overrides.gasLimit = Math.floor(gasLimit * (1 + GAS_LIMIT_BUFFER));
 
-    if (USE_BLOCKNATIVE_GAS_PLATFORM && overrides.gasPrice == null) {
+    if (
+      USE_BLOCKNATIVE_GAS_PLATFORM &&
+      overrides.gasPrice == null &&
+      overrides.maxFeePerGas == null &&
+      overrides.maxPriorityFeePerGas == null
```
evgenyboxer, line 49: if any of these are set I think the intent is for manual control over the gas price.

1saf: If any are changed then shouldn't we use an OR condition

src/services/gas-price/providers/blocknative.provider.ts
```diff
@@ -64,7 +67,13 @@ export default class BlocknativeProvider {
       }
 
       // gas price is in gwei
-      return gasPrice != null ? gasPrice.price * GWEI_UNIT : null;
+      return gasPrice != null
```
1saf, line 70: nit: i'd prefer a traditional if statement here compared to a ternary for the purpose of readability

1saf (approved):

src/lib/utils/balancer/web3.ts
evgenyboxer: I mean, I need all of them to be `null` to proceed (or more likely `undefined`) - but that will match both anyway.

src/services/gas-price/providers/blocknative.provider.ts
evgenyboxer: yeah I see what you mean, its a bit hard to read in this instance - I've pushed a fix for it

PR #663, opened 2021-08-11
Hotfix 1.12.8

garethfuller:
# Description

Cleans up references to chainId. We have a few exceptions being thrown in Sentry where chainId is missing. It's not easy to see exactly where this is happening so this PR is cleaning up all references. In particular, we should be using the `appNetworkConfig.chainId` everywhere except where it's a user action that explicitly needs to know the user's current chainId, for example when checking if they're on the wrong network.

## Type of change

- [ ] Other

## How should this be tested?

- [ ] Test all functionality

## Checklist:

### DIFF
--- a/src/services/web3/web3.plugin.ts
+++ b/src/services/web3/web3.plugin.ts
@@ -89,10 +89,7 @@ export default {
     });
 
     const chainId = computed(() => {
-      if (pluginState.connector) {
-        return pluginState.connector.chainId;
-      }
-      return Number(process.env.VUE_APP_NETWORK);
+      return pluginState.connector?.chainId;
     });
 
     const provider = computed(
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer-labs/frontend-v2",
-  "version": "1.12.7",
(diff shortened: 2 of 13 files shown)


src/services/web3/web3.plugin.ts
```diff
@@ -89,10 +89,7 @@ export default {
     });
 
     const chainId = computed(() => {
-      if (pluginState.connector) {
-        return pluginState.connector.chainId;
-      }
-      return Number(process.env.VUE_APP_NETWORK);
+      return pluginState.connector?.chainId;
```
garethfuller, line 92: I think we should explicitly return the user's chainId here (or null if not connected) and not mix it up with the app chainId. wdyt @1saf?

1saf: I think the appChainId return here was a guard for when user chain id here was undefined in some cases

garethfuller: Yeah that's my point, in those cases we should be either handling that it's undefined or using the app network chainId. Since I've fixed all the other references, I think it's less likely to be a problem. Will need to thoroughly test the previews.

PR #732, opened 2021-08-24
Lazy network switching (no more millions of Metamask popups)

TomAFrench:
# Description

We're currently immediately requesting that the user switch their network to match whatever the UI is expecting to as soon as they enter the page. This is a massive pain (especially if you have polygon and mainnet open at the same time.) and we've got multiple complaints about it (including @alexvansande).

I've removed the current logic which checks the network on connecting to a wallet and instead added a button to the alert navbar to ask Metamask to switch. 

## Type of change

- [ ] Other

## How should this be tested?

Load the app with metamask connected to the wrong network. Click button on nav and check that UI behaves as expected afterwards.

## Checklist:

### DIFF
--- a/src/composables/queries/useUserClaimsQuery.ts
+++ b/src/composables/queries/useUserClaimsQuery.ts
@@ -74,7 +74,6 @@ export default function useUserClaimsQuery(
   const queryOptions = reactive({
     enabled: isQueryEnabled,
     refetchOnMount: false,
-    refetchOnWindowFocus: false,
     ...options
   });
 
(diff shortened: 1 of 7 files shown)


src/composables/queries/useUserClaimsQuery.ts
```diff
@@ -74,7 +74,6 @@ export default function useUserClaimsQuery(
   const queryOptions = reactive({
     enabled: isQueryEnabled,
     refetchOnMount: false,
-    refetchOnWindowFocus: false,
```
TomAFrench, line 77: I found that without removing this the pending rewards pill didn't update on switching networks.

garethfuller:
@1saf is the best person to review this PR since he implemented the web3 connection module. As far as I remember there is a reason we needed to trigger this on page load, I think it means super snappy loading if you had previously connected, so we'd be losing some benefit here too.

PR #756, opened 2021-08-30
Fix trade interface switch

evgenyboxer:
# Description

Langer reported strange behavior with the trade switch (being visible on Polygon deployment). This fix replaces the `chainId` check to the app network one instead of the user one.

## Type of change

- [ ] Other

## How should this be tested?

- [x] Network switch should not appear on Polygon deployment regardless of what network the user has selected.
- [x] Ethereum networks should have the switch

## Checklist:

### DIFF
--- a/src/components/navs/AppNav/AppNavSettings.vue
+++ b/src/components/navs/AppNav/AppNavSettings.vue
@@ -218,12 +218,12 @@ export default defineComponent({
     const {
       explorerLinks,
       account,
-      chainId,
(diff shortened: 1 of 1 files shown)


TomAFrench (approved):
LGTM.

I wasn't aware that `chainId` referred to the user's connected chain rather than the one the app is using. It seems like a little bit of a footgun so maybe we should rename it or just remove it and rely on `userNetworkConfig`?

PR #757, opened 2021-08-30
Add a switch for Balancer <> Gnosis trading

evgenyboxer:
# Description

This PR implements the new trade switch once Gnosis becomes the default interface. This allows the user to switch between gasless (Gnosis) and Balancer trades.

## Type of change

- [x] Other

## How should this be tested?

- [ ] Hit the switch button and watch rates change.

## Checklist:


src/components/cards/TradeCardGP/TradeCardGP.vue
```diff
@@ -53,20 +53,34 @@
...
-            @click="switchToWETH()"
-            class="text-blue-500"
-            v-text="$t('tradeFromWeth')"
-          />
-          {{ $t('saveGas') }}
-        </span>
+      <div class="mt-6 text-sm flex items-center">
+        <BalTooltip
+          width="64"
+          :disabled="!trading.isGaslessTradingDisabled.value"
```
garethfuller, line 59: `.value` shouldn't be required in templates. What's going on here?

src/composables/trade/useTrading.ts
```diff
@@ -226,6 +246,14 @@ export default function useTrading(
     handleAmountChange();
   });
 
+  onMounted(() => {
+    const gaslessDisabled = window.location.href.includes('gasless=false');
```
evgenyboxer, line 250: This is used to force the trade interface to use Balancer instead of Gnosis (for example, when copperlaunch links to us during LBP).

It's important that this happens only once (when the component mounts), so that the user can switch back to gasless.

@garethfuller This was the "bullet proof" way to ensure this query string is in the URL... Using `vue-router` doesn't quite work with `onMounted` because it won't pick up there is a query string - it will only pick it up when the URL is being watched/computed etc.

garethfuller: Happy to do it this way since it's straightforward. I think it might have been possible with router guards but I'm not sure: https://next.router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards

src/constants/local-storage.keys.ts
```diff
@@ -8,6 +8,9 @@ export default {
     TradeDeadline: 'app.tradeDeadline',
     EthereumTxType: 'app.ethereumTxType'
   },
+  Trade: {
+    Gasless: 'Trade.Gasless'
```
garethfuller, line 12: Can we stick to the syntax used by all the other keys in this file (camelCase)?

garethfuller: Also, can we now get rid of the `app.tradeInterface` key?

evgenyboxer: yes! sorry my bad.

src/components/cards/TradeCardGP/TradeCardGP.vue
garethfuller: hmm something doesn't seem right here. Could you wrap the `trading` export in `toRefs(trading)` to see if that helps?

PR #793, opened 2021-09-07
[DO NOT MERGE]: Open discussion about replacing singleton services with dependency injection (TypeDI)

TimvanScherpenzeel:
# Description

This PR partially implements [TypeDI](https://github.com/typestack/typedi) as a possible replacement for the current singleton service design. I have included several examples of how one could use TypeDI in both classes, functions and other services. The main benefits of using this dependency injection system is that it guarantees that the service only gets instantiated once and is accessible on the instance (https://github.com/typestack/typedi#basic-usage). It is recommended to use `private readonly` to guarantee that users do not override what is meant to be `readonly`.

Notable changes:
- Add `reflect-metadata` (2.8kb minified + gzipped) import at root (`main.ts`) and Jest setup (required for `typedi` (2.7kb minified + gzipped)
- Replaced ConfigService singleton with TypeDI service (`@Service()` decorator).
- Instance is now callable through `Container.get(ConfigService)` and by passing it into the constructor of a `@Service()` decorated class. From that point onwards you can access it through `this`. The cleanest example of the constructor case is `src/services/ipfs/ipfs.service.ts`.
- Add `.env` to gitignore, noticed that it was required to run `npm run build`. In the Dockerfile the `.env.development` gets passed in explicitly.

I've generally tried to stick to the existing code style except for default exports. When using the service pattern with TypeDI I have found it more natural to work with named exports (this is of course a personal preference).

Fixes # (issue)

This PR has no ticket number assigned but is part of an ongoing internal discussion.

## Type of change

- [X ] Other, part of internal discussion with the team

## How should this be tested?

Application should run and behave exactly the same.
The unit tests should still pass.
Running the development server does not introduce new warnings or errors.

## Checklist:


tsconfig.json
```diff
@@ -15,22 +16,11 @@
     "baseUrl": ".",
     "noImplicitAny": false,
     "resolveJsonModule": true,
-    "types": [
```
TimvanScherpenzeel, line 18: Prettier autoformatted these changes, can revert if preferred.

src/services/balancer/subgraph/balancer-subgraph.client.ts
```diff
@@ -1,12 +1,13 @@
...
 import { jsonToGraphQLQuery } from 'json-to-graphql-query';
-import { configService as _configService } from '@/services/config/config.service';
+import { ConfigService } from '@/services/config/config.service';
 
 export default class BalancerSubgraphClient {
   url: string;
 
-  constructor(private readonly configService = _configService) {
-    this.url = configService.network.subgraph;
+  constructor(private readonly configService = Container.get(ConfigService)) {
```
garethfuller, line 9: Can this not be:
```suggestion
  constructor(private readonly configService = ConfigService) {
```
I thought this was partly the point of TypeDI, to clean up the dependency injection.

src/services/balancer/subgraph/entities/poolActivities/index.ts
```diff
@@ -1,12 +1,15 @@
-import Service from '../../balancer-subgraph.service';
+import { BalancerSubgraphService } from '../../balancer-subgraph.service';
```
garethfuller, line 1: What's the reasoning behind removing the default export here?

src/services/balancer/subgraph/entities/pools/index.ts
```diff
@@ -26,18 +27,18 @@ import { lidoService } from '@/services/lido/lido.service';
...
-    service: Service,
+    service: BalancerSubgraphService,
     query: QueryBuilder = queryBuilder,
-    private readonly configService = _configService
+    private readonly configService = Container.get(ConfigService)
   ) {
     this.service = service;
     this.query = query;
-    this.networkId = Number(configService.env.NETWORK) as NetworkId;
+    this.networkId = Number(this.configService.env.NETWORK) as NetworkId;
```
garethfuller, line 41: Since we're only accessing this instance passed inside the constructor and it's passed in as a param, what's the need to use `this` here?

garethfuller (commented):
Thanks for you work on this Tim! Using a DI system has been something I've been wanting to look into for a while.

I've commented throughout but I'll repeat the important ones here:

1. It seems unnecessary to use `Container.get` everywhere instead of just instantiating those commonly used classes in the class files, like we were doing before. Is there a reason why you chose to do it this way?
2. It was my understanding that one of the primary benefits would be cleaner dependency injection. i.e. the ability to do this in constructors:
    ```ts
    constructor(private readonly configService: ConfigService) { ...
    ```
    Instead you seem to have used Container.get in these cases as well. Is there a reason that's required?

src/services/config/config.service.ts
```diff
@@ -38,5 +40,3 @@ export default class ConfigService {
     return configs[key];
   }
 }
-
-export const configService = new ConfigService();
```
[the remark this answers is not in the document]
TimvanScherpenzeel: That makes sense to me and that would work yes. Ideally you would only use the dependency injection system through the `@Service` decorator and avoid referring to `Container.get(...)` directly. In a hybrid codebase like this I understand why you would export it like this. If we were to implement this I think this would be an appropriate solution for the short term (basically keep exporting it as a singleton).

README.md
```diff
@@ -8,6 +8,11 @@ git clone https://github.com/balancer-labs/frontend-v2.git && cd frontend-v2
 ```
 
 ### Local env
+Copy .env.development to .env
+```bash
+cp .env.development .env
```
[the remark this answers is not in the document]
TimvanScherpenzeel: >    Instead you seem to have used Container.get in these cases as well. Is there a reason that's required?

To my understanding the dependency only gets injected in constructors that are decorated with `@Service` themselves. If you were to move all the services to the dependency injection based system it would look a lot cleaner and you wouldn't need to use `Container.get(...)`.

src/services/balancer/subgraph/balancer-subgraph.client.ts
TimvanScherpenzeel: To my understanding that would be possible if `BalancerSubgraphClient` was a `@Service` itself, something I felt was beyond the scope of this ticket as it has implications on how it can be used (not instantiated in a singleton).

From the docs (https://docs.typestack.community/typedi/02-basic-usage-guide) it mentions:

```
IMPORTANT NOTE:
Don't forget to annotate your classes with the @Service decorator! Both the ones being injected and those which requests the dependencies should be annotated.
```

src/services/balancer/subgraph/entities/poolActivities/index.ts
TimvanScherpenzeel: Mostly a personal preference. When working with `TypeDI` I prefer to know exactly which service one is referring to. It is also my opinion that `Service` - as exported here - is not clear enough if we introduce the formal concept of services through `TypeDI`.

src/services/balancer/subgraph/entities/pools/index.ts
TimvanScherpenzeel: Good catch! When using `TypeDI` I prefer using `this.`, even in the constructor, to indicate that the service you are passing in is actually accessible on `this` (as apposed to having to be manually assigned in the constructor). This is useful if the class implements any methods so you know you already have access to `this.configService`. Again, mostly personal preference.

In this case `Pools` does not implement the `@Service` decorator - making it necessary to use the `Container.get(...)` syntax. Ideally all classes in `/services` would be decorated with `@Service` which would eliminate the concept of `passing as a param` in exchange for `injecting a service` that one has access to using `this`.

README.md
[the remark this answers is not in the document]
TimvanScherpenzeel: Perhaps it is a configuration issue on my end - I noticed it wouldn't succesfully run `npm run build` without a `.env`.

PR #797, opened 2021-09-08
New token inputs

garethfuller:
# Description

Replaces [BalTextInput](https://github.com/balancer-labs/frontend-v2/blob/572d70f2b36cd11c2b264ed1cb198cda6db65eb6/src/components/_global/BalTextInput/BalTextInput.vue) and builds [TokenInput](https://github.com/balancer-labs/frontend-v2/blob/572d70f2b36cd11c2b264ed1cb198cda6db65eb6/src/components/inputs/TokenInput/TokenInput.vue) as an abstraction on top.

## Type of change

- [ ] Other

## How to test

- [ ] Test trade form functionality (Gnosis and Balancer)
- [ ] Test invest / withdraw input functionality i.e. custom investment & single asset withdrawals.

## Checklist:


src/components/_global/BalTextInput/composables/useInputStyles.ts
```diff
@@ -0,0 +1,93 @@
...
+  const inputHeightClass = (): string => {
+    switch (props.size) {
+      case 'sm':
+        return 'h-8';
+      case 'lg':
+        return 'h-12';
+      default:
+        return 'h-10';
+    }
+  };
```
garethfuller, line 46: I have set these so that the default size of the inputs is correct for the trade / invest interfaces. sm & lg sized inputs may not be appropriate and so we'll have to adjust these classes as we make more use of the component.

src/components/cards/TradeCard/TradePair.vue
```diff
@@ -1,373 +1,161 @@
...
-          ? (balanceNumber - ETH_BUFFER).toString()
-          : '0';
-      handleInAmountChange(maxAmount);
-    }
-
-    function handleInAmountChange(value: string): void {
-      emit('exactInChange', true);
-      emit('tokenInAmountChange', value);
-      emit('change', value);
-    }
```
garethfuller, line 273: A lot of this logic is now handled by the TokenInput component. One thing I might not have handled correctly is setting the max in the Gnosis case.

src/components/inputs/TokenInput/TokenInput.vue
```diff
@@ -0,0 +1,219 @@
...
+});
+
+const barColor = computed(() => {
+  if (bnum(_amount.value).gt(tokenBalance.value)) return 'red';
+  return 'green';
+});
+
+const priceImpactSign = computed(() =>
+  (props.priceImpact || 0) >= 0 ? '-' : '+'
+);
```
garethfuller, line 116: I don't think we actually know if the PI is positive. Did you not look into this @evgenyboxer? It seems that for negative price impact the SOR returns a positive number. So I assume this is an absolute number coming from the SOR. We may need to fix this.

For now I've assumed if the PI number coming out of the SOR is negative, it's actually a positive PI. I don't think the SOR will return a negative PI numbers so this should never happen. This will need to be refactored in a separate PR.

evgenyboxer (approved):
@garethfuller amazing work!

src/components/inputs/TokenInput/TokenInput.vue
evgenyboxer: I have looked into it and it *can* be a negative if we don't do the `Math.max` operation to get a "min" price impact.

However, sometimes that negative was actually a positive which got me confused... gotta follow this up with John. I think its great there is support for it in the UI.

PR #805, opened 2021-09-10
Add Element.fi trading support

TomAFrench:
# Description

This PR adds support for us to trade using Element.fi assets - most notably their principle/yield tokens (The tokenlist also includes some irrelevant tokens but we can filter these out at some point.) We keep the same behaviour of not showing them on the Invest pages.

## Type of change

- [ ] Other

## How should this be tested?

- [ ] Check that no Element pools appear in the invest pages
- [ ] Check that you can route a trade from an asset (e.g. USDC) to the relevant principle/yield tokens
- [ ] Check that you can route a trade from a principle/yield tokens to other principle/yield tokens with the same underlying
- [ ] Check that prices are near those quoted on Element.fi

## Checklist:

### DIFF
--- a/src/composables/queries/usePoolQuery.ts
+++ b/src/composables/queries/usePoolQuery.ts
@@ -10,7 +10,7 @@ import { POOLS } from '@/constants/pools';
 import useApp from '../useApp';
 import useUserSettings from '../useUserSettings';
 import { forChange } from '@/lib/utils';
-import { isStableLike } from '../usePool';
+import { isElement, isStableLike } from '../usePool';
 
 export default function usePoolQuery(
   id: string,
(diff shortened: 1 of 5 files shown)


evgenyboxer (approved):

garethfuller:
@TomAFrench we chose not to include the Element pools in out investments list. That is why we went for the featured section below that links directly to their pool pages. Reason being, we don't have the required info to calculate the same stuff we we do for our own pools in the table, e.g. total value / APR.

TomAFrench:
I haven't had a chance to fix this issue yet so haven't pushed any changes since the review.

garethfuller:
Closing since this became too out of sync.

PR #835, opened 2021-09-17
Show the no liquidity error message when relavent

beethovenxfi:
# Description

Currently, the no liquidity error message never shows. The user isn't provided with any visual cue on why they can't execute their trade. This PR is a one line change to only show the EMPTY state when both tokenInAmount and tokenOutAmount are empty, rather than either.

**Before:**
Screen Shot 2021-09-17 at 10 31 28 AM

**After:**
Screen Shot 2021-09-17 at 10 34 42 AM

## Type of change

- [ ] Other

## How should this be tested?

- [ ] Navigate to the Trade view, select ETH -> DAI, enter a very large amount of ETH, verify that the not enough liquidity error message appears.
- [ ] Navigate to the Trade view, select ETH -> DAI, provide empty or zero values to either side, verify that the EMPTY state is present.

## Checklist:

### DIFF
--- a/src/composables/trade/useValidation.ts
+++ b/src/composables/trade/useValidation.ts
@@ -31,7 +31,11 @@ export default function useValidation(
   const validationStatus = computed(() => {
     if (!isWalletReady) return TradeValidation.NO_ACCOUNT;
 
-    if (!tokensAmountsValid.value) return TradeValidation.EMPTY;
+    if (
+      !isValidTokenAmount(tokenInAmount.value) &&
+      !isValidTokenAmount(tokenOutAmount.value)
+    )
+      return TradeValidation.EMPTY;
 
     const nativeAssetBalance = parseFloat(balances.value[nativeAsset.address]);
     if (nativeAssetBalance < MIN_NATIVE_ASSET_REQUIRED) {


garethfuller (commented):
Thanks for looking into this, but the changes here don't actually change the logic. Also, TradeValidation.EMPTY does not prevent the 'no liquidity' error from showing, as long as you have the funds to input a large enough amount otherwise you will instead get the 'not enough funds' error.

Closing this PR for now.

beethovenxfi:
Hey, totally fine to close it, but your statement isn't exactly right. With the current logic, the no liquidity message will never show. `if (!tokensAmountsValid.value)` will always be true if there is no liquidity, so it will always return `TradeValidation.EMPTY`.

It's very easy to test on the deployment. The two images below show the case. In the first image, the amount far exceeds my balance, but I still get a quoted price. Once I enter an amount that exceeds the available liquidity, it hits the `TradeValidation.EMPTY` case and shows nothing.

Screen Shot 2021-09-23 at 12 42 19 PM

Screen Shot 2021-09-23 at 12 42 24 PM

garethfuller:
@beethovenxfi thanks for pushing for this again, you are right about the logic but I think it was still highly unlikely you'd see the `NO_LIQUIDITY` warning unless you had enough funds to not trigger the `NO_BALANCE` warning first. I've reopened and refactored your original work slightly here #867.

PR #869, opened 2021-09-24
Refactor app network id

evgenyboxer:
# Description

This is the first PR in a series of PRs that will make our app network switching within the app (and not in a new subdomain).
I will try to make them minimal so we can make sure it doesn't break the app.

There is now only one place that uses `VUE_APP_NETWORK` (in the new composable `useNetwork`)

Fixes # (issue)

## Type of change

- [x] Other

## How should this be tested?

- [ ] App should work on all networks, nothing should change.

## Checklist:


src/composables/useNetwork.ts
```diff
@@ -0,0 +1,31 @@
...
+  POLYGON = 137,
+  ARBITRUM = 42161
+}
+
+const DEFAULT_NETWORK =
+  process.env.VUE_APP_NETWORK != null
+    ? (Number(process.env.VUE_APP_NETWORK) as NetworkId)
+    : Network.MAINNET;
+
+export const networkId = ref<NetworkId>(DEFAULT_NETWORK);
```
evgenyboxer, line 20: Changing a network is as easy as changing this initial value to something else. Try, refresh and see!

src/components/cards/TradeCard/TradeRoute.vue
```diff
@@ -433,9 +434,9 @@ export default defineComponent({
     function getPoolLink(id: string): string {
       const chainId = appNetworkConfig.chainId;
       const prefixMap = {
-        1: 'app.',
-        42: 'kovan.',
-        137: 'polygon.'
+        [Network.MAINNET]: 'app.',
+        [Network.KOVAN]: 'kovan.',
+        [Network.POLYGON]: 'polygon.'
```
evgenyboxer, line 439: Throughout this refactor, I'll replace all the numeric/string network ids to use the `Network` enum for better readability.

src/composables/useNetwork.ts
```diff
@@ -0,0 +1,31 @@
+import { ref } from 'vue';
+
+export type NetworkId = 1 | 3 | 4 | 5 | 42 | 137 | 42161;
```
evgenyboxer, line 3: For now I'm keeping this `NetworkId`, but I believe we can end up just using the enum.

src/lib/config/index.ts
```diff
@@ -54,15 +55,15 @@ export interface Config {
...
+const config: Record<NetworkId | 12345 | 17, Config> = {
+  [Network.MAINNET]: homestead,
+  [Network.KOVAN]: kovan,
+  [Network.RINKEBY]: rinkeby,
+  [Network.POLYGON]: polygon,
+  [Network.ARBITRUM]: arbitrum,
+  12345: test,
   // @ts-ignore
-  '17': docker
+  17: docker
```
garethfuller, line 66: Is there maybe a good reason to use strings instead of numbers as the key?

src/plugins/sentry.ts
```diff
@@ -2,20 +2,20 @@ import { App } from 'vue';
...
 const ENV = process.env.VUE_APP_ENV || 'development';
-const NETWORK = process.env.VUE_APP_NETWORK || '42';
 const networkMap = {
   '1': 'mainnet',
   '42': 'kovan',
   '137': 'polygon',
   '42161': 'arbitrum-one'
 };
-const environment = `${ENV}-${networkMap[NETWORK]}`;
+const environment = `${ENV}-${networkMap[networkId.value]}`;
```
garethfuller, line 18: Will this work? The `networkMap` keys are strings but I think the `networkId.value` is a number right? This is a more general issue. There are a lot of places where the networkId is expected to be a string and not a number.

garethfuller (commented):
In general seems like a great first step! I'm just a bit concerned about using numbers as the key type. If it doesn't cause any problems I'm totally fine with it though.

src/composables/useNetwork.ts
evgenyboxer: It is my understanding that Balancer is deployed on all TestNets, and that we just need to support it on the frontend.
The Rinkeby/Ropsten networks are actually used in the code for Matcha service / Gnosis endpoint.

I guess its possible to remove it until we support it, but also feels like there is no harm for now?

src/plugins/sentry.ts
evgenyboxer: Ah, nice catch. I'll fix this. (change to number)

`networksMap[1] === networksMap['1']` because numeric keys are always converted to strings.

src/lib/config/index.ts
evgenyboxer: Just a preference I think... just gotta pick one and stick with it. (its going to be easier once everything uses the `enum`).

The docker/test are special cases which I don't think need to be included in that `enum`.

src/composables/useNetwork.ts
garethfuller: Yeah fine to leave it, just checking.

garethfuller (approved):

PR #870, opened 2021-09-24
Add Github Action to build + push Dockerfile

timjrobinson:
# Description

- Add a new Github Action to build Dockerfile.dev and push it to Dockerhub
- Tag image with 'latest' and current package.json version

This needs a dockerhub login added to settings to work, see checklist. 

## Type of change

- [ ] Other

## How should this be tested?

- I tested this on my personal fork of this repository. Can see test here: https://github.com/timjrobinson/frontend-v2/pull/1. This successfully tagged and pushed the image to Dockerhub
- I tested running `docker run timbalancer/frontend-v2` locally and it ran correctly.

This can be tested by creating a PR in your own fork of this repository, changing the `on` section, and seeing it build + push there. Or you can test with `docker run timbalancer/frontend-v2`

## Checklist:

- [ ] Create Dockerhub build user and add username/password to secrets.


timjrobinson:
I made it do a multi-stage build. Though this makes it do a production build on Ethereum Mainnet which loads but then can't connect to Infura so can't pull any data. 

What should we do about this? Possible options:

- Docker image could connect to Kovan by default, like dev mode does. 
- Could have 2 docker images?

Do we want a dev docker image with hot reloading / un-minified sources? I imagine people developing locally will just git clone and run locally rather than using a Docker image. 

I'd like the Production + Mainnet Docker image to be something people can run to self-host their own frontend. This would require them being able to pass Infura keys or fall back to using some other method of fetching data (maybe connecting to a local Ethereum node?) But I'm not sure how we should allow them to specify these config settings.

timjrobinson:
I've added a new production Dockerfile that allows users to specify their Infura API key when running it. Also changed the configuration so that the Infura URL and API Key are separate variables so that the API key can be easily overwritten and will be used across all networks. 

Main difficulty was environment variables are specified at build time and we don't want users to have to rebuild before being able to run the image. So I included a small shell script that checks the API key is set and if it is it `sed's` the `./dist` directory inserting the key where appropriate. 

Possible issues:
- How does Vercel insert production environment variables? Want to ensure that the new .env.production file won't break things. Vercel should use its own variables. 

Still TODO:
- Change http-server to nginx.

timjrobinson:
I've updated the docker image to use Nginx now, it also runs on port 80. Also changed the infura env variable to `INFURA_PROJECT_ID` to be consistent with what they call it in their dashboard.

src/lib/config/arbitrum.json
```diff
@@ -37,6 +37,10 @@
     "lidoRelayer": "",
     "balancerHelpers": "0x77d46184d22CA6a3726a2F500c776767b6A3d6Ab"
   },
+  "keys": {
+    "infura": "daaa68ec242643719749dd1caba2fc66",
+    "alchemy": "ODJ9G5Ipv-Hb2zTWMNbUFIqv9WtqBOc2"
+  },
```
garethfuller, line 43: Why do we need the keys in the file here? For defaults? We could potentially move these to ENV vars in Vercel if that helps?

src/services/config/config.service.ts
```diff
@@ -37,6 +44,27 @@ export default class ConfigService {
       throw new Error(`No config for network key: ${key}`);
     return configs[key];
   }
+
+  public get rpc(): string {
+    return template(this.network.rpc, {
```
garethfuller, line 49: I've not seen this es6 dynamic template before, what is it doing?

.env.production
```diff
@@ -0,0 +1,9 @@
+VUE_APP_ENV=production
+VUE_APP_DOMAIN=localhost:8080
+VUE_APP_IPFS_NODE=ipfs.fleek.co
+VUE_APP_NETWORK=1
+VUE_APP_BLOCKNATIVE_DAPP_ID=032e2fb8-6c66-46a5-bf1c-a049ac7eded2
```
garethfuller, line 5: The deployer will need to provide the blocknative ID.

src/services/config/config.service.ts
timjrobinson: Just a regex replace. I'll actually move this function into our utils instead of using it from npm. Or add something more vue like (my worry about using vue here was it might create a dependency loop).

.env.production
timjrobinson: I'll actually rename this to .env.docker, so it doesn't interfere with our own build and deploy process.

src/lib/config/arbitrum.json
timjrobinson: Yea it was inconsistencies between environment configs, different environments had different keys for the same services.

src/lib/config/polygon.json
```diff
@@ -6,9 +6,9 @@
   "shortName": "Polygon",
   "network": "polygon",
   "unknown": false,
-  "rpc": "https://polygon-mainnet.infura.io/v3/daaa68ec242643719749dd1caba2fc66",
-  "ws": "wss://polygon-mainnet.g.alchemy.com/v2/ODJ9G5Ipv-Hb2zTWMNbUFIqv9WtqBOc2",
-  "loggingRpc": "https://polygon-mainnet.g.alchemy.com/v2/hsAEQ4l4c6uz3ciAdQ9w3ODWjVcND-Ti",
```
timjrobinson, line 11: Anyone know why we have two different alchemy keys for Polygon? Which one is correct?

timjrobinson:
I've added the ability to specify your Blocknative Key/Dapp ID (required) and Portis Dapp ID (optional). I tested Blocknative and it works correctly. 

When testing Portis I found a bug which is causing Portis to break in prod - the network is set to `configService.network.network` which is currently `homestead` but should be `mainnet`. After fixing this Portis worked correctly. I did have to run the container with `--network=host` for Portis to work as it only allows connections from https sites and localhost. 

I added a script `scripts/run-docker.sh` to make it easier for the user to fill in the variables and run. Maybe this should be a docker-compose file? Though having two docker-compose files may get confusing.

garethfuller (approved):

src/lib/config/arbitrum.json
```diff
@@ -37,6 +37,10 @@
     "lidoRelayer": "",
     "balancerHelpers": "0x77d46184d22CA6a3726a2F500c776767b6A3d6Ab"
   },
+  "keys": {
+    "infura": "daaa68ec242643719749dd1caba2fc66",
+    "alchemy": "VBeQgTCRqqPtuuEPsFzRdwKXzDyN6aFh"
```
timjrobinson, line 42: The old key was incorrect, it was the Polygon key. This new one is the correct one for Arbitrum.

PR #875, opened 2021-09-25
Support proportional joins on InvestmentPools when trading is halted

EndymionJkb:
# Description

Investment Pools can only be joined proportionally when trading is halted. That join kind was not supported, so it tries to join with EXACT_TOKENS_IN_FOR_BPT_OUT - and the contract reverts with INVALID_JOIN_EXIT_KIND_WHILE_SWAPS_DISABLED. In this case, it has to do ALL_TOKENS_IN_FOR_EXACT_BPT_OUT.

Also - there was one bug, which accidentally worked because the stable and weighted join kinds were the same (until now).
Line 18 of joinParams.ts:     this.isStablePool = isStable(exchange.pool);

isStable(), isInvestment(), etc. take a PoolType parameter, but it was passing in the whole pool object

This is a draft now because:

1) Had trouble calling getSwapEnabled on chain at first; fixed by Gareth. I had one commit (reverted) that changed to using the subgraph instead of onchain - but since the onchain check works now, that's probably safer

2) It is letting the system calculate the BPT out as if you were doing an exact tokens in with "best price" - and then doing a proportional join with that BPT out. This works, but it should compute and check maxAmountsIn; I don't think it is.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue) - kinda; the bug was a sleeper - more future-proofing
- [ ] Other

## How should this be tested?

Should be able to join the Investment Pool, while trading is halted:
http://localhost:8080/#/pool/0xe7b1d394f3b40abeaa0b64a545dbcf89da1ecb3f00010000000000000000009a

I succeeded in adding proportional liquidity when running it locally.

## Checklist:


src/services/pool/exchange/serializers/JoinParams.ts
```diff
@@ -54,15 +61,27 @@ export default class JoinParams {
...
     if (this.exchange.pool.onchain.totalSupply === '0') {
       return this.dataEncodeFn({ kind: 'Init', amountsIn });
     } else {
-      return this.dataEncodeFn({
-        kind: 'ExactTokensInForBPTOut',
-        amountsIn,
-        minimumBPT
-      });
+      // Investment Pools can only be joined proportionally if trading is hallted
+      if (this.isInvestmentPool && isProportional && !this.isSwapEnabled) {
```
garethfuller, line 73: @EndymionJkb I don't think the `isProportional` is required here. We are hiding custom inputs in the form if it's an investment pool in this state so you can assume it is proportional.

garethfuller (approved):

src/services/balancer/contracts/contracts/vault.ts
```diff
@@ -43,11 +47,20 @@ export default class Vault {
 
     if (isWeightedLike(type)) {
       tokenMultiCaller.call('weights', poolAddress, 'getNormalizedWeights', []);
+
+      if (isTradingHaltable(type)) {
+        //TODO: This says Error: no matching function (argument="name", value="getSwapEnabled", code=INVALID_ARGUMENT, version=abi/5.4.0)
+        //tokenMultiCaller.call('swapEnabled', poolAddress, 'getSwapEnabled');
```
garethfuller: @EndymionJkb I think this is fixed by importing the InvestmentPool__factory as one of the ABIs available to the tokenMulticaller. I've made the changes here: https://github.com/balancer-labs/frontend-v2/pull/875/commits/0c17b7588701423d2ee2c2f5d00ba2dd6ca6855f

src/services/pool/exchange/serializers/JoinParams.ts
EndymionJkb: Yes - I wrote that before I hid the custom inputs

PR #881, opened 2021-09-27
Add "networkId" to all queries

evgenyboxer:
# Description

Adds "networkId" to all queries

## Type of change

- [ ] Other

## How should this be tested?

- [ ] Everything should load normally

## Checklist:


src/composables/queries/usePoolQuery.ts
```diff
@@ -34,7 +36,7 @@ export default function usePoolQuery(
   /**
    * QUERY INPUTS
    */
-  const queryKey = QUERY_KEYS.Pools.Current(id);
+  const queryKey = QUERY_KEYS.Pools.Current(networkId, id);
```
garethfuller, line 39: This is what I'm not sure about, but I suppose we can change it later when it becomes relevant. Given a pool ID we should always only load that pool. So I'm not sure what this networkId is going to represent, is it the user's networkId or is it the path defined networkId?

garethfuller (approved):
Looks good and doesn't seem to break anything. 

Although, I'm not sure if the network key is going to be needed in certain queries where the path defines what the network is and so there will never be a case where you have pool with the same ID but different networkIds. In these cases I'm not sure what the benefit is of using the networkId in the query is. I guess it doesn't hurt.

src/composables/queries/usePoolQuery.ts
evgenyboxer: You are right, I don't think we need `networId` here.

PR #895, opened 2021-09-28
Hotfix/1.21.5

johngrantuk:
# Description

`getCostOfSwapInToken` was being called with default swap cost value, this value was then used as V1 swap cost. For V2 `getSwap` is being called with a higher swap cost (SWAP_COST). This meant that when V1 and V2 return amounts considering fees were being compared V1 would show a better amount. I've changed `getCostOfSwapInToken` to use SWAP_COST so we have a fair comparrison.

Also V1 uses EVM scaled value for swap cost but we were passing in human scale number. It's now scaled correctly.

Fixes # (issue)

## Type of change

- [ ] Other

## How should this be tested?

Monitor console logs and check V1 return amount with fees and V2 return amount with fees outputs. V2 liquidity should usually be chosen even if prices are similar.

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer-labs/frontend-v2",
-  "version": "1.21.4",
+  "version": "1.21.5",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@balancer-labs/frontend-v2",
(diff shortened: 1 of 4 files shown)


TomAFrench:
Looks good to me based on ETH <> USDC swaps

TomAFrench (approved):

PR #899, opened 2021-09-29
hotfix/1.21.7

evgenyboxer:
Hides the chosen token from the other `TokenInput`

The bug shown in the screenshot:

Screen Shot 2021-09-28 at 8 49 17 pm

### DIFF
--- a/src/components/inputs/TokenSelectInput/TokenSelectInput.vue
+++ b/src/components/inputs/TokenSelectInput/TokenSelectInput.vue
@@ -12,6 +12,7 @@ type Props = {
   modelValue: string;
   fixed?: boolean;
   weight?: number;
+  excludedTokens?: string[];
 };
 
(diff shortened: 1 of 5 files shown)


src/components/inputs/TokenSelectInput/TokenSelectInput.vue
```diff
@@ -91,7 +93,7 @@ const toggleModal = () => {
     <teleport to="#modal">
       <SelectTokenModal
         v-if="openTokenModal"
-        :excludedTokens="[modelValue]"
+        :excludedTokens="[...excludedTokens, modelValue]"
```
evgenyboxer, line 96: By default `excludedTokens` is defined as an empty array so this would always work.

garethfuller (approved):
Nice catch, thanks!

PR #920, opened 2021-10-07
hotfix/1.22.3

evgenyboxer:
This is a fix for a small bug of missing space after "Effective price" for Gnosis trade interface (preview screen)

Here's how it looks now:
Screen Shot 2021-10-07 at 10 23 08 pm

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer-labs/frontend-v2",
-  "version": "1.22.2",
+  "version": "1.22.3",
   "lockfileVersion": 2,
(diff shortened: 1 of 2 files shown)


src/components/modals/TradePreviewModalGP.vue
```diff
@@ -8,14 +8,12 @@
           >
             <span>
               {{ $t('effectivePrice') }}
-            </span>
-            <span
-              v-html="
```
evgenyboxer, line 13: This isn't `html` anymore, so I've removed it and it also fixed the spacing issue.

PR #942, opened 2021-10-13
hotfix/1.23.4

evgenyboxer:
Fixes https://sentry.io/organizations/balancer-labs/issues/2699308234/?project=5725878

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer-labs/frontend-v2",
-  "version": "1.23.3",
+  "version": "1.23.4",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@balancer-labs/frontend-v2",
-      "version": "1.23.3",
+      "version": "1.23.4",
       "license": "MIT",
       "dependencies": {
         "@balancer-labs/assets": "github:balancer-labs/assets#master",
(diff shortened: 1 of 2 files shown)


src/composables/watchers/useWeb3Watchers.ts
```diff
@@ -92,6 +93,8 @@ export default function useWeb3Watchers() {
   });
 
   watch(blockNumber, async () => {
-    handlePendingTransactions();
+    if (isWalletReady.value) {
+      handlePendingTransactions();
```
evgenyboxer, line 97: Pending transaction handling only makes sense when a wallet is connected

src/composables/useTransactions.ts
```diff
@@ -396,25 +396,27 @@ export default function useTransactions() {
...
-        console.log(
-          '[Transactions]: Failed to fetch tx information',
-          transaction,
-          e
-        )
-      )
-      .finally(() =>
-        updateTransaction(transaction.id, 'tx', {
-          lastCheckedBlockNumber: blockNumber.value
+    if (provider.value != null) {
```
evgenyboxer, line 399: I've added this guard here as well (which was causing the error in the first place)

1saf (approved):

timjrobinson (approved):

src/composables/useTransactions.ts
garethfuller: I think this might still fail because the exception is complaining that `provider` is undefined. Maybe it could be:
```suggestion
    if (provider?.value) {
```
or
```suggestion
    if (provider !== undefined) {
```

PR #958, opened 2021-10-19
UI 746 refactor swapper/web3 into services

timjrobinson:
# Description

- Create Swap service that holds most of the logic that swapper used to
do.
- Create new SwapToken type that can work for both swapIn and swapOut
methods to avoid duplication.
- Create services for accessing each of the different solidity
contracts, so that it is clear what contracts are being called with what
parameters.
- Create Contract Caller service that is similar to web3 with cleaner
code and less parameters required.
- Add new method to gas price service to generate gas price overrides
when calling contracts.

This has not been tested yet, Creating PR up to get feedback on code layout and style. 

## TODO

- Make sure this works
- Add unit tests for functions for services
- Make it work with multiple networks. 

## Type of change

- [ ] Other

## How should this be tested?

- [x] Ensure Unit tests pass and test functionality correctly
- [x] Perform a batch swap given a max input token value that does a V1 swap
- [x] Perform a batch swap given a min output token value that does a V1 swap
- [x] Perform a native asset swap given max input that does a V1 swap
- [x] Perform a batch swap given a max input token value that does a V2 swap
- [x] Perform a batch swap given a min output token value that does a V2 swap
- [x] Perform a native asset swap given max input that does a V2 swap
- [x] Perform a batch swap given a max input token value that does a Lido V2 swap
- [x] Perform a batch swap given a min output token value that does a Lido V2 swap
- [x] Perform a boosted pools invest swap
- [x] Perform a boosted pools withdraw swap

## Checklist:

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201312345085445
- To learn more about using Asana, visit [asana.com](https://app.asana.com) to get started today.


src/services/contracts/exchange-proxy.service.ts
```diff
@@ -0,0 +1,88 @@
...
+import { TransactionResponse } from '@ethersproject/abstract-provider';
+import { contractCaller } from './contract-caller.service';
+
+export default class ExchangeProxyService {
+  network: string;
+  address: string;
+  abi: any;
+  signer: any;
+  contract: any;
+  contractWithSigner: any;
```
evgenyboxer, line 14: I think you can find the types for these in the `ethers` library

src/services/contracts/exchange-proxy.service.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  constructor() {
+    this.network = 'homestead';
+    this.address = configs[this.network].addresses.exchangeProxy;
```
evgenyboxer, line 18: lets grab the network/address right before the usage (because the app will switch networks soon)

garethfuller (commented):
This is a good start. I've left some minor comments but I have some higher level / architectural comments which are difficult to include here. Let's discuss elsewhere.

src/services/web3/web3.plugin.ts
```diff
@@ -105,6 +106,9 @@ export default {
         rpcProviderService.getJsonProvider(chainId.value)
     );
     const signer = computed(() => pluginState.connector?.provider?.getSigner());
+    const web3Provider = computed(() => {
+      return new Web3Provider(pluginState.connector.provider as any);
+    });
```
garethfuller, line 111: Lets rename this to be more explicit.
```suggestion
    const userProvider = computed(() => {
      return new Web3Provider(pluginState.connector.provider as any);
    });
```

src/services/web3/web3.service.ts
```diff
@@ -1,20 +1,40 @@
...
 export default class Web3Service {
-  provider: JsonRpcProvider;
+  provider: ComputedRef<Web3Provider | JsonRpcProvider>;
+
+  constructor(
+    private readonly rpcProviderService = _rpcProviderService,
+    private readonly config: ConfigService = configService
+  ) {
+    this.provider = computed(() => this.rpcProviderService.jsonProvider);
+  }
```
garethfuller, line 29: We are mixing up providers here, lets be more explicit about them, there are two. The appProvider and the userProvider. The appProvider is for reading onchain data, the userProvider is for writing onchain data (signing transactions).

```suggestion
  appProvider: ComputedRef<JsonRpcProvider>;

  constructor(
    private readonly rpcProviderService = _rpcProviderService,
    private readonly config: ConfigService = configService
  ) {
    this.appProvider = computed(() => this.rpcProviderService.jsonProvider);
  }
```

src/services/web3/web3.service.ts
```diff
@@ -1,20 +1,40 @@
...
 
-  constructor(readonly rpcProviderService = _rpcProviderService) {
-    this.provider = this.rpcProviderService.jsonProvider;
+  public setProvider(provider: ComputedRef<Web3Provider | JsonRpcProvider>) {
+    this.provider = provider;
```
garethfuller, line 32:
```suggestion
  public setUserProvider(provider: ComputedRef<Web3Provider>) {
    this.userProvider = provider;
```

src/services/web3/web3.service.ts
```diff
@@ -1,20 +1,40 @@
...
   }
 
   async getEnsName(address: string): Promise<string | null> {
     try {
-      return await this.provider.lookupAddress(address);
+      return await this.provider.value.lookupAddress(address);
```
garethfuller, line 37: We're reading data here so use appProvider
```suggestion
      return await this.appProvider.value.lookupAddress(address);
```

src/services/web3/web3.service.ts
```diff
@@ -25,6 +45,59 @@ export default class Web3Service {
       ens: await this.getEnsName(address)
     };
   }
+
+  async getUserAddress(): Promise<string> {
+    const signer = this.provider.value.getSigner();
+    const userAddress: string = await signer.getAddress();
+    return userAddress;
+  }
```
garethfuller, line 53: This is a little exception, we are reading data here but it's the userProvider address we're interested in.
```suggestion
  async getUserAddress(): Promise<string> {
    const signer = this.userProvider.value.getSigner();
    const userAddress: string = await signer.getAddress();
    return userAddress;
  }
```

src/services/web3/web3.service.ts
```diff
@@ -25,6 +45,59 @@ export default class Web3Service {
...
+
+  public async sendTransaction(
+    contractAddress: string,
+    abi: any[],
+    action: string,
+    params: any[],
+    options: Record<string, any>,
+    forceEthereumLegacyTxType = false
+  ): Promise<TransactionResponse> {
+    const signer = this.provider.value.getSigner();
```
garethfuller, line 63:
```suggestion
    const signer = this.userProvider.value.getSigner();
```

garethfuller (changes requested):
Looking good. Main comment is that I think this is a good place to start distinguishing the app/user provider difference.

src/services/contracts/exchange-proxy.service.ts
```diff
@@ -0,0 +1,86 @@
...
+export default class ExchangeProxyService {
+  abi: any;
+
+  constructor() {
+    this.abi = exchangeProxyAbi;
+  }
+
+  private get address() {
+    return configs[networkId.value].addresses.exchangeProxy;
+  }
```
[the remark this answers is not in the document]
timjrobinson: ConfigService makes sense. I thought we were moving away from storing the address in the constructor though as it won't change if the user switches networks.

timjrobinson:
Unit tests for all major functionality have been added and this is now ready for review.

src/services/contracts/exchange-proxy.service.ts
```diff
@@ -0,0 +1,88 @@
+import exchangeProxyAbi from '@/lib/abi/ExchangeProxy.json';
+import ConfigService, { configService } from '@/services/config/config.service';
+import { Address, SwapToken, SwapTokenType } from '../swap/swap.service';
+import { Swap } from '@balancer-labs/sor/dist/types';
+import { TransactionResponse } from '@ethersproject/abstract-provider';
+import Web3Service, { web3Service } from '../web3/web3.service';
+
+export default class ExchangeProxyService {
+  abi: any;
```
evgenyboxer, line 9: I believe typing it as `ContractInterface` should do the trick since this is what is needed when creating a contract with ethers.

Note - you can probably leave multicall's `abi: any` because it's a bit different since its creating a new `Interface` (and not `Contract`)

evgenyboxer:
lgtm!

src/services/swap/swap.service.ts
```diff
@@ -0,0 +1,244 @@
...
+    swaps: Swap[][]
+  ): Promise<TransactionResponse> {
+    console.log('[Swap Service] batchSwapV1');
+
+    const options: any = {};
+
+    /* I have no idea what this is doing still... */
+    if (tokenIn.address === NATIVE_ASSET_ADDRESS) {
+      options.value = `0x${tokenIn.amount.toHexString()}`;
+    }
```
garethfuller, line 51: Did you write this comment? It's just a kind of weird way of setting the transaction value if the input token is ETH.

src/services/swap/swap.service.ts
```diff
@@ -0,0 +1,244 @@
...
+    swaps: SwapV2[],
+    tokenAddresses: string[]
+  ): Promise<TransactionResponse> {
+    if (isStETH(tokenIn.address, tokenOut.address)) {
+      return this.lidoBatchSwap(tokenIn, tokenOut, swaps, tokenAddresses);
+    }
+    console.log('[Swap Service] batchSwapV2');
+    const overrides: any = {};
+
+    if (tokenIn.address === AddressZero) {
```
garethfuller, line 78: I'm interested why this is AddressZero (0x000...) and for the v1 batch swap we're using the native asset address which is (0xEeee...). Please test to see if tokenIn.address for a v2 batch swap will actually be 0x00.. when ETH is the token in and not 0xEeee...

src/services/web3/web3.service.ts
```diff
@@ -1,20 +1,42 @@
...
 export default class Web3Service {
-  provider: JsonRpcProvider;
+  appProvider: ComputedRef<JsonRpcProvider>;
+  userProvider!: ComputedRef<Web3Provider>;
+
+  constructor(
+    private readonly rpcProviderService = _rpcProviderService,
+    private readonly config: ConfigService = configService
+  ) {
+    this.appProvider = computed(() => this.rpcProviderService.jsonProvider);
```
garethfuller, line 30: For now at least, the appProvider/jsonProvider is static and since `this.rpcProviderService.jsonProvider` is not a reactive property there is no point setting appProvider as a computed property because it will never update.

Also, although I think it's ok to pass in reactive properties from the frontend as inputs, I think we should avoid defining reactive properties within services.

garethfuller (changes requested):

src/services/swap/swap.service.ts
garethfuller: lets try setting options.value to this instead:
```suggestion
    /* I have no idea what this is doing still... */
    if (tokenIn.address === NATIVE_ASSET_ADDRESS) {
      options.value = tokenIn.amount;
    }
```

src/services/swap/swap.service.ts
timjrobinson: Yea I'm not sure why it's like this, it was like this in Swapper so left it as is.

src/services/web3/web3.service.ts
timjrobinson: Makes sense

timjrobinson:
I've re-tested all combinations on Kovan (except boosted pools, will ping @garethfuller on how to test this) after rebasing onto develop and all are working. Swapping ETH with V1 liquidity only worked after I changed the WETH token address to `0xd0a1e359811322d97991e03f863a0c30c2cf029c` in the Kovan config. It seems this was the old WETH used in testing V1 and is what was specified in the ExchangeProxy contract constructor: https://kovan.etherscan.io/address/0x4e67bf5bD28Dd4b570FBAFe11D0633eCbA2754Ec#code

timjrobinson:
I've tested boosted pools invest and it's working. 

Boosted pools withdraw isn't working on develop or this branch, throwing an error when clicking the final button. Asked @garethfuller if he knows what's going on.

src/services/swap/swap.service.ts
```diff
@@ -0,0 +1,349 @@
...
+  ): Promise<TransactionResponse> {
+    console.log('[Swapper] lidoBatchSwap');
+    const overrides: any = {};
+
+    if (tokenIn.address === AddressZero) {
+      overrides.value = tokenIn.amount;
+    }
+
+    // Convert tokenIn/tokenOut so that it matches what's in tokenAddresses
+    const { stETH, wstETH } = this.config.network.addresses;
```
evgenyboxer, line 143: minor - I think it would nice if was `stETHaddr` or even using `this.config.network.addresses.stETH` because `stETH.toLowerCase()` reads as if this was a symbol

evgenyboxer (approved):
lgtm!

garethfuller (approved):
Looks good. Just some minor comments and a suggestion for a further refactor.

Tested
- [x] Invest (weighted pool)
- [x] Withdraw (weighted pool)
- [x] Invest (stable pool)
- [x] Withdraw (stable pool)
- [x] Invest (boosted pool)
- [x] Withdraw (boosted pool)
- [x] Trade WETH -> MKR
- [x] Create a pool

src/services/contracts/exchange-proxy.service.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+      const maxTotalAmountIn = '10';
+      await exchangeProxyService.multihopBatchSwapExactOut(
+        swaps,
+        tokens.USDC.address,
+        tokens.DAI.address,
+        maxTotalAmountIn
+      );
+      const sendTransactionArgs = require('@/services/web3/web3.service')
+        .web3Service.sendTransaction.mock.calls[0];
+      const transactionParams = sendTransactionArgs[3];
```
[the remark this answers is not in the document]
timjrobinson: When this require is called it actually loads `@/services/web3/__mocks__/web3.service` because it's mocked at the top. This exports a constant `web3Service` that is initialized in the mock file. Jest automatically uses this mock for all requires of this file throughout the app, then because require()'ing something twice returns the same object (as it caches it) it's able to read in the arguments that were passed to the mock function.

src/services/swap/swap.service.ts
```diff
@@ -0,0 +1,349 @@
...
+      const limits: string[] = [];
+      tokenAddresses.forEach((token, i) => {
+        if (tokensIn.includes(token.toLowerCase())) {
+          limits[i] = amountsInMap[token].toString();
+        } else if (token.toLowerCase() === tokenOut.toLowerCase()) {
+          limits[i] = amountOutMin.mul(-1).toString();
+        } else {
+          limits[i] = '0';
+        }
+      });
```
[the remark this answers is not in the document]
timjrobinson: Done

src/services/swap/swap.service.ts
```diff
@@ -0,0 +1,349 @@
...
+    } catch (e) {
+      console.log('[Swapper] lidoBatchSwap Error:', e);
+      return Promise.reject(e);
+    }
+  }
+
+  /**
+   * Join a Boosted Pool (StablePhantom) using a batch swap
+   */
+  public async boostedJoinBatchSwap(
```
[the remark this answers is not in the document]
timjrobinson: I think this could work. I'd like to do this refactor in another PR though as it'll mean a lot more testing again.

src/services/swap/swap.service.ts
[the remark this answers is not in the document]
garethfuller: But it doesn't look like it's changed? It's still generating the limits within this function.

garethfuller:
Tested boosted pool invest/withdraw again after latest changes and looks like it's working as expected. Let's merge this in after next week's weekly release so it sits on develop for a week for further testing.

PR #964, opened 2021-10-20
DRAFT PoolCreator service

timjrobinson:
# Description

- Create new service for creating pools
- Create test file for service

Only very basic right now, working on a way to test this and see how it responds. 

## Type of change

- [ ] Other

## How should this be tested?

- Make sure unit tests pass
- Use with Pool creation UI, ensure pools create correctly. 

## Checklist:

### DIFF
--- a/src/composables/pools/usePoolCreation.ts
+++ b/src/composables/pools/usePoolCreation.ts
@@ -4,12 +4,13 @@ import { BigNumber } from 'bignumber.js';
 import { flatten } from 'lodash';
 import { ref, reactive, toRefs, watch, computed, toRef } from 'vue';
 import { useQuery } from 'vue-query';
-import usePoolsQuery from '../queries/usePoolsQuery';
+import usePoolsQuery from '@/composables/queries/usePoolsQuery';
 import useTokens from '../useTokens';
 import useWeb3 from '@/services/web3/useWeb3';
-import { poolCreator } from '@/services/pool-creator/pool-creator.service';
+import { balancerService } from '@/services/balancer/balancer.service';
 import { AddressZero } from '@ethersproject/constants';
 import { scale } from '@/lib/utils';
+import { PoolType } from '@/services/balancer/subgraph/types';
 
 export type TokenWeight = {
   tokenAddress: string;
@@ -127,7 +128,7 @@ export default function usePoolCreation() {
 
   const createPool = async () => {
     const provider = getProvider();
-    const poolDetails = await poolCreator.createWeightedPool(
+    const poolDetails = await balancerService.pools.create(PoolType.Weighted)(
       provider,
       'MyPool',
       getPoolSymbol(),
@@ -140,9 +141,9 @@ export default function usePoolCreation() {
     poolCreationState.poolAddress = poolDetails.address;
   };
 
-  const joinPool = () => {
+  const joinPool = async () => {
     const provider = getProvider();
-    poolCreator.joinPool(
+    const result = await balancerService.pools.join(
       provider,
       poolCreationState.poolId,
       account.value,
@@ -150,6 +151,7 @@ export default function usePoolCreation() {
       poolCreationState.tokenWeights,
       getScaledAmounts()
     );
+    console.log("Got join pool result: ", result);
   };
 
   const tokensList = computed(() => poolCreationState.tokensList);
@@ -161,6 +163,7 @@ export default function usePoolCreation() {
     return _balances;
   });
 
+  const result = usePoolsQuery(tokensList, {}, { isExactTokensList: true });
   const {
     data: similarPoolsResponse,
     isLoading: isLoadingSimilarPools
--- a/src/composables/pools/usePoolCreation.ts
+++ b/src/composables/pools/usePoolCreation.ts
@@ -1,5 +1,6 @@
 import { balancerSubgraphService } from '@/services/balancer/subgraph/balancer-subgraph.service';
 import { getAddress } from '@ethersproject/address';
+import { BigNumber } from 'bignumber.js';
 import { flatten } from 'lodash';
 import { ref, reactive, toRefs, watch, computed, toRef } from 'vue';
 import { useQuery } from 'vue-query';
@@ -7,6 +8,8 @@ import usePoolsQuery from '../queries/usePoolsQuery';
 import useTokens from '../useTokens';
 import useWeb3 from '@/services/web3/useWeb3';
 import { poolCreator } from '@/services/pool-creator/pool-creator.service';
+import { AddressZero } from '@ethersproject/constants';
+import { scale } from '@/lib/utils';
 
 export type TokenWeight = {
   tokenAddress: string;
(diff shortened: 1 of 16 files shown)


src/services/pool-creator/pool-creator.service.ts
```diff
@@ -0,0 +1,167 @@
...
+      provider,
+      vaultAddress,
+      Vault__factory.abi,
+      'joinPool',
+      [poolId, sender, receiver, joinPoolRequest]
+    );
+
+    const receipt: TransactionReceipt = await tx.wait();
+
+    return receipt;
```
garethfuller, line 126: This service should return the transaction promise and the frontend should handle the transaction so we can add it to the notifications etc.
```suggestion
    return await sendTransaction(
      provider,
      vaultAddress,
      Vault__factory.abi,
      'joinPool',
      [poolId, sender, receiver, joinPoolRequest]
    );
```

src/services/pool-creator/pool-creator.service.ts
```diff
@@ -0,0 +1,167 @@
...
+    tokens: TokenWeight[],
+    owner: Address
+  ): Promise<CreatePoolReturn> {
+    const weightedPoolFactoryAddress =
+      configService.network.addresses.weightedPoolFactory;
+
+    // Tokens must be sorted in order of addresses
+    tokens = this.sortTokens(tokens);
+
+    const tokenAddresses: Address[] = tokens.map((token: TokenWeight) => {
```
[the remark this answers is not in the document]
timjrobinson: This makes sense, it was re-used from the ui code @1saf what do you think about renaming this to `NewPoolToken` everywhere? And lists can be `tokens` instead of `tokenWeights`.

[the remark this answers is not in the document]
garethfuller: Yeah sounds good, I commented the same thing on Ansaf's PR too so he's aware. Not sure who's responsible for changing.

src/services/pool-creator/pool-creator.service.ts
```diff
@@ -0,0 +1,167 @@
...
+const JOIN_KIND_INIT = 0;
+
+export interface JoinPoolRequest {
+  assets: Address[];
+  maxAmountsIn: string[];
+  userData: any;
+  fromInternalBalance: boolean;
+}
+
+export class PoolCreator {
```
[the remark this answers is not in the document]
timjrobinson: That sounds good. I've created that service and updated it to be an API of:

```
balancerService.pools.create(PoolType.WeightedPool)(...args)
```

src/services/pool-creator/pool-creator.service.ts
timjrobinson: What about for the create? Issue with create is it's two steps to first create the pool then retrieve its ID.

src/services/balancer/entities/pools.ts
```diff
@@ -0,0 +1,157 @@
...
+import { configService } from '@/services/config/config.service';
+import BigNumber from 'bignumber.js';
+import { BigNumber as EPBigNumber } from '@ethersproject/bignumber';
+import { sendTransaction } from '@/lib/utils/balancer/web3';
+import { defaultAbiCoder } from '@ethersproject/abi';
+import { TokenWeight } from '@/composables/pools/usePoolCreation';
+import { toNormalizedWeights } from '@balancer-labs/balancer-js';
+import { scale } from '@/lib/utils';
+
+type Address = string;
```
garethfuller, line 19: I think this is an unnecessary type.

src/services/balancer/entities/pools.ts
```diff
@@ -0,0 +1,157 @@
...
+      owner
+    ];
+
+    const tx: TransactionResponse = await sendTransaction(
+      provider,
+      weightedPoolFactoryAddress,
+      WeightedPoolFactory__factory.abi,
+      'create',
+      params
+    );
```
garethfuller, line 83: This promise is what should be returned, the frontend should then be responsible for waiting for it and making another call to get the id.
```suggestion
    return await sendTransaction(
      provider,
      weightedPoolFactoryAddress,
      WeightedPoolFactory__factory.abi,
      'create',
      params
    );
```

src/services/balancer/entities/pools.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    const poolDetails: CreatePoolReturn = {
+      id: poolId,
+      address: poolAddress
+    };
+
+    return poolDetails;
+  }
+
+  public async join(
```
garethfuller, line 100: Since this pool service may eventually contain all pool related logic I think we should call this `initJoin` so we are being specific about the type of join. We have other logic elsewhere about different types of joins that may move here if it makes sense.

garethfuller (changes requested):

src/services/pool-creator/pool-creator.service.ts
garethfuller: I've answered below, if there are two steps, e.g. create and fetch id, then there should really be two functions, and the frontend should be responsible for waiting for the first and then calling the second.

src/composables/pools/usePoolCreation.ts
```diff
@@ -93,19 +102,75 @@ export default function usePoolCreation() {
...
+    const response = await balancerService.pools.join(
+      provider,
+      poolCreationState.poolId,
+      account.value,
+      account.value,
+      poolCreationState.tokenWeights,
+      getScaledAmounts()
+    );
+    console.log("Got join pool response: ", response);
+  };
```
garethfuller, line 155: Because these are submitting transactions we need to handle them like we handle others. See the investment transaction for example: https://github.com/balancer-labs/frontend-v2/blob/develop/src/components/forms/pool_actions/InvestForm/components/InvestPreviewModal/components/InvestActions.vue#L185-L239

You will need to use a try/catch block to catch any transaction errors, get the pending tx returned from the service and pass it to some kind of tx handler function like in the example which adds the transaction to the global state and uses the txListener to figure out when it's confirmed.

src/services/balancer/entities/pools.ts
timjrobinson: So I was thinking I'd like this to be more standard, rather than just in this location. As there's many places where a string could mean an address, a stringified big number, or something else, and I thought it would be useful to type that explicitly so there's no confusion.

src/services/pool-creator/pool-creator.service.ts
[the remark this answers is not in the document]
timjrobinson: I think that makes even more sense, may as well do it now to establish that pattern, so will update it to that.

src/services/pool-creator/pool-creator.service.ts
timjrobinson: That makes sense

src/services/pool-creator/pool-creator.service.ts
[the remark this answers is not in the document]
garethfuller: What would also be helpful to establish now is a common API for all pool types. So we should have parent Pool class that defines the API, e.g. has a create function. Then the pool type classes extend that class, e.g. `class Weighted extends Pool {`

PR #1013, opened 2021-10-28
Boosted pool support

garethfuller:
# Description

This PR adds initial support for the StablePhantom pool type. e.g. the BB-A-USD pool. In particular:
- Adds the pool id to the stable pool allowlist
- Removes the pre-minted BPT token from the pool tokens list
- Adds a linear pool schema to the pool data, which maps linear pool data to the pool tokens
- Allows investments in BB-A-USD with stables (DAI, USDC, USDT) via batch swaps
- Allows withdrawals from BB-A-USD with stables (DAI, USDC, USDT) via batch swaps and batch relayer swaps
- Formats the pool page to present similar data to other pool types
- Formats the pool listing in pool lists to conform to other pool types

## Type of change

- [ ] Other

## How should this be tested?

- [ ] Test kovan pool `0x8fd162f338b770f7e879030830cde9173367f3010000000000000000000004d8`
- [ ] Test the regular pools have not been affected by the changes

## Checklist:


src/lib/utils/balancer/swapper.ts
```diff
@@ -247,6 +247,59 @@ async function batchSwapGivenInV2(
...
+      'batchSwap',
+      [SwapKind.GivenIn, swaps, tokenAddresses, funds, limits, MaxUint256],
+      overrides
+    );
+  } catch (error) {
+    console.log('[Swapper] batchSwapGivenInV2 Error:', error);
+    throw error;
+  }
+}
+
```
garethfuller, line 302: This will need to be refactored/moved into Tim's swap service when it's merged with develop.

src/services/balancer/contracts/contracts/vault.ts
```diff
@@ -27,88 +38,204 @@ export default class Vault {
...
+        mainToken: {
+          address: getAddress(mainToken.address),
+          index: mainToken.index.toNumber(),
+          balance: tokenData.balances[mainToken.index.toNumber()].toString()
+        },
+        wrappedToken: {
+          address: getAddress(wrappedToken.address),
+          index: wrappedToken.index.toNumber(),
+          balance: tokenData.balances[wrappedToken.index.toNumber()].toString()
+        }
```
garethfuller, line 224: @evgenyboxer this is the linear pool schema I've got so far we can add to it or adjust it as needed. Note that the balance is scaled and would need to be normalised with the token decimals, I don't think we should do that here since passing in the token info for the sub tokens is a bit tricky.

PR #1020, opened 2021-10-31
Replace multicall with multicall2

TomAFrench:
# Description

I've experienced a couple of times where upon switching network the entire app fails to load due to a multicall reverting (my feeling is that it's trying to pull the balances/allowance for a token that doesn't exist on the current network). This is generally fixed by clearing the cache, etc. but a bad address shouldn't prevent the entire app from loading.

To address this I've replaced the use of [multicall](https://github.com/makerdao/multicall/blob/master/src/Multicall.sol) with [multicall2](https://github.com/makerdao/multicall/blob/master/src/Multicall2.sol). The main difference being that we can optionally allow individual calls to fail without the entire set failing. Any queries to get a token's balance/allowance will now return zero if the query failed.

As a bonus I've made `multicall()` a generic function so we can pass in the expected type of the return values.

## Type of change

- [ ] Other

## How should this be tested?

Inject an address which isn't an erc20 token into the app's list of tokens. The app should continue to work well.

## Checklist:


evgenyboxer:
lgtm!

evgenyboxer (approved):

PR #1075, opened 2021-11-11
fix: trade input validation fix for Balancer routed trades

evgenyboxer:
# Description

This is a follow to https://github.com/balancer-labs/frontend-v2/pull/1065
The problem was that Gnosis didn't really need this check since its "loading" state is more like "refreshing". For Balancer routed trades - this is a good solution.

## Type of change

- [ ] Other

## How should this be tested?

- [ ] Try to input quickly before the pools are loaded (Balancer trades) - it won't let you since the input is disabled.
- [ ] Gnosis / Wrapped trades should be working as normal (without losing their focus)

## Checklist:

### DIFF
--- a/src/components/cards/TradeCardGP/TradeCardGP.vue
+++ b/src/components/cards/TradeCardGP/TradeCardGP.vue
@@ -52,8 +52,8 @@
       />
       <BalBtn
         v-if="trading.isLoading.value"
-        :loading="true"
-        :disabled="true"
+        loading
+        disabled
         :loading-label="
           trading.isGnosisTrade.value ? $t('loadingBestPrice') : $t('loading')
         "
--- a/src/components/cards/TradeCardGP/TradeCardGP.vue
+++ b/src/components/cards/TradeCardGP/TradeCardGP.vue
@@ -16,6 +16,9 @@
         v-model:tokenOutAmount="tokenOutAmount"
         v-model:tokenOutAddress="tokenOutAddress"
         v-model:exactIn="exactIn"
+        :tradeLoading="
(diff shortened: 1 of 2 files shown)


src/components/cards/TradeCardGP/TradeCardGP.vue
```diff
@@ -49,8 +52,8 @@
       />
       <BalBtn
         v-if="trading.isLoading.value"
-        loading
-        disabled
+        :loading="true"
+        :disabled="true"
```
garethfuller, line 56: You don't need to pass `true` here, simply setting the attribute is the same thing, `loading` is equal to `:loading="true"`

garethfuller (commented):
Again I think this is a bit of a hacky solution. Really the inputs should always be accessible otherwise it can lead to confusion for the user. I don't understand why validation and the trade data is not reactive. If a user enters a number before the trade data is ready then:
1. We shouldn't run validation
2. When trade data is ready we should re-assess the inputs and present the trade
3. Re-trigger validations

I've tried to explain the problem in more detail here: https://www.loom.com/share/d7c266ab7df7438e893ebb5b760b137d

src/components/cards/TradeCardGP/TradeCardGP.vue
evgenyboxer: oh oops, you are right.

evgenyboxer:
I totally agree with you - this needs a proper solution.
But for now, this might be a good quick win since the current behaviour is indeed worse.

garethfuller (approved):

PR #1077, opened 2021-11-12
feat: add progress bar buffer

evgenyboxer:
# Description

Adds a little buffer for the progress bar to accommodate for a recommended minimum gas fee along with a warning.

Added a new `minTransactionBuffer` to the `nativeAsset` config to have custom values for each network.

## Type of change

- [ ] Other

## How should this be tested?

- [ ] Trying clicking "max" (the buffer along with warning should appear)
- [ ] Try manually input a value that is close to the "max" (the buffer along with warning should appear)

## Checklist:

### DIFF
--- a/src/components/_global/BalProgressBar/BalProgressBar.vue
+++ b/src/components/_global/BalProgressBar/BalProgressBar.vue
@@ -24,12 +24,12 @@ const props = withDefaults(defineProps<Props>(), {
  */
 const barClasses = computed(() => ({
   [`h-${props.size}`]: true,
-  [`bg-${props.color}-500 dark:bg-${props.color}-400`]: true
+  [`bg-${props.color}-400`]: true
 }));
 
 const bufferBarClasses = computed(() => ({
   [`h-${props.size}`]: true,
-  [`bg-yellow-500 dark:bg-yellow-400`]: true
+  [`bg-yellow-500`]: true
 }));
 
 const barStyles = computed(() => ({
--- a/src/components/_global/BalProgressBar/BalProgressBar.vue
+++ b/src/components/_global/BalProgressBar/BalProgressBar.vue
@@ -6,6 +6,7 @@ import { computed } from 'vue';
  */
 type Props = {
   width: string | number;
+  bufferWidth: string | number;
   size?: string;
   color?: string;
 };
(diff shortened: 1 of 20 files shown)


src/lib/config/polygon.json
```diff
@@ -23,7 +23,8 @@
     "symbol": "MATIC",
     "decimals": 18,
     "deeplinkId": "matic",
-    "logoURI": "https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0x7D1AfA7B718fb893dB30A3aBc0Cfc608AaCfeBB0/logo.png"
+    "logoURI": "https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0x7D1AfA7B718fb893dB30A3aBc0Cfc608AaCfeBB0/logo.png",
+    "minTransactionBuffer": "0.1"
```
evgenyboxer, line 27: I believe for MATIC it should be enough

src/components/forms/pool_actions/InvestForm/composables/useInvestMath.ts
```diff
@@ -45,7 +47,7 @@ export default function useInvestFormMath(
    * COMPOSABLES
    */
   const { toFiat } = useNumbers();
-  const { tokens, balances, balanceFor, nativeAsset } = useTokens();
```
garethfuller, line 48: This native asset returned from useTokens pulls the schema from the config. No need to import this manually above. https://github.com/balancer-labs/frontend-v2/blob/develop/src/providers/tokens.provider.ts#L112-L115

src/components/inputs/TokenInput/TokenInput.vue
```diff
@@ -1,17 +1,24 @@
...
-import { TokenInfo } from '@/types/TokenList';
+import { configService } from '@/services/config/config.service';
+
 import { Rules } from '@/components/_global/BalTextInput/BalTextInput.vue';
-import { ETH_TX_BUFFER } from '@/constants/transactions';
+import TokenSelectInput from '@/components/inputs/TokenSelectInput/TokenSelectInput.vue';
+
+import { TokenInfo } from '@/types/TokenList';
+
+const { nativeAsset } = configService.network;
```
garethfuller, line 21: Again, you can pull this from useTokens.

garethfuller (changes requested):

src/components/forms/pool_actions/InvestForm/composables/useInvestMath.ts
evgenyboxer: Just pushed a fix for this.

The problem I was facing is that `nativeAsset` wasn't really a `TokenInfo`, it was more of a decorated one because it had custom props on it.

I've typed it now properly and it should be good now.

src/components/inputs/TokenInput/TokenInput.vue
```diff
@@ -250,9 +283,21 @@ watchEffect(() => {
         <BalProgressBar
           v-if="hasBalance && !noMax"
           :width="maxPercentage"
+          :buffer-width="bufferPercentage"
           :color="barColor"
           class="mt-2"
         />
+        <div
+          v-if="shouldShowTxBufferMessage"
+          class="mt-2 text-yellow-600 dark:text-yellow-400 text-sm"
```
garethfuller, line 292: Text should be `xs` to match BalTextInput error text size.
```suggestion
          class="mt-2 text-yellow-600 dark:text-yellow-400 text-xs"
```

garethfuller (changes requested):

src/components/_global/BalProgressBar/BalProgressBar.vue
```diff
@@ -23,27 +24,40 @@ const props = withDefaults(defineProps<Props>(), {
  */
 const barClasses = computed(() => ({
   [`h-${props.size}`]: true,
-  [`bg-${props.color}-400`]: true
+  [`bg-${props.color}-500 dark:bg-${props.color}-400`]: true
```
garethfuller, line 27: I realise Pon's designs might have said something different, but for now we should probably stick to green-400 since that's what we seem to be using as default everywhere else. Not sure why that is, longer term we should update the green color scale and use 500 as the default.
Screenshot 2021-11-12 at 10 45 16

evgenyboxer: Makes sense yeah. What about the yellow tho? stick with 500 as well without changes to the dark one?

garethfuller: Because we're not really using the yellow much elsewhere, if 500 looks good lets just stick with that.

evgenyboxer:
@garethfuller all reported issues should be fixed now

garethfuller (approved):

PR #1093, opened 2021-11-16
hotfix/1.28.3

evgenyboxer:
# Description

This PR has 2 fixes:

1. I was debugging an issue with our wallet module where switching networks would disconnect the wallet. This happens because the `disconnect` event is not really about disconnecting a user, it's more about when an RPC endpoint might have some issues (see - https://docs.metamask.io/guide/ethereum-provider.html#disconnect). So I would go from Mainnet -> Fantom and sometimes it would work, sometimes disconnect, etc... 

2. It seems that auto disconnecting a wallet is not working. (try to "lock" your MetaMask and you will see that the app does not pick up the fact you are disconnected). I've added a check that looks for an empty accounts array and then triggers the disconnect logic. I believe it's the right behavior.

## Type of change

- [ ] Other

## How should this be tested?

- [ ] Try switching to any network - you should not be disconnected.
- [ ] Try to lock your account and you should be disconnected.

## Checklist:

### DIFF
--- a/src/services/web3/connectors/connector.ts
+++ b/src/services/web3/connectors/connector.ts
@@ -25,21 +25,22 @@ export abstract class Connector {
   handleAccountsChanged = (accounts: string[]) => {
     if (accounts.length === 0) {
       this.handleDisconnect();
-    }
-    if (this.selectedAccount !== '') {
-      const account = accounts.find(
-        account => getAddress(account) === getAddress(this.selectedAccount)
-      );
-      // sense check the account that was previously connected
-      if (!account) {
-        this.account.value = accounts[0];
-        console.warn(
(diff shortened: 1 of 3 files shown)


src/services/web3/connectors/connector.ts
```diff
@@ -69,6 +73,5 @@ export abstract class Connector {
 
     this.provider.on('accountsChanged', this.handleAccountsChanged);
     this.provider.on('chainChanged', this.handleChainChanged);
-    this.provider.on('disconnect', this.handleDisconnect);
```
evgenyboxer, line 72: I've looked at `bnc-onboard` and it seems they don't use this event as well.

garethfuller (approved):

1saf (approved):

PR #1096, opened 2021-11-16
Fix asset row rendering issue when an asset is missing an icon

danielmkm:
# Description

When an asset is missing an icon, the rendering of the "My Wallet" component breaks down.

Screenshot 2021-11-16 at 07 19 38

Apply a `display: flex` to the row so that it properly inlines the div component that gets rendered when the asset has no icon image.

Tested the fix with assets only with icons, a mixture, and all without. Every combination renders properly now.

Screenshot 2021-11-16 at 07 38 14

Screenshot 2021-11-16 at 07 42 02

Screenshot 2021-11-16 at 07 40 07

## Type of change

- [ ] Other

## How should this be tested?

- [ ] Have an asset in your wallet that does not have an icon, navigate to the trade page, ensure the items are still laid out correctly 
- [ ] Ensure that assets with images still render correctly

## Checklist:

### DIFF
--- a/src/components/_global/BalAsset/BalAssetSet.vue
+++ b/src/components/_global/BalAsset/BalAssetSet.vue
@@ -102,7 +102,7 @@ export default defineComponent({
 
 <style scoped>
 .addresses-row {
-  @apply relative mb-3;
+  @apply relative mb-3 flex;
 }
 .addresses-row:last-child {
   @apply mb-0;


garethfuller (approved):
Thanks again @danielmkm! @1saf was going to look into this after pool creation so I'll let him review first before merging. One thing I'm not sure about is if the scope of BalAssetSet should consider rows at all. I originally built this component and it was only supposed to handle the stacked horizontal asset images we have in the pool tables.

1saf (approved):

1saf:
Thanks for this

PR #1113, opened 2021-11-22
Add ability to initJoin WETH pools with ETH

timjrobinson:
# Description

Add the ability to invest in pools which take a wrapped native asset (such as WETH) using the native asset. 

I made it so the token address in global state is always kept as the wrapped asset to not cause state issues. On create the wrapped native asset is used, then on join it is changed to `AddressZero` and `value` is passed along with the transaction. 

## Type of change

- [ ] Other

## How should this be tested?

- Go through create pool flow. Create a pool with WETH as an asset
- On initial balances page, switch WETH to ETH and set a value
- Complete creation and join pool with ETH
- Repeat the above with WETH to ensure that also works correctly. 

## Checklist:

### DIFF
--- a/src/composables/pools/usePoolCreation.ts
+++ b/src/composables/pools/usePoolCreation.ts
@@ -7,6 +7,7 @@ import useTransactions from '@/composables/useTransactions';
 import useEthers from '@/composables/useEthers';
 import useTokens from '../useTokens';
 import useWeb3 from '@/services/web3/useWeb3';
+import { configService } from '@/services/config/config.service';
 import { balancerService } from '@/services/balancer/balancer.service';
 import { AddressZero } from '@ethersproject/constants';
 import { bnum, scale } from '@/lib/utils';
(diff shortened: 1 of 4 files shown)


src/composables/pools/usePoolCreation.ts
```diff
@@ -115,9 +123,11 @@ export default function usePoolCreation() {
   const scaledLiquidity = computed(
     (): Record<string, OptimisedLiquidity> => {
       const scaledLiquidity = {};
-      const modifiedToken = findSeedTokenByAddress(
-        poolCreationState.manuallySetToken
-      );
+      const manuallySetToken =
```
timjrobinson, line 126: I tried always passing the wrapped token address to `poolCreationState.manuallySetToken` instead of doing this, but this computed value stopped updating and returned undefined instead and I couldn't figure out why.

garethfuller (changes requested):
If I'm seeding a WETH pool with ETH then I shouldn't need to approve WETH. See how we're handling this for investment approvals.

Other than that, seeding with ETH seems to work well! 🙌

timjrobinson:
> If I'm seeding a WETH pool with ETH then I shouldn't need to approve WETH. See how we're handling this for investment approvals.

This is fixed.

garethfuller (approved):

1saf (approved):

PR #1121, opened 2021-11-25
feat: add phantom pool chart support

evgenyboxer:
# Description

Adds pool chart

Fixes # (issue)

## Type of change

- [ ] Other

## How should this be tested?

- [ ] Check pool chart

## Checklist:


src/composables/queries/usePoolSnapshotsQuery.ts
```diff
@@ -48,16 +54,34 @@ export default function usePoolSnapshotsQuery(
...
+    const [prices, snapshots] = await Promise.all([
+      coingeckoService.prices.getTokensHistorical(tokens, days),
+      balancerSubgraphService.poolSnapshots.get(id, days)
+    ]);
 
-    const prices = await coingeckoService.prices.getTokensHistorical(
-      tokens,
-      days
-    );
-    const snapshots = await balancerSubgraphService.poolSnapshots.get(id, days);
```
evgenyboxer, line 60: I've merged these into 1 `promise.all`

src/components/contextual/pages/pool/PoolChart.vue
```diff
@@ -1,180 +1,193 @@
-<template>
```
evgenyboxer, line 1: Lots of changes here are the result of updating the file to the new style

src/components/contextual/pages/pool/PoolChart.vue
```diff
@@ -1,180 +1,193 @@
...
-      darkMode.value
-        ? tailwind.theme.colors.gray['600']
-        : tailwind.theme.colors.black
-    );
-
-    const chartColors = [tailwind.theme.colors.green['400'], hodlColor.value];
-
-    const nonEmptySnapshots = computed(() => {
-      if (!history.value || !history.value) return [];
-      return history.value.filter((state: any) => state.totalShares !== '0');
```
evgenyboxer, line 73: I've removed all the `any` since its actually being inferred correctly by TS.

src/composables/queries/usePoolQuery.ts
```diff
@@ -100,13 +100,16 @@ export default function usePoolQuery(
...
+        ...linearPool.tokens
+          .filter(token => token.address !== linearPool.address)
+          .map(token => ({ ...token, address: getAddress(token.address) }))
       );
     });
 
     pool.linearPoolTokens = linearPoolTokens;
+    pool.linearPoolTokensAddresses = linearPoolTokens.map(
+      token => token.address
+    );
```
garethfuller, line 112: I'm starting to think a single `linearPoolTokenMap` might be better here?

src/pages/pool/_id.vue
```diff
@@ -308,9 +310,14 @@ export default defineComponent({
...
         const tokensWithPrice = Object.keys(prices.value);
-        return !pool.value.tokenAddresses.every(token =>
-          tokensWithPrice.includes(token)
-        );
+
+        const tokens =
+          isStablePhantomPool.value &&
+          pool.value.linearPoolTokensAddresses != null
+            ? pool.value.linearPoolTokensAddresses
+            : pool.value.tokenAddresses;
```
garethfuller, line 318: This is a little hard to parse, maybe move the conditional into an understandable const like:
```suggestion
        const hasLinearPoolTokens =  isStablePhantomPool.value && pool.value.linearPoolTokensAddresses != null
        const tokens = hasLinearPoolTokens
            ? pool.value.linearPoolTokensAddresses
            : pool.value.tokenAddresses;
```

garethfuller (commented):
In general looks good, but we probably need to handle the price rate conversion.

src/composables/queries/usePoolQuery.ts
evgenyboxer: Agree, this would allow for an easy `Object.keys` / `Object.values` for both scenarios. Fixed that.

src/composables/queries/usePoolSnapshotsQuery.ts
```diff
@@ -48,18 +56,57 @@ export default function usePoolSnapshotsQuery(
...
+      tokens = pool.value.tokenAddresses;
+    }
+
+    const [historicalPrices, snapshots] = await Promise.all([
+      coingeckoService.prices.getTokensHistorical(tokens, days),
+      balancerSubgraphService.poolSnapshots.get(id, days)
+    ]);
+
+    if (isStablePhantomPool) {
+      const tokensWithoutPoolToken = tail(pool.value.tokens);
```
evgenyboxer, line 79: @garethfuller do you think this is a good method? Essentially I'm assuming the first token is always the pool token, so the index number matters.

src/composables/queries/usePoolSnapshotsQuery.ts
```diff
@@ -48,18 +56,57 @@ export default function usePoolSnapshotsQuery(
...
+
+      return {
+        // Adjust the price from coingecko by the priceRate (1 LINEAR-AUSDC -> priceRate * 1 USDC)
+        prices:
+          tokensWithoutPoolToken.length === tokens.length
+            ? mapValues(historicalPrices, prices =>
+                prices.map((price, index) => {
+                  // indexes of the tokens and mainTokens should match
+
+                  const token = tokensWithoutPoolToken[index];
```
evgenyboxer, line 89: This is kinda tricky... since what I've sent to coingecko are the `mainTokens` I need to figure out what was the `priceRate` of the `LINEAR-A{TOKEN}`, so for that I need to grab it from the `tokens`.

Another potential solution is perhaps to get it from `onchain.linearPools` - but I thought its kinda the same?

garethfuller: The formula is
```
LINEAR-AUSDC * priceRate = USDC
```

garethfuller: I think you should get the price rate via the linearPools map, e.g.
```
onchain.linearPools[token.address].priceRate
```

garethfuller (changes requested):

src/composables/queries/usePoolQuery.ts
```diff
@@ -99,14 +95,21 @@ export default function usePoolQuery(
...
+            ...token,
+            address
+          };
+        });
     });
 
-    pool.linearPoolTokens = linearPoolTokens;
+    pool.linearPoolTokensMap = linearPoolTokensMap;
+
+    console.log(pool.linearPoolTokensMap);
```
1saf, line 112: intended?

evgenyboxer:
@garethfuller are you happy for this to be merged?

src/components/contextual/pages/pool/PoolChart.vue
```diff
@@ -1,180 +1,209 @@
...
+    .filter(state => state.totalShares > 0);
+});
+
+const timestamps = computed(() => {
+  if (history.value.length === 0) {
+    return [];
+  }
+
+  return history.value.map(state =>
+    format(fromUnixTime(state.timestamp / 1000), 'yyyy/MM/dd')
```
garethfuller, line 106: Interesting that you still need to divide by 1000 in a fromUnixTime fn. I started adding helper functions in `useTime` maybe you could add something more universal to handle this case there?

garethfuller (changes requested):
Breaks charts on existing pool pages.

Screenshot 2021-12-03 at 13 27 40

evgenyboxer:
@garethfuller sorry my bad, we need to wait for the updated subgraph that has `liquidity` field on the `PoolSnapshot` entity. I'll follow up with Fabio.

src/components/contextual/pages/pool/PoolChart.vue
evgenyboxer: Yeah I don't think its needed actually because the timestamp is already normalized... so we can just `format` it.

BTW, unrelated by we seem to also have a `date.ts` which does a similar thing... we should merge them.

evgenyboxer:
@garethfuller added Mainnet/Arbitrum subgraphs, now just waiting for Polygon (and also asked for Rinkeby) and then we can merge this.

I've also guarded from a weird issue where the chart would return an invalid value.

evgenyboxer:
@garethfuller all subgraph links updated, so it should be good to merge now.

garethfuller (approved):
Not sure if it's related to this PR or if the subgraph isn't synced or something else, but there are no pools on the polygon preview. Otherwise the charts looks good everywhere else.

evgenyboxer:
Yeah it says its still syncing, I think it should be good.

PR #1127, opened 2021-11-26
feat: add unwrapped token address to linearPools

evgenyboxer:
# Description

Adds "unwrappedTokenAddress" to linearPool data.

Fixes # (issue)

## Type of change

- [ ] Other

## How should this be tested?

- [ ] `unwrappedTokenAddress` should be added to the onchain linearPool data

## Checklist:

### DIFF
--- a/src/composables/queries/usePoolQuery.ts
+++ b/src/composables/queries/usePoolQuery.ts
@@ -126,7 +126,9 @@ export default function usePoolQuery(
 
     if (isBlocked(pool)) throw new Error('Pool not allowed');
 
-    if (isStablePhantom(pool.poolType)) {
+    const isStablePhantomPool = isStablePhantom(pool.poolType);
+
+    if (isStablePhantomPool) {
       pool = removePreMintedBPT(pool);
       pool = await getLinearPoolAttrs(pool);
     }
@@ -157,8 +159,20 @@ export default function usePoolQuery(
       currency.value
     );
 
-    console.log('pool', { onchain: onchainData, ...decoratedPool });
-    return { onchain: onchainData, ...decoratedPool };
+    let unwrappedTokens: Pool['unwrappedTokens'];
+
+    if (isStablePhantomPool && onchainData.linearPools != null) {
+      unwrappedTokens = Object.entries(onchainData.linearPools).map(
+        ([, linearPool]) => linearPool.unwrappedTokenAddress
+      );
+    }
+
+    console.log('pool', {
+      onchain: onchainData,
+      unwrappedTokens,
+      ...decoratedPool
+    });
+    return { onchain: onchainData, unwrappedTokens, ...decoratedPool };
   };
 
   const queryOptions = reactive({
--- a/src/services/balancer/contracts/contracts/vault.ts
+++ b/src/services/balancer/contracts/contracts/vault.ts
@@ -22,7 +22,7 @@ import {
   isStablePhantom
 } from '@/composables/usePool';
 import { toNormalizedWeights } from '@balancer-labs/balancer-js';
-import { pick } from 'lodash';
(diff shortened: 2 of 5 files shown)


src/services/balancer/contracts/contracts/vault.ts
```diff
@@ -112,17 +112,34 @@ export default class Vault {
...
+        poolMulticaller.call(
+          `${address}.unwrappedTokenAddress`,
+          wrappedToken,
+          'ATOKEN'
+        );
       });
+
+      const callResult = await poolMulticaller.execute();
+      if (callResult != null) {
+        merge(result.linearPools, callResult);
```
evgenyboxer, line 141: @garethfuller WDYT about this way of merging the results? They both share the same `key` (the address), so I think its safe to do it.

evgenyboxer:
I've named the ABI file: `StaticATokenLM.json` as the name of the contract appears on etherscan

src/services/balancer/contracts/contracts/vault.ts
garethfuller: Can you not just modify the `result` like we're doing elsewhere? I think execute already handles this kind of merging of data.
```ts
Object.entries(wrappedTokensMap).forEach(([address, wrappedToken]) => {
  poolMulticaller.call(
    `linearPools.${address}.unwrappedTokenAddress`,
    wrappedToken,
    'ATOKEN'
  );
});

const result = await poolMulticaller.execute(result);
```

garethfuller (commented):
Looks good, just one other thing, in [usePoolQuery](https://github.com/balancer-labs/frontend-v2/blob/feat/evgeny/add-unwrapped-tokens/src/composables/queries/usePoolQuery.ts#L95-L100) could you add an `unwrappedTokens` array too.

src/services/balancer/contracts/contracts/vault.ts
evgenyboxer: Thanks, much cleaner :)

src/composables/queries/usePoolQuery.ts
```diff
@@ -157,8 +159,20 @@ export default function usePoolQuery(
...
 
-    console.log('pool', { onchain: onchainData, ...decoratedPool });
-    return { onchain: onchainData, ...decoratedPool };
+    let unwrappedTokens: Pool['unwrappedTokens'];
+
+    if (isStablePhantomPool && onchainData.linearPools != null) {
+      unwrappedTokens = Object.entries(onchainData.linearPools).map(
+        ([, linearPool]) => linearPool.unwrappedTokenAddress
+      );
+    }
```
garethfuller, line 168: We're already modifying the pool object with two new arrays of pool tokens in `getLinearPoolAttrs` could you not just move the additional construction of unwrappedTokens here too: https://github.com/balancer-labs/frontend-v2/blob/0f3c9c1569df580b109e4d1563ffbdc7faa4ea5b/src/composables/queries/usePoolQuery.ts#L95-L100

evgenyboxer: Right, but `onchainData ` (where the `unwrappedTokenAddress` is) is computed after the `getLinearPoolAttrs` method. (so it doesn't exist there yet).

So you mean we want to compute `onchainData` before calling `getLinearPoolAttrs` right?

garethfuller: Ahh crap, sorry! You're right. Would be good if we could get these tokens from the subgraph call to the linear pools in `getLinearPoolAttrs` but I guess we can't. We'll have to clean all this up later.

evgenyboxer: Yeah I agree.

Are you happy for this to be merged for now?

garethfuller (approved):

PR #1144, opened 2021-12-01
Adds Aave APR

evgenyboxer:
# Description

Adds AAVE aToken APRs.

Note - there is a new design to how liquidity tooltip works:
Screen Shot 2021-12-02 at 7 09 32 PM

Essentially its always visible - when its LM or boosted it will show the stars icon, otherwise a mutated info icon.

## Type of change

- [ ] Other

## How should this be tested?

- [ ] Hard to test on Kovan, best to emulate the APRs.

## Checklist:


src/components/contextual/pages/pool/PoolStatCards.vue
```diff
@@ -10,7 +10,7 @@
         </div>
         <div class="text-xl font-medium truncate flex items-center">
           {{ stat.value }}
-          <LiquidityMiningTooltip :pool="pool" v-if="stat.id === 'apr'" />
+          <LiquidityAPRTooltip :pool="pool" v-if="stat.id === 'apr'" />
```
evgenyboxer, line 13: Its not just about LM now, so the tooltip name changed to something more generic.

src/composables/queries/usePoolQuery.ts
```diff
@@ -99,14 +94,19 @@ export default function usePoolQuery(
...
+
+          linearPoolTokensMap[address] = {
+            ...token,
+            address
+          };
+        });
     });
 
-    pool.linearPoolTokens = linearPoolTokens;
+    pool.linearPoolTokensMap = linearPoolTokensMap;
```
evgenyboxer, line 109: These changes are already included here - https://github.com/balancer-labs/frontend-v2/pull/1121 (awaiting approval), but I needed them here too.

src/lib/config/homestead.json
```diff
@@ -15,6 +15,9 @@
   "subgraph": "https://api.thegraph.com/subgraphs/name/balancer-labs/balancer-v2",
   "poolsUrlV1": "https://storageapi.fleek.co/balancer-bucket/balancer-exchange/pools",
   "poolsUrlV2": "https://storageapi.fleek.co/johngrantuk-team-bucket/poolsV2.json",
+  "subgraphs": {
+    "aave": "https://api.thegraph.com/subgraphs/name/aave/protocol-v2"
+  },
```
evgenyboxer, line 20: I think this approach should give us the flexibility to add multiple subgraphs in the future (and perhaps also move balancer inside at some point)

src/services/aave/subgraph/aave-subgraph.client.ts
```diff
@@ -0,0 +1,30 @@
...
+      console.error(error);
+    }
+  }
+
+  public toPayload(query) {
+    return JSON.stringify({ query: jsonToGraphQLQuery({ query }) });
+  }
+}
+
+export const aaveSubgraphClient = new AaveSubgraphClient();
```
evgenyboxer, line 30: This is a straight copy/paste of the balancer-subgraph client - it was the quickest way to get it working. This part should be refactored later on to reuse the same logic.

src/services/aave/subgraph/entities/reserves/index.ts
```diff
@@ -0,0 +1,79 @@
...
+
+            total = total.plus(weightedAPR);
+          }
+        });
+      }
+    }
+
+    return {
+      total: total.toString(),
+      tokenBreakdown
```
evgenyboxer, line 76: To test the new tooltip UI you can return something like

```
      total: '0.3',
      tokenBreakdown: {
        "0x0FbdDC06a4720408a2f5Eb78e62Bc31AC6E2A3c4": '0.1',
        "0xE8191aACfcdb32260Cda25830Dc6C9342142f310": '0.1',
        "0x4811A7Bb9061A46753486E5e84B3CD3D668fB596": '0.1'
      }
```

src/services/balancer/subgraph/entities/pools/index.ts
```diff
@@ -1,28 +1,37 @@
...
+import { bnum } from '@/lib/utils';
+import {
+  currentLiquidityMiningRewards,
+  computeTotalAPRForPool,
+  computeAPRsForPool
+} from '@/lib/utils/liquidityMining';
+
+import { Network } from '@/composables/useNetwork';
+import { isStable, isStablePhantom, isWstETH } from '@/composables/usePool';
+import { oneSecondInMs, twentyFourHoursInSecs } from '@/composables/useTime';
```
evgenyboxer, line 22: ^ just some resort of imports

src/types/index.ts
```diff
@@ -60,3 +60,11 @@ export type Step = {
...
   state: StepState;
 };
+
+export type Address = string;
+export type QueryArgs = Record<string, any>;
+export type QueryAttrs = Record<string, any>;
+export type QueryBuilder = (
+  args?: QueryArgs,
+  attrs?: QueryAttrs
+) => Record<string, any>;
```
evgenyboxer, line 70: @garethfuller is this a good place for them? Right now only the aave one uses it, if you are happy with it I can change the balancer one too.

src/services/aave/subgraph/entities/reserves/index.ts
```diff
@@ -0,0 +1,79 @@
...
+          underlyingAsset_in: pool.mainTokens,
+          isActive: true,
+          aEmissionPerSecond_gt: 0
+        }
+      });
+
+      if (reserves.length > 0) {
+        reserves.forEach(reserve => {
+          // @ts-ignore
+          const token = getAddress(reserve.aToken.underlyingAssetAddress);
```
evgenyboxer, line 52: Actually this is wrong... I just realised.

Essentially, given a `mainToken` I need to get the corresponding `wrappedToken`

src/services/aave/subgraph/entities/reserves/index.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  public async get(args = {}, attrs = {}): Promise<ComputedReserveData[]> {
+    const query = this.query(args, attrs);
+    const { reserves } = await this.service.client.get(query);
+
+    return v2.formatReserves(reserves);
+  }
+
+  public async calcWeightedSupplyAPRFor(
```
garethfuller, line 32: This function seems like it shouldn't be in the subgraph service. It looks like it's a more general Aave service function, whereas this file is specifically about interacting with the reserves entity or formatting the responses. I would move it to a `/services/aave/aave.service.ts` class that has the subgraph service as an injected dependency.

src/services/aave/subgraph/entities/reserves/index.ts
```diff
@@ -0,0 +1,79 @@
...
+    pool: Pool,
+    prices: TokenPrices,
+    currency: FiatCurrency
+  ) {
+    let total = bnum(0);
+    const tokenBreakdown = {};
+
+    if (pool.wrappedTokens != null) {
+      const reserves = await this.service.reserves.get({
```
garethfuller, line 41: If you do move this function, this will need to be changed anyway, but you could have just done:
```suggestion
      const reserves = await this.get({
```

src/types/index.ts
garethfuller: Sure, yeah fine to add them here.

garethfuller (approved):
Maybe some minor restructuring is required for the aave calc function (as per my comment) but otherwise looks good.

src/services/aave/subgraph/entities/reserves/index.ts
evgenyboxer: Good idea!

PR #1154, opened 2021-12-01
Search for tokens through all tokenlists

TomAFrench:
# Description

Currently when searching for tokens we only look in the currently active tokenlists. This hampers discoverability as it requires us to have listed a token in order for it to show up when the user is looking for it (without the user having to go out of their way to enable the 3rd party tokenlists first).

This PR means that as long as the token exists _somewhere_ in one of our tokenlists then the user will be able to find and inject it.

As a result of this I've removed some of the larger tokenlists as they included a lot of dross and weren't generally helpful. In future we want to add a semi-greyed-out state for tokens which are being loaded from external token lists until the user has injected them to show the distinction.

## Type of change

- [ ] Other

## How should this be tested?

Search for the name of a token which is not included in the main Balancer tokenlist but in other tokenlists on the relevant network, the token should appear for selection.

## Checklist:

### DIFF
--- a/src/providers/tokens.provider.ts
+++ b/src/providers/tokens.provider.ts
@@ -142,6 +142,16 @@ export default {
      */
 
     /**
+     * All tokens from all token lists.
+     */
+    const allTokenListTokens = computed(
+      (): TokenInfoMap => ({
+        ...mapTokenListTokens(Object.values(allTokenLists.value)),
+        ...state.injectedTokens
+      })
+    );
+
+    /**
      * All tokens from token lists that are toggled on.
      */
     const activeTokenListTokens = computed(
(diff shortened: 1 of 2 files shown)


src/providers/tokens.provider.ts
```diff
@@ -168,6 +168,11 @@ export default {
       ...state.injectedTokens
     }));
 
+    const allTokens = computed(() => ({
+      ...mapTokenListTokens(Object.values(allTokenLists.value)),
+      ...state.injectedTokens
+    }));
```
garethfuller, line 174: Could you move this up above `activeTokenListTokens` and call it `allTokenListTokens` and also define it's type just so it's more clear like some of the others (I see we haven't done that for `tokens` so maybe you could add the type there too).
```suggestion
    const allTokenListTokens = computed((): TokenInfoMap => ({
      ...mapTokenListTokens(Object.values(allTokenLists.value)),
      ...state.injectedTokens
    }));
```

garethfuller (approved):
Thanks, looks good, the less crazy the token lists, the better!

PR #1173, opened 2021-12-07
Linear pool tweaks

evgenyboxer:
# Description

- Removal of "totals" in Pool composition
- Copy change of pool title
- Alignment of the BalBreakdown component

Fixes # (issue)

## Type of change

- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist:

### DIFF
--- a/src/components/_global/BalBreakdown/BalBreakdown.vue
+++ b/src/components/_global/BalBreakdown/BalBreakdown.vue
@@ -9,6 +9,8 @@ type Props = {
   title?: string;
   hideItems?: boolean;
   size?: string;
+  offsetClassOverrides?: string;
+  initVertBarClassOverrides?: string;
 };
 
 /**
@@ -24,6 +26,10 @@ const props = withDefaults(defineProps<Props>(), {
  * COMPUTED
  */
 const initVertBarSizes = computed(() => {
+  if (props.initVertBarClassOverrides != null) {
+    return props.initVertBarClassOverrides;
+  }
+
   switch (props.size) {
     case 'lg':
       return 'h-8 -mt-8';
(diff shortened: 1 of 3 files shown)


src/components/_global/BalBreakdown/BalBreakdown.vue
```diff
@@ -9,6 +9,8 @@ type Props = {
   title?: string;
   hideItems?: boolean;
   size?: string;
+  offsetClassOverrides?: string;
+  initVertBarClassOverrides?: string;
```
evgenyboxer, line 13: @garethfuller I wasn't sure what's the best practice in Vue, but how do I pass "custom" overrides? It seems that to get close to what Pon has in his designs I need something very specific.

garethfuller: I don't think overrides are the way to go here. It looks like you need to add `md` size class option to `initVertBarSizes` to cover your `h-6 -mt-6` case. And you need some prop for a left offset maybe?

garethfuller (changes requested):

src/components/_global/BalBreakdown/BalBreakdown.vue
evgenyboxer: I like the direction, but still unsure of correct implementation.

Adding something like `leftOffset` would work, but I need two more things:
1. Having "mt-4" as an `offsetClass`.
2. Having "h-6 -mt-6" as an `initVertBarSizes`.

It seems that one is "xl" and the other is indeed "md", but you can't really mix/match unless we have something like a sizing object where you can pick the sizes for each?

Something like

```
size: {
  vertBar: 'md',
  offset: 'xl',
  ...
}
```

garethfuller (approved):

PR #1174, opened 2021-12-07
update: add ability to save pool creation state and recover when retu…

1saf:
# Description

Currently we have a problem with pool creation in that there is no recoverability mechanism built into the flow. This PR introduces that ability to recover previous state for the user to return to where they were before.

The main concept is that once the user reaches the PreviewPool stage of pool creation (the step where the buttons for creation and seeding are) the UI will save the pool creation state to the users local storage. The creation state has a new flag called `needsSeeding` which is set after a successful create action.

So the two main points of recovery are: 
- user reaches the final step of creation and doesn't create the pool. the state is saved and the user can return to this
- the user creates the pool but doesn't seed liquidity, or the seed liquidity transaction fails, the user can refresh or come back to this page and it'll recover and allow them to fund the pool

## Type of change

- [ ] Other

## How should this be tested?

1. Go to pool creation, go through the flow as you would and once you reach the final step, exit out and open up the UI again. It should open up to the last step with the inputs you had selected before
2. Go through test 1, but this time create the pool - then exit out. Coming back to the UI it should go to the last step and the only action should be to fund the pool
3. 
## Checklist:

### DIFF
--- a/src/composables/pools/usePoolCreation.ts
+++ b/src/composables/pools/usePoolCreation.ts
@@ -1,18 +1,24 @@
-import { PoolType } from '@/services/balancer/subgraph/types';
-import { flatten, sumBy } from 'lodash';
-import { ref, reactive, toRefs, watch, computed } from 'vue';
+import { ref, reactive, toRefs, watch, computed, onMounted } from 'vue';
+
 import { useI18n } from 'vue-i18n';
 import usePoolsQuery from '@/composables/queries/usePoolsQuery';
 import useTransactions from '@/composables/useTransactions';
 import useEthers from '@/composables/useEthers';
 import useTokens from '../useTokens';
 import useWeb3 from '@/services/web3/useWeb3';
-import { configService } from '@/services/config/config.service';
-import { balancerService } from '@/services/balancer/balancer.service';
-import { bnum, scale } from '@/lib/utils';
+
 import BigNumber from 'bignumber.js';
+import { flatten, sumBy } from 'lodash';
+import { bnum, lsGet, lsRemove, lsSet, scale } from '@/lib/utils';
+
+import { PoolType } from '@/services/balancer/subgraph/types';
+import { balancerService } from '@/services/balancer/balancer.service';
+import { configService } from '@/services/config/config.service';
 import { TransactionResponse } from '@ethersproject/providers';
 import { POOLS } from '@/constants/pools';
+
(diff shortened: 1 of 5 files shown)


garethfuller:
@1saf the THX team actually failed because the create pool action was never confirmed in our UI. Is there any way we can handle this? One potential option would be to store the tx hash for the create pool transaction in LS under some kind of `unconfirmedPoolCreation` flag and remove it once we know it's confirmed.

src/pages/pool/create.vue
```diff
@@ -59,16 +65,31 @@ const {
...
 
 onMounted(() => {
+  removeAlert('return-to-pool-creation');
   if (accordionWrapper.value) {
     anime.set(accordionWrapper.value, {
       opacity: 0
     });
   }
+
+  let previouslySavedState = lsGet(
```
timjrobinson, line 83: I think this code should go in usePoolCreation so it's alongside the saveState code which is similar. Then the key and version doesn't need to be known outside usePoolCreation.

src/composables/pools/usePoolCreation.ts
```diff
@@ -91,6 +99,21 @@ export default function usePoolCreation() {
...
+    // before changing the step. Changing it immediately
+    // results in a blank page being rendered.
+    // The hasRestoredFromSavedState flag is also used to
+    // hide the mount of the first step so it 'seems' like
+    // that the preview page is mounted first.
+    setTimeout(() => {
+      if (hasRestoredFromSavedState.value) {
+        setActiveStep(4);
+      }
+    }, 3500);
```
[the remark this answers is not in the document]
1saf: yeah, had a discussion with gareth and this isnt ideal. will find another way around it

src/pages/pool/create.vue
1saf: unfortunately this cannot be the case since this logic needs to be inside create.vue. If called inside usePoolCreation, it means all the other usages of usePoolCreation in the other children components will make this call again and conflict with the restore

1saf:
> @1saf the THX team actually failed because the create pool action was never confirmed in our UI. Is there any way we can handle this? One potential option would be to store the tx hash for the create pool transaction in LS under some kind of `unconfirmedPoolCreation` flag and remove it once we know it's confirmed.

Yep

PR #1185, opened 2021-12-09
Fiat price fix for "My pool balance"

evgenyboxer:
# Description

Apparently the fix is quicker than I thought... I initially implemented the "latestPrices" subgraph query only to find out that this is the existing `priceRate` we already used.

Since total was calculated correctly, all what was needed is to use `fiatAmounts` (which already calcs the fiat amount)

## Type of change

- [ ] Other

## How should this be tested?

- [ ] Check "My pool balance" - total should match the sum of all the pool tokens.

## Checklist:

### DIFF
--- a/src/components/contextual/pages/pool/MyPoolBalancesCard.vue
+++ b/src/components/contextual/pages/pool/MyPoolBalancesCard.vue
@@ -26,11 +26,12 @@ const props = defineProps<Props>();
  */
 const {
   fiatTotalLabel,
+  fiatAmounts,
   initMath,
   proportionalPoolTokenAmounts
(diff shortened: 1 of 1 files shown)


garethfuller (changes requested):
So `proportionalMainTokenAmounts` which this solution is based on has changed in the withdrawals PR. It will no longer be correct to base this calculation on the `fiatAmounts` array. Once I've merged the withdrawals PR we need to re-think this one. Potentially add a new array to withdrawal math that calculates what we need for this case.

garethfuller:
@evgenyboxer I think I've fixed what this PR was aimed at fixing in the feature base branch.

PR #1208, opened 2021-12-14
Fix AAVE APR breakdown

markusbkoch:
Previous implementation failed because it couldn't find the price of the wrappedToken


evgenyboxer (approved):
lgtm!

PR #1283, opened 2022-01-07
Hotfix/1.35.15

1saf:
# Description

When creating a new pool and reaching the stage when you need to add initial liquidity, there was an issue where the Preview screen would be greyed out for no reason. (See image)

You can see that the user has balances for both ETH and WAMPL, they should be allowed to proceed with pool creation.
The issue here is that when the UI is checking for invalid balances, it checks against the token addresses that are selected for the pool. So in this case WETH/WAMPL. 

So in this case, the user has ETH balance but not WETH balance, if you were to switch to WETH via the dropdown it would show the exceeding balance error. This shouldn't be an issue anyways since we allow creation using native assets. 

I've fixed this by making sure the invalid balance check uses the token addresses that are used on the initial liquidity page rather than the initial token ones.

image

Fixes # (issue)

## Type of change

- [ ] Other

## How should this be tested?

To test, create a pool which uses WETH as a token. Make sure you have an ETH balance and no WETH balance. Then when you go to create that pool, make sure you can proceed with just an ETH balance and the other token.

## Checklist:

### DIFF
--- a/src/components/cards/CreatePool/InitialLiquidity.vue
+++ b/src/components/cards/CreatePool/InitialLiquidity.vue
@@ -22,7 +22,13 @@ const cardWrapper = ref<HTMLElement>();
  * COMPOSBALES
  */
 const { userNetworkConfig } = useWeb3();
-const { balanceFor, priceFor, nativeAsset, wrappedNativeAsset } = useTokens();
+const {
+  balances,
+  balanceFor,
+  priceFor,
+  nativeAsset,
+  wrappedNativeAsset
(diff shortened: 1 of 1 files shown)


garethfuller (approved):

timjrobinson (approved):

1saf:
Closing in favour of the new hotfix branch just to keep it consistent with versions here:
https://github.com/balancer-labs/frontend-v2/pull/1290

Will use admin to merge.

PR #1293, opened 2022-01-11
Pool creation unknown tokens

1saf:
# Description

Currently the pool creation UI does not support arbitrage warnings and optimisations for tokens which may not have a current price on coingecko. This update adds the ability for the user to add in those prices themselves.

## Type of change

- [ ] Other

## How should this be tested?

Here is a custom token: `0xbF3122A7a98a2F9971C339a8511378d853585876`
Basically adding a custom token which doesn't have a price on coingecko to the pool creation token list should trigger a modal to pop up requesting the user to add the price for that token.
Once you enter that price the remainder of the pool creation flow should work as expected.

A quick way to check that the price has been injected correctly is to type paste the token address back into the token select modal and you should see a price there, compared to a '-'

## Checklist:


1saf:
@markusbkoch @pkattera  What happens when a user intends to add BPT as a pool token during pool creation. Do they need to enter in a price? Or does the UI need to handle that case separately? Same goes with other tokens which may represent a share of staked/invested balance from another protocol.

src/locales/default.json
```diff
@@ -112,39 +112,43 @@
     "copyAddress": "Copy address",
     "create": "Create",
     "createAPool": {
-        "title": "Create a pool",
```
1saf, line 115: These needed a sort for maintainability

1saf:
@pkattera @garethfuller Another big question is when the user enters in a price for an unknown token, if they decide to refresh the page or come back at a later time. Should the price they entered remain as the price they entered in before?
What if the 'actual' price of that asset has changed since that last entry time and they are prone to being rekt as a result?

src/providers/tokens.provider.ts
```diff
@@ -198,7 +203,7 @@ export default {
       isLoading: priceQueryLoading,
       isError: priceQueryError,
       refetch: refetchPrices
-    } = useTokenPricesQuery(tokenAddresses);
+    } = useTokenPricesQuery(tokenAddresses, injectedPrices);
```
garethfuller, line 206: Like we do for the allowance contracts.
```suggestion
    } = useTokenPricesQuery(tokenAddresses, toRef(state, 'injectedPrices'));
```

src/components/modals/UnknownTokenPrice/UnknownTokenPriceModal.vue
```diff
@@ -0,0 +1,113 @@
...
+      </span>
+      <BalStack isDynamic vertical>
+        <TokenInput
+          v-for="(address, i) in unknownTokens"
+          :key="i"
+          fixedToken
+          placeholder="$0.00"
+          v-model:amount="unknownTokenPrices[address]"
+          v-model:address="
+            seedTokens[getIndexOfUnknownToken(address)].tokenAddress
```
timjrobinson, line 97: Why is this needed instead of just passing `address`? Same with on line 101.

garethfuller (changes requested):
Works really nicely and looks good! 🙌 Just a couple of minor comments.

src/lib/utils/index.ts
```diff
@@ -113,3 +114,17 @@ export async function forChange(
 export function bnSum(amounts: string[]): BigNumber {
   return amounts.reduce((a, b) => bnum(a).plus(b), bnum(0));
 }
+
+/**
+ *  creates a sentence readable list for an array for string
+ *  e.g. 'UBT, DSB and BAL' for [UBT, DSB, BAL]
+ */
+export function formatWordListAsSentence(words: string[]) {
```
timjrobinson, line 122: This should have a test to ensure it behaves as you expect in different cases.

garethfuller:
> @pkattera @garethfuller Another big question is when the user enters in a price for an unknown token, if they decide to refresh the page or come back at a later time. Should the price they entered remain as the price they entered in before? What if the 'actual' price of that asset has changed since that last entry time and they are prone to being rekt as a result?

Good question. I think we should probably invalidate custom prices if they leave the flow. It looks like that's how it works already? The one other thing I think should be possible is if they want to update the price after entering it for the first time. There doesn't seem to be a mechanism for that?

src/components/modals/UnknownTokenPrice/UnknownTokenPriceModal.vue
1saf: It's not possible to bind address to a v-model since it comes from a iterated variable

src/pages/pool/create.vue
```diff
@@ -80,6 +96,21 @@ onMounted(async () => {
 const doSimilarPoolsExist = computed(() => similarPools.value.length > 0);
 const validTokens = computed(() => tokensList.value.filter(t => t !== ''));
 
+const unknownTokens = computed(() => {
+  if (dynamicDataLoading.value) return [];
+  return validTokens.value.filter(token => {
+    return priceFor(token) === 0;
```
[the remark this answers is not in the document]
1saf: This is a difficult one since the value returned from `priceFor` will be 0 if its not there, I'm not sure if we should adjust the existing behaviour and nullify it, if there is no value?
@garethfuller thoughts?

src/lib/utils/index.ts
1saf: > > @pkattera @garethfuller Another big question is when the user enters in a price for an unknown token, if they decide to refresh the page or come back at a later time. Should the price they entered remain as the price they entered in before? What if the 'actual' price of that asset has changed since that last entry time and they are prone to being rekt as a result?
> 
> Good question. I think we should probably invalidate custom prices if they leave the flow. It looks like that's how it works already? The one other thing I think should be possible is if they want to update the price after entering it for the first time. There doesn't seem to be a mechanism for that?

No, but I will discuss with pon

1saf: added those tests

timjrobinson (approved):

src/components/modals/UnknownTokenPrice/UnknownTokenPriceModal.vue
```diff
@@ -0,0 +1,113 @@
...
+});
+
+const isSubmitDisabled = computed(() => {
+  return props.unknownTokens.some(token =>
+    [null, ''].includes(unknownTokenPrices[token])
+  );
+});
+
+/**
+ * FUNCTIONS
```
evgenyboxer, line 59:
```suggestion
 * METHODS
```

for consistency

src/components/modals/UnknownTokenPrice/UnknownTokenPriceModal.vue
```diff
@@ -0,0 +1,113 @@
...
+
+function injectUnknownPrices() {
+  injectPrices(unknownTokenPrices);
+  emit('close');
+}
+</script>
+
+<template>
+  <BalModal
+    title="Unknown token price"
```
evgenyboxer, line 73: ^ Missing translation

src/lib/utils/index.ts
```diff
@@ -113,3 +114,17 @@ export async function forChange(
...
+  if (!words.length) return '';
+  if (words.length >= 2) {
+    const commaSeperatedWords = initial(words);
+    return `${commaSeperatedWords.join(', ')} and ${words[words.length - 1]}`;
```
evgenyboxer, line 126: The word `and` would need to be translated, I think the quickest way is probably to pass `t` here and translate it.

src/lib/utils/utils.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+    });
+
+    it('should return just the first word if there is only one word in the list', () => {
+      const words = ['USD'];
+      const formatted = formatWordListAsSentence(words);
+
+      expect(formatted).toStrictEqual('USD');
+    });
+
+    it('should return nothing if there are no words', () => {
```
evgenyboxer, line 19:
```suggestion
    it('should return and empty string if there are no words', () => {
```

timjrobinson (approved):

timjrobinson:
Bug: If you enter too large of a number it goes to NaN, and before then the formatting goes wonky. Should have a max value.

timjrobinson (approved):

timjrobinson (changes requested):
It's still allowing me to save the very large numbers, I see the warning but can save anyway.

src/components/modals/UnknownTokenPrice/UnknownTokenPriceModal.vue
garethfuller: It's because you can't mutate props and unknownTokens is a prop. Because you are not modifying the address here you can just not use v-model, I think `:address="address"` will just work.

garethfuller (approved):

timjrobinson (approved):

PR #1294, opened 2022-01-11
Fix pool symbol and pool name disappearing

1saf:
# Description

Was seeing the 'Missing pool name or prices' tooltip pop-up where it shouldn't have and after some deeper investigation, found the root cause for the problem. This hotfix addresses that by making sure that the pool name and symbol (autogenerated) are saved BEFORE loading the final step as well as make sure that the pool name and symbol are saved on click of the 'save' button rather than relying just on the update of the model value which seemed to be hit and miss.

## Type of change

- [ ] Other

## How should this be tested?

Create a pool as normal, make sure to get to the final step. 
1. On the final step, refresh the page. It should keep the pool name and symbol as to what it was previously (autogenerated one)
2. Adjust the pool name and symbol yourself, then refresh the page. It should keep those.
3. On the final step, empty out the value for either pool name or symbol or both and make sure that an alert shows up and also disables the create pool button.

## Checklist:

### Walkthrough

(Note, you will notice in the video that on refresh of the page, the symbols for the tokens dissapear! This was an injection bug which has since been fixed)
https://www.loom.com/share/b64e4de4606c4238895910897623d2fe


src/composables/pools/usePoolCreation.ts
```diff
@@ -92,8 +92,12 @@ export default function usePoolCreation() {
         w => w.tokenAddress
       );
 
-      poolCreationState.name = getPoolSymbol();
-      poolCreationState.symbol = getPoolSymbol();
```
1saf, line 96: This overrides the custom name or symbol that the user might have entered

timjrobinson (approved):
LGTM

EndymionJkb (approved):
Added the new key translations

src/composables/pools/usePoolCreation.ts
```diff
@@ -92,8 +92,12 @@ export default function usePoolCreation() {
...
+      if (!poolCreationState.name) {
+        poolCreationState.name = getPoolSymbol();
+      }
```
garethfuller, line 97: Is it possible to clean this up a bit with?
```suggestion
      !poolCreationState.name && poolCreationState.name = getPoolSymbol();
```
Same for conditional below if it works.

garethfuller (approved):
Looks good! Again just a couple of minor comments.

src/composables/pools/usePoolCreation.ts
timjrobinson: Could also be `poolCreationState.name = poolCreationState.name || getPoolSymbol()`

src/components/_global/BalInlineInput/BalInlineInput.vue
```diff
@@ -98,6 +99,11 @@ const inputAttrs = computed(() => omit(attrs, 'class'));
  */
 function toggleEditable() {
   isEditable.value = !isEditable.value;
+  // if the value was changed to false it means that
+  // the user clicked the save icon
+  if (isEditable.value === false) {
+    emit('save');
+  }
```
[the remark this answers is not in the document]
1saf: done

PR #1320, opened 2022-01-19
Pool creation minor fixes

1saf:
# Description

Few fixes for pool creation:
- A bit of cleanup on styling
- A bit of cleanup on code reusability
- Add in a warning for when a pool token has an initial weight less than 1% which may cause funding to fail
- Add in a validation so you cannot continue to previewing a pool with a 0 balance

## Type of change

- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## How should this be tested?

1. Most important test is to make sure to test that pool creation funding doesn't fail. To do this go through with pool creation and try to create a pool with a token of which you have no balance OR you try to continue to create where you've just entered 0 for initial liquidity of that token. Shouldn't be allowed to get to the create step.

2. Once you've reached the create and funding steps, IF your balance choices for the tokens have a token where the initial weight is less than 1%. That tokens initial weight should be highlighted with a yellow colour and a warning should pop up when you get to the funding step.

3. Make sure to also test the case where you need to refresh the page and come back to funding. It should restore perfectly and you should be able to continue with funding.

## Visual context

That alert disappearing has been fixed since and also the fact that it should only show on the fund step.
https://www.loom.com/share/768eb671f8ed4e8b8c0b732314700006

## Checklist:


garethfuller (approved):
Looks good. Thanks for making the PR super easy to review! 🙌

timjrobinson (approved):

PR #1321, opened 2022-01-19
Refactor fNum to allow for Internationalisation / Multiple Currencies

timjrobinson:
# Description

I've created a new function [fNum2](https://github.com/balancer-labs/frontend-v2/pull/1321/files#diff-8b279947dd8a9ce4af0f314eab942df05750b46dbc2aa4fe71b65d00d33c46a6R90) that uses Javascripts native [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) to format numbers instead of the old numeral package. 

Nothing should be changed UX wise in this PR. It's purely a code refactor. To ensure nothing has changed I created a series of tests that have an old fNum call and an fNum2 equivalent that runs through many test cases to ensure the results are always the same. 

fNum2 has several advantages over the old fNum:

- Users can change their locale and have both numbers and values show the correct format
- Users can change currency and have it show in the correct format for that currency
- Users can combine any locale / currency they wish. 
- We don't have custom formats that change based on the size of the number, it can auto-adjust on the fly.
- Because each usage has it's own specification and the same presets aren't used all over the place, it's more easy to change the formatting of one section of Balancer without accidentally breaking other areas. 

In a future PR I'll remove fNum and rename fNum2 to fNum. I didn't want to do it in this PR so that we can ensure all changes are correct first. 

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- For each code change, have a look at the relevant unit test and ensure that the new code is the correct equivalent code. 
- Use Balancer and ensure all 

## Checklist:

### DIFF
--- a/src/components/contextual/pages/pool/PoolBalancesCard/Pool.vue
+++ b/src/components/contextual/pages/pool/PoolBalancesCard/Pool.vue
@@ -84,7 +84,7 @@ export default defineComponent({
     /**
      * COMPOSABLES
      */
-    const { fNum } = useNumbers();
+    const { fNum2 } = useNumbers();
     const { explorerLinks } = useWeb3();
     const { t } = useI18n();
     const { upToLargeBreakpoint } = useBreakpoints();
@@ -151,12 +151,19 @@ export default defineComponent({
 
     function balanceFor(address: string): string {
       if (!pool || !pool.value) return '-';
-      return fNum(pool.value.onchain.tokens[address].balance, 'token');
+      return fNum2(pool.value.onchain.tokens[address].balance, {
+        maximumFractionDigits: 4
+      });
     }
 
     function weightFor(address: string): string {
       if (!pool || !pool.value) return '-';
-      return fNum(pool.value.onchain.tokens[address].weight, 'percent');
+      return fNum2(pool.value.onchain.tokens[address].weight, {
+        style: 'unit',
+        unit: 'percent',
+        minimumFractionDigits: 2,
+        maximumFractionDigits: 2
+      });
     }
 
     function fiatValueFor(address: string): string {
@@ -164,7 +171,7 @@ export default defineComponent({
       if (!pool || !pool.value || price === 0) return '-';
 
       const balance = Number(pool.value.onchain.tokens[address].balance);
-      return fNum(balance * price, 'usd');
+      return fNum2(balance * price, { style: 'currency' });
     }
 
     return {
@@ -172,7 +179,7 @@ export default defineComponent({
       balanceFor,
       weightFor,
       fiatValueFor,
-      fNum,
+      fNum2,
       explorer: explorerLinks,
       columns,
       tableData,
--- a/src/components/contextual/pages/pool/PoolBalancesCard/Pool.vue
+++ b/src/components/contextual/pages/pool/PoolBalancesCard/Pool.vue
@@ -57,7 +57,7 @@
 
 <script lang="ts">
 import { PropType, defineComponent, toRefs, computed, Ref } from 'vue';
-import useNumbers from '@/composables/useNumbers';
+import useNumbers, { FNumFormats } from '@/composables/useNumbers';
 import { useI18n } from 'vue-i18n';
 import { FullPool } from '@/services/balancer/subgraph/types';
 import numeral from 'numeral';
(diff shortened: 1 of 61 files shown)


src/composables/useNumbers.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+          const token = Object.values(mockTokens).find(
+            token => address === token.address
+          );
+          return token?.price[currency];
+        })
+    };
+  });
+});
+
+describe('useNumbers', () => {
```
garethfuller, line 33: This is great!

src/composables/useNumbers.ts
```diff
@@ -1,17 +1,37 @@
...
+  percent: {
+    style: 'unit',
+    unit: 'percent',
+    minimumFractionDigits: 2,
+    maximumFractionDigits: 2
+  },
+  token: {
+    maximumFractionDigits: 4
+  }
+};
```
garethfuller, line 26: As mentioned above I think we could add a fiat preset:
```suggestion
export const FNumFormats = {
  percent: {
    style: 'unit',
    unit: 'percent',
    minimumFractionDigits: 2,
    maximumFractionDigits: 2
  },
  token: {
    maximumFractionDigits: 4
  },
  fiat: {
    style: 'currency'
  }
};
```

garethfuller (changes requested):
Nice work! 🙌  just a few minor comments.

src/components/contextual/pages/pool/PoolBalancesCard/Pool.vue
```diff
@@ -151,28 +151,33 @@ export default defineComponent({
 
     function balanceFor(address: string): string {
       if (!pool || !pool.value) return '-';
-      return fNum(pool.value.onchain.tokens[address].balance, 'token');
+      return fNum2(pool.value.onchain.tokens[address].balance, {
+        maximumFractionDigits: 4
+      });
```
garethfuller, line 156: Could be .token right?

src/components/contextual/pages/pool/PoolTransactionsCard/BoostedPoolActivities/Table.vue
```diff
@@ -285,7 +286,9 @@ function getMainTokenEquivalentAmount(address: string, amount: string) {
                 class="mr-2 flex-shrink-0"
               />
               <span class="font-numeric">{{
-                fNum(action.tokenAmounts[0].amount, 'token')
+                fNum2(action.tokenAmounts[0].amount, {
+                  maximumFractionDigits: 4
+                })
```
garethfuller, line 291: Same here and below

src/components/modals/TradePreviewModalGP.vue
```diff
@@ -48,7 +52,11 @@
               </div>
               <div>
                 <div class="font-medium">
-                  {{ fNum(trading.tokenOutAmountInput.value, 'token') }}
+                  {{
+                    fNum2(trading.tokenOutAmountInput.value, {
+                      maximumFractionDigits: 4
+                    })
```
garethfuller, line 58: And here

garethfuller (approved):
Thanks, seems to be a few more cases where we could use `FNumFormats.token` I think. Otherwise, I think we can merge this.

src/composables/useNumbers.ts
```diff
@@ -62,5 +90,67 @@ export default function useNumbers() {
...
 
-  return { fNum, toFiat };
+  function fNum2(
+    number: number | string,
+    options: FNumOptions | undefined = {}
+  ): string {
+    if (typeof number === 'string') {
+      number = number || 0;
+      number = new BigNumber(number).toNumber();
+    }
```
1saf, line 100: could you leave a comment as to why this is needed

src/composables/useNumbers.ts
```diff
@@ -62,5 +90,67 @@ export default function useNumbers() {
...
+
+    const formatterOptions: Intl.NumberFormatOptions = Object.assign(
+      {},
+      options
+    );
```
1saf, line 105: nit: is it possible to just spread it here to make it more concise (unless there is a reason you are using obj.assign)

1saf (approved):

evgenyboxer (approved):
lgtm!

src/composables/useNumbers.ts
timjrobinson: I thought using `parseFloat` or `Number` may cause issues with floating point precision. I added a test and it seems this isn't the case so made it just use `Number` instead now rather than this.

PR #1335, opened 2022-01-24
Handle withdrawal abs max

garethfuller:
# Description

If the user trying to withdraw from a pool is a significant LP, i.e. they can withdraw more than 30% of the BPT in a pool, then currently the withdrawal form fails on selecting a single asset to withdraw.

This PR fixes the call to fetch the single asset max withdrawal by using the minimum of either the user's full balance or 30% of the BPT in the chosen token as the max value. It also adds a warning to explain why the maximum withdrawable amount is less than the user might be expecting.

This issue was less relevant before we launched pool creation. However, with pool creation live it's likely more users will run into this problem.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Create a pool so you are the sole LP (or go to a pool you have previously created) and try to withdraw in a single asset. You should notice that the max withdrawable amount is not the full pool balance and you should see a warning.
- [ ] Ensure that single asset maxes work as expected for other pools where you are not the majority LP

## Visual context

[Demo](https://www.loom.com/share/10c507663a504503990065c20f82acbe)

## Checklist:


timjrobinson:
I haven't tested this yet because the graph is still having issues on Kovan. In the loom video though it seemed like you could only withdraw 30% of your DAI, not 30% of the pools value in DAI. As the overall pool value looks to be ~$1k.

1saf:
Bug here:
1. Start withdrawing
2. Choose single asset (I chose WBTC)
3. Dismiss the warning
4. Enter '1' or any input to the withdrawal amount

https://www.loom.com/share/a642f543f48040ec8d5f0f2a130228ee

src/components/forms/pool_actions/WithdrawForm/composables/useWithdrawMath.ts
```diff
@@ -148,11 +149,30 @@ export default function useWithdrawMath(
...
+   */
+  const absMaxBpt = computed((): string => {
+    if (!isWeightedPool.value) return bptBalance.value;
+    // Maximum BPT allowed from weighted pool is 30%
+    const poolMax = bnum(pool.value.totalShares)
+      .times(0.3)
+      .toFixed(poolDecimals.value, OldBigNumber.ROUND_DOWN);
+    // If the user's bpt balance is greater than the withdrawal limit for
+    // weighted pools we need to return the poolMax bpt value.
+    return OldBigNumber.min(bptBalance.value, poolMax).toString();
```
1saf, line 173: oh god, theres an oldbignumber now?

src/components/forms/pool_actions/WithdrawForm/composables/useWithdrawalState.ts
```diff
@@ -21,9 +46,34 @@ const state = reactive({
...
     interval: 1
-  }
+  },
+  error: null
 });
 
+/**
+ * METHODS
+ */
+export function setError(error: WithdrawalError | null): void {
```
1saf, line 56: is this necessary as the error variable is bound two ways when its returned

PR #1340, opened 2022-01-27
feat: implement excluded addresses from totalLiquidity calculation

evgenyboxer:
# Description

Remove excluded addresses from the `totalLiquidity` as defined in the config file - https://github.com/balancer-labs/bal-mining-scripts/blob/master/config/exclude.json (it contains pools per network with a list of addresses to exclude).

I have created a new `rawTotalLiquidity` which preserves the previous value - this is handy for computing user's shares of a pool.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] I was mostly testing it with a mock `exclude.json` by putting pools that I have shares in and look at the pool value (it should deduct the value)
- [ ] Perhaps @markusbkoch can modify this file (temporarily) on Mainnet to exclude some popular addresses to see that it deducts it (and multiple pools).

## Visual context

## Checklist:


src/composables/queries/useUserPoolsQuery.ts
```diff
@@ -221,14 +221,14 @@ export default function useUserPoolsQuery(
...
-          decoratedPools[i].totalLiquidity = totalLiquidity.toString();
+          decoratedPools[i].rawTotalLiquidity = totalLiquidity.toString();
         }
       }
     }
 
     const poolsWithShares = decoratedPools.map(pool => ({
       ...pool,
-      shares: bnum(pool.totalLiquidity)
+      shares: bnum(pool.rawTotalLiquidity)
```
evgenyboxer, line 231: The shares need to be computed with the "raw" value to be accurate.

src/services/balancer/subgraph/entities/pools/index.ts
```diff
@@ -79,9 +88,39 @@ export default class Pools {
...
+    }
+
     return this.serialize(pools, pastPools, period, prices, currency);
   }
 
+  public removeExcludedAddressesFromTotalLiquidty(
+    pool: Pool,
+    totalLiquidity: string
+  ) {
+    let _totalLiquidity = bnum(totalLiquidity);
```
evgenyboxer, line 104: This is a handy function to apply the removal of addresses externally (when having a custom `totalLiquidity` like stable pools).

For simplicity - it assumes the decoration phase already ran and so `this.excludedAddresses` is defined.

src/services/balancer/subgraph/entities/pools/index.ts
```diff
@@ -97,7 +136,11 @@ export default class Pools {
       pool.address = this.addressFor(pool.id);
       pool.tokenAddresses = pool.tokensList.map(t => getAddress(t));
       pool.tokens = this.formatPoolTokens(pool);
-      pool.totalLiquidity = poolService.calcTotalLiquidity(prices, currency);
+      pool.rawTotalLiquidity = poolService.calcTotalLiquidity(prices, currency);
```
evgenyboxer, line 139: preserve a "raw" value to be used for things like computing user shares.

src/services/balancer/subgraph/entities/pools/index.ts
```diff
@@ -151,6 +194,43 @@ export default class Pools {
     return Promise.all(promises);
   }
 
+  private async getExcludedAddresses() {
+    try {
+      const { data } = await axios.get<ExcludedPoolAddressesResponse>(
+        'https://raw.githubusercontent.com/balancer-labs/bal-mining-scripts/master/config/exclude.json'
```
evgenyboxer, line 200: @markusbkoch are you happy for me to use this URL?

src/services/balancer/subgraph/entities/pools/index.ts
```diff
@@ -79,9 +88,40 @@ export default class Pools {
...
+    }
+
     return this.serialize(pools, pastPools, period, prices, currency);
   }
 
+  public removeExcludedAddressesFromTotalLiquidity(
+    pool: Pool,
+    rawTotalLiquidity: string
+  ) {
+    let totalLiquidity = bnum(rawTotalLiquidity);
```
evgenyboxer, line 104: @garethfuller since we are going to refactor this later, I've made the decision to simplify and assume this function is executed 

For simplicity - this function assumes it was ran after the decoration phase so that `excludedAddresses` is already available.

src/services/balancer/subgraph/entities/pools/index.ts
markusbkoch: Yes. I've also pushed this with mock data, excluding some whales from a couple of pools on each network for testing purposes if you want to point to it for now
https://raw.githubusercontent.com/balancer-labs/bal-mining-scripts/frontend-exclude-test/config/exclude.json

src/services/balancer/subgraph/entities/pools/index.ts
```diff
@@ -79,9 +88,40 @@ export default class Pools {
...
+      this.excludedAddresses != null &&
+      this.excludedAddresses[pool.address] != null
+    ) {
+      Object.values(this.excludedAddresses[pool.address]).forEach(
+        accountBalance => {
+          const accountBalanceFormatted = formatUnits(accountBalance, 18);
+          const poolShare = bnum(accountBalanceFormatted).div(pool.totalShares);
+
+          totalLiquidity = totalLiquidity.minus(
+            totalLiquidity.times(poolShare)
```
markusbkoch, line 116: We have to multiply the share by the original `totalLiquidity`, not the value we're updating inside the loop. See 62da75b678d8f1905960925333f5404af668b1e7 for a proposed fix

markusbkoch (changes requested):
Thank you Ev. I've left a few comments and submitted a PR to your branch (#1341) with my proposed changes.

src/composables/queries/usePoolsQuery.ts
```diff
@@ -215,15 +215,22 @@ export default function usePoolsQuery(
...
                   .plus(wrappedTokenValue);
               }
             }
           );
 
           decoratedPools[i].onchain = onchainData;
-          decoratedPools[i].totalLiquidity = totalLiquidity.toString();
+          decoratedPools[i].rawTotalLiquidity = rawTotalLiquidity.toString();
+
+          const totalLiquidity = balancerSubgraphService.pools.removeExcludedAddressesFromTotalLiquidity(
```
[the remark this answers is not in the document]
markusbkoch: See how the pool value decreases from 589k to 152k

image

image

garethfuller (changes requested):
As per Markus's review, we only need to subtract from the totalLiquidity value as an input for calculating the LM APR, not in general. Sorry Ev if that wasn't clear before you started on this.

evgenyboxer:
@markusbkoch @garethfuller I've tested it with the mocked ones - LGTM.

Once approved I will switch back to the original file.

src/services/balancer/subgraph/entities/pools/index.ts
```diff
@@ -79,9 +88,41 @@ export default class Pools {
...
+
+    if (this.excludedAddresses == null) {
+      // fetch only once
+      this.excludedAddresses = await this.getExcludedAddresses();
+    }
+
     return this.serialize(pools, pastPools, period, prices, currency);
   }
 
+  public removeExcludedAddressesFromTotalLiquidity(
```
1saf, line 100: was just wondering if for future whether we should be keeping methods like this inside the service or have more as a isolated helper function?

evgenyboxer: Yeah prob could be a good idea - I think we should keep this in mind as part of a larger refactor we want to do around the pools stuff.

markusbkoch:
> Note - for the GRO one, the crazy high APR makes sense right? (our frontend has a `APR_THRESHOLD` for LM which in this case makes it invalid, but I think it's OK in this case?)

@evgenyboxer  Yes the hypothetical GRO APR makes sense because in the mock json I excluded the two largest holders which add up to 99.8% of the pool's liquidity (one is a staking contract). I noticed the stars and APR breakdown are missing from that pool, is it because of the `APR_THRESHOLD` or something else? 
image

garethfuller:
> > Note - for the GRO one, the crazy high APR makes sense right? (our frontend has a `APR_THRESHOLD` for LM which in this case makes it invalid, but I think it's OK in this case?)
> 
> @evgenyboxer Yes the hypothetical GRO APR makes sense because in the mock json I excluded the two largest holders which add up to 99.8% of the pool's liquidity (one is a staking contract). I noticed the stars and APR breakdown are missing from that pool, is it because of the `APR_THRESHOLD` or something else? image

@markusbkoch yes, I think Jeff added a threshold of 100K

markusbkoch:
> @markusbkoch yes, I think Jeff added a threshold of 100K

That should only apply to the swap fee APR, but in practice it should be fine to keep it as is.

markusbkoch (approved):

timjrobinson (approved):
LGTM. Could you add some tests for `removeExcludedAddressesFromTotalLiquidity` to ensure it works as expected.

src/services/balancer/subgraph/entities/pools/index.ts
timjrobinson: Yea +1 to making it a generic `removeAddressesFromTotalLiquidity` and the excluded addresses is passed as the first argument.

src/services/balancer/subgraph/entities/pools/__mocks__/decoratedPool.ts
```diff
@@ -0,0 +1,83 @@
+// import { Pool } from '../../../types';
+
+const decoratedPool: any = {
```
evgenyboxer, line 3: @timjrobinson any ideas why if I import the `Pool` type jest dies? It complains about the defined `enum` there... perhaps its not compatible with jest quite yet?

src/services/balancer/subgraph/entities/pools/helpers.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+      totalLiquidity: 2000
+    };
+
+    it('Should correctly remove excluded addresses', () => {
+      const excludedAddresses: ExcludedAddresses = {
+        [mockedDecoratedPool.address]: {
+          '0x359f4fe841f246a095a82cb26f5819e10a91fe0d': parseUnits('250'),
+          '0x001c249c09090d79dc350a286247479f08c7aad7': parseUnits('250')
+        }
+      };
```
evgenyboxer, line 24: The idea here is to remove `500` shares of the `1000` available ones which should result in a reduction of 50% of the `totalLiquidity` (AKA the `miningTotalLiquidity`)

src/services/balancer/subgraph/entities/pools/helpers.spec.ts
```diff
@@ -0,0 +1,70 @@
+import { parseUnits } from '@ethersproject/units';
+
+import decoratedPool from './__mocks__/decoratedPool';
```
timjrobinson, line 3: Mocks shouldn't be imported like this. This should be `jest.mock('./decoratedPool')` then whenever that file is imported in any other file during this test the mock will be imported instead.

src/services/balancer/subgraph/entities/pools/__mocks__/decoratedPool.ts
evgenyboxer: nvm its working now.

1saf (approved):

timjrobinson (approved):

PR #1395, opened 2022-02-10
Locking & Voting (Tokenomics)

evgenyboxer:
# Description

This is the base branch for the veBAL page + getVeBAL modal.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:

### DIFF
--- a/src/components/forms/lock_actions/LockForm/components/VeBalForm/components/LockEndDate.vue
+++ b/src/components/forms/lock_actions/LockForm/components/VeBalForm/components/LockEndDate.vue
@@ -47,6 +47,7 @@ onBeforeMount(() => {
       v-model="lockEndDate"
       :min="format(minLockEndDateTimestamp, INPUT_DATE_FORMAT)"
       :max="format(maxLockEndDateTimestamp, INPUT_DATE_FORMAT)"
+      step="7"
     />
   </div>
 </template>
--- a/src/components/forms/lock_actions/LockForm/components/VeBalForm/components/LockEndDate.vue
+++ b/src/components/forms/lock_actions/LockForm/components/VeBalForm/components/LockEndDate.vue
@@ -1,7 +1,9 @@
 <script setup lang="ts">
+import { computed, onBeforeMount } from 'vue';
+import { useI18n } from 'vue-i18n';
+import { format, addWeeks } from 'date-fns';
+
 import { VeBalLockInfo } from '@/services/balancer/contracts/contracts/veBAL';
-import { onBeforeMount } from '@vue/runtime-core';
-import { format } from 'date-fns';
 
 import useLockState from '../../../composables/useLockState';
 
@@ -17,6 +19,12 @@ type Props = {
 };
 
 /**
+ * COMPOSABLES
+ */
+
+const { t } = useI18n();
+
+/**
  * PROPS
  */
 const props = defineProps<Props>();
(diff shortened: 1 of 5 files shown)


src/composables/queries/useAllowancesQuery.ts
```diff
@@ -49,11 +49,13 @@ export default function useAllowancesQuery(
...
   const queryFn = async () => {
     console.log('Fetching', tokenAddresses.value.length, 'allowances');
-    return await tokenService.allowances.get(
+    const allowances = await tokenService.allowances.get(
       account.value,
       contractAddresses.value,
       tokens.value
     );
+
+    return allowances;
```
garethfuller, line 58: I think this was probably to log allowances right? Can we revert to original please.

src/plugins/router.ts
```diff
@@ -51,12 +53,6 @@ const routes: RouteRecordRaw[] = [
     component: PoolWithdrawPage,
     meta: { layout: 'PoolTransferLayout' }
   },
-  {
-    path: '/pool/migrate/:from/:to',
-    name: 'migrate-pool',
-    component: MigratePoolPage,
-    meta: { layout: 'FocusedLayout' }
-  },
```
garethfuller, line 59: Why has the migration path been deleted here?

garethfuller (commented):
This is my first pass. Code looks great, really clean and well structured.

I talked with Pon/Fernando about the term 'Unlock', lets use the word 'Redeem' instead. Please replace all references to Unlock in component names, labels and routes.

garethfuller:
@evgenyboxer when unlocking, the link back to the vebal page is broken.

garethfuller:
@evgenyboxer when my lock expired, I was unable to extend the lock again, it just kept reverting the tx. I had to unlock and then relock.

markusbkoch:
The veBAL dashboard tells me I have no BPT to lock on this address 0x67dc6f6ffa7De114c33144f3b1630Fd1Eecd27dD, even though I do
image

The lock UI does recognize them
image

The only relevant thing I can think of is that I just withdrew BPT from the veBAL contract, so maybe there's an unhandled edged case in the dashboard that causes it to show like this? cc @garethfuller

timjrobinson:
@markusbkoch That is fixed, was just not showing the BPT value unless you had veBAL locked.

src/composables/useTime.ts
```diff
@@ -24,6 +28,16 @@ export function toJsTimestamp(unixTimestamp: number): number {
   return unixTimestamp * oneSecondInMs;
 }
 
+export function getPreviousThursday(date: Date): Date {
+  const daysSinceThursday = (date.getDay() - 4) % 7;
```
TomAFrench, line 32:
```suggestion
  const daysSinceThursday = (date.getDay() - 4 + 7) % 7;
```

This doesn't work for Mon-Wed currently as it doesn't handle negatives correctly so we need to bump up by 7

TomAFrench (changes requested):

src/composables/useTime.ts
TomAFrench: Granted, I think this bug got added by copy-pasting my snippet.

garethfuller:
> Is this working correctly? total at the bottom says 0.0091 vs actual says 0.0086 veBAL
> 
> Screen Shot 2022-03-20 at 4 18 24 PM

@zekraken-bot the Total voting escrow value in the form is what you get after you've locked. The value on the right is your current veBAL. If this is not clear enough, we may need to change labels. cc @pkattera

PR #1418, opened 2022-02-24
fix wrong labels on PreviewPool

acryptosx:


garethfuller (approved):
Looks like a reasonable change to me. I'll let @1saf make the final call and merge.

1saf (approved):

PR #1460, opened 2022-03-10
Add a composable for making flexible subgraph queries

1saf:
# Description

Currently our subgraph client/server architecture doesn't allow us to make more flexible on the go subgraph queries. This pull request introduces a composable which allows us to make those flexible queries. 

You can supply the subgraph you wish to query in the request for the composable and the composable returns the familiar `useQuery` response from `vue-query`

One thing you may note is that the `query` parameter for the request is a function which returns an object, this is due to reactivity constraints. If you pass in a query as a plain object (not returned by a function) E.g. (Also usage)

```
type PoolShares = {
  poolShares: {
    pool: {
      id: string;
    },
    balance: string;
  }
}

const customQuery = useSubgraphQuery2<PoolShares>(subgraphs.balancer, ['custom', 'pools'], () => `
  {
    poolShares(where: { userAddress: ${account.value} }) {
      pool {
        id
      }
      balance
    }
  }
`);
```

In the above query (assuming `account` is passed in the query key), if the query is refetched due to the `account` property (which is reactive) changing the query will refetch with the old `account` value. This is because the query object
is passed in at time of function call and isn't reactive. You may be thinking, why not make the `query` object be passed in as a reactive or computed property? I've tried this way, and there was too much extra code for that. So a function seems to be best remedy for that reactivity problem.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

N/A. Usages will be visible in coming PRs

## Visual context

Code only change. N/A

## Checklist:


garethfuller:
@1saf I'll wait to see an example of this composable in action. My only comment so far is that the usage seems a little overcomplicated. I made an attempt at a simpler API here but I'm not 100% sure if it will work properly in the wild: https://github.com/balancer-labs/frontend-v2/pull/1463 (It does work in terms of returning data).

garethfuller (approved):
Looks good, my only comment would be that I think this file should be in the queries folder, I don't think it needs its own. It's just a more generic query than the others.

1saf:
> 

done deal

PR #1462, opened 2022-03-10
Fix/confirming issue

talentdev2020:
# Description
  While approving, it still shows the "confirming".
This is because the error is happening when get the block using getBlock(). Need to investigate more.
For now, I fixed this issue manually.


garethfuller:
This looks like a mistake, the diff is huge, closing for now.

PR #1463, opened 2022-03-10
Subgraph helper proposal

garethfuller:
# Description

Proposal for a simpler approach. This composable could be used like this:
```ts
type PoolShares = {
  poolShares: {
    pool: {
      id: string;
    },
    balance: string;
  }
}

const query = computed(() => `
  {
    poolShares(where: { userAddress: ${account.value} }) {
      pool {
        id
      }
      balance
    }
  }
`);

const customQuery = useSubgraphQuery2<PoolShares>(subgraphs.balancer, query, ['custom', 'pools']);
```

Note: I'm not 100% sure this solves the problems with reactivity you might have experienced since I didn't have time to test extensively.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:


garethfuller:
@1saf I'll wait to see an example of this composable in action. My only comment so far is that the usage seems a little overcomplicated. I made an attempt at a simpler API here but I'm not 100% sure if it will work properly in the wild: https://github.com/balancer-labs/frontend-v2/pull/1463 (It does work in terms of returning data).

1saf:
How we pass in the query is mainly irrelevant, I went with the JSON format as that was something I thought we were already familiar with, so Im easy on whichever direction we go with in regards to that. 

About the computed property and reactivity however. I wanted to avoid the use of a separate computed property as I wanted to keep all the information regarding the query in one single place, not seperated out into different computed variables (even if, for the sake of reactivity) because with larger components, or if we need to fetch more data - we end up with a nightmare soup of differing computed query variables to manage, which are separated away from their usages.
It seems minor but once it adds up: it is a nightmare to work with

src/composables/subgraph/useSubgraphQuery2.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const subgraphs = {
+  gauge: configService.network.subgraphs.gauge,
+  balancer: configService.network.subgraph
+};
+
+export default function useSubgraphQuery2<T>(
+  subgraphUrl: string,
+  query: Ref<string>,
+  key: string[],
```
1saf, line 14: i think the key should take precedent in the list of params before the query

src/composables/subgraph/useSubgraphQuery2.ts
```diff
@@ -0,0 +1,39 @@
...
+import axios from 'axios';
+import { configService } from '@/services/config/config.service';
+
+export const subgraphs = {
+  gauge: configService.network.subgraphs.gauge,
+  balancer: configService.network.subgraph
+};
+
+export default function useSubgraphQuery2<T>(
+  subgraphUrl: string,
```
1saf, line 12: yeah I think having this as an actual URL is a pretty good idea, given the fact that we might need to query subgraphs apart from above for some use cases

src/composables/subgraph/useSubgraphQuery2.ts
1saf: this is because it isnt an object anymore

PR #1466, opened 2022-03-11
useTokenApprovals/Actions to support a custom spender

1saf:
# Description

Currently our token approvals architecture only takes into account two main spenders (The vault, and wstETH [Here](https://github.com/balancer-labs/frontend-v2/blob/develop/src/providers/tokens.provider.ts#L136-L139) ).
How it works is that it checks the allowances for all tokens in the store for approvals against those contracts. 
Adding new spenders is fine to the list of spenders above is fine - if we know those spenders are pretty static.

However, for the upcoming tokenomics pull requests we need to have the ability to approve against many different spenders as the spender is the gauge contract - which is dynamic based on the pool. We could just keep a static list and have it imported but I feel that we should have support for custom or dynamic spenders.

This pull request adds support for checking token approvals/allowances against a custom spender via the `useTokenApprovals` and `useTokenApprovalActions` composables. 
Instead of modifying the entire allowance architecture from the ground up (from the tokens provider), I've simplified it down to just modifying those above composables. All existing behaviour is maintained, just that we have additional parameters and functions to support the custom spender usecase.

Usage
```
const { getTokenApprovalActionsForSpender } = useTokenApprovalActions(
  [props.pool.address],
  ref([balanceFor(props.pool.address).toString()])
);

const actions = ref<TransactionActionInfo[]>([]);

 const approvalActions = await getTokenApprovalActionsForSpender('mySpenderAddress');
 actions.value.unshift(...approvalActions);

<BalActionSteps
  :actions="actions"
/>
```

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [X] Other

## How should this be tested?

You will be able to test this with the tokenomics pull request. But make sure that any existing approval flows for pool creation, trading and investing are not broken.

## Visual context

N/A

## Checklist:

### DIFF
--- a/src/composables/useTokenApprovalActions.ts
+++ b/src/composables/useTokenApprovalActions.ts
@@ -1,6 +1,8 @@
 import { Ref } from 'vue';
 import { useI18n } from 'vue-i18n';
-import useTokenApprovals from '@/composables/pools/useTokenApprovals';
+import useTokenApprovals, {
+  ApprovalState
+} from '@/composables/pools/useTokenApprovals';
 import useTokens from '@/composables/useTokens';
 import { TransactionActionInfo } from '@/types/transactions';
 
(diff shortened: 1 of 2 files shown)


src/composables/useTokenApprovalActions.ts
```diff
@@ -10,27 +12,52 @@ export default function useTokenApprovalActions(
...
+      spender
+    );
+    const actions = getTokenApprovalActions(
+      requiredApprovalStateForSpender,
+      spender
+    );
+    return actions;
+  }
+
+  function getTokenApprovalActions(
```
1saf, line 37: moved into a common function

src/composables/useTokenApprovalActions.ts
```diff
@@ -10,27 +12,52 @@ export default function useTokenApprovalActions(
...
-  ).map(address => {
-    const token = getToken(address);
-    return {
-      label: t('transactionSummary.approveForInvesting', [token.symbol]),
-      loadingLabel: t('investment.preview.loadingLabel.approval'),
-      confirmingLabel: t('confirming'),
-      stepTooltip: t('investment.preview.tooltips.approval', [token.symbol]),
-      action: () => {
-        return approveToken(token.address);
+  const tokenApprovalActions: TransactionActionInfo[] = getTokenApprovalActions();
```
1saf, line 21: existing behaviour is maintained, just from the common function now

garethfuller (approved):
I tested approvals elsewhere and they don't seem to be affected so I think it's all good. Just a minor cleanup comment

PR #1513, opened 2022-03-21
Refactor BalLink

garethfuller:
# Description

The current implementation wasn't very flexible. This PR allows for the BalLink component to also be a router-link and also disables the 'correct' way.

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test BalLinks throughout app
- [ ] Add a BalLink to a page and test props like disabled and external

## Visual context

n/a

## Checklist:

### DIFF
--- a/src/components/_global/BalLink/BalLink.vue
+++ b/src/components/_global/BalLink/BalLink.vue
@@ -1,56 +1,72 @@
-<template>
-  <a v-if="!disabled" :class="[classes]" v-bind="attrs_">
-    <slot />
-  </a>
-  <div v-if="disabled" :class="[classes]" v-bind="attrs_">
-    <slot />
-  </div>
-</template>
+<script lang="ts">
+export default {
+  inheritAttrs: false
+};
+</script>
 
-<script>
-import { defineComponent, computed } from 'vue';
+<script lang="ts" setup>
+import { computed, useAttrs } from 'vue';
 
-export default defineComponent({
-  name: 'BalLink',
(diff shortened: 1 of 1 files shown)


src/components/_global/BalLink/BalLink.vue
```diff
@@ -1,56 +1,72 @@
...
+/**
+ * TYPES
+ */
+type LinkTag = 'a' | 'router-link';
+
+type Props = {
+  tag: LinkTag;
+  external: boolean;
+  disabled: boolean;
+  noStyle: boolean;
```
evgenyboxer, line 19: minor - shouldn't most of these props be optional? (I know they got a default, but was just wondering)

evgenyboxer (approved):

src/components/_global/BalLink/BalLink.vue
garethfuller: By having defaults they are essentially optional. The reason I always set boolean prop defaults to false is:
1. So that the prop value is always a boolean. I'm not 100% what the type will be if I just make it optional but don't pass it, maybe null? I just like being sure.
2. I like to force boolean props to be false so that to use them you just need to pass the prop name in the tag, like:
```vue
<BalLink external noStyle>A link</BalLink>
```

garethfuller: It's probably a preference thing really

Issue #1527, opened 2022-03-23
Input currency rounds up when output currency has fewer decimals causing insufficient balance when max button is pressed


mikebmikeb:
**Describe the bug**
When the input token has more decimals than the output token, the input token gets rounded up instead of the output token getting rounded down which makes the "max" button nonfunctional since it rounds up--giving you an insufficient balance error. 

**To Reproduce**
Go to the trading screen for Arbitrum and set the input as NDX and output as USDC. 
Set 100.366613 NDX tokens as the input
100.366613 will get rounded up to 100.367 to account for the fewer decimals for the USDC token.
If you only had 100.366613 ndx in your account and hit the max button, because its rounding up, you would be unable to trade due to insufficient balance.

PR #1577, opened 2022-03-28
Don't format numbers the user entered into trade

timjrobinson:
# Description

Fixes: #1527 

The `updateTradeAmounts` function is called whenever a new block is received. It re-calculates the swap value and updates the numbers in the trade view. It is currently calling `formatAmount` on these values before re-writing them to the page. This is causing issues because it re-formats the number the user just entered to only have 6 significant digits. 

This fix makes it only reformat + display the value that the user did not enter. 

@evgenyboxer This also means `formatFixed` with the deltas is not run on the number the user entered, does that seem fine to you? 

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

Refer to issue #1527 and attempt to replicate

- Load the Trade UI
- Perform a trade between any two assets. 
- Specify a large number or a number with a large number of decimals 
- Ensure that the number you are entering is not rounded or changed at all. 

## Checklist:

### DIFF
--- a/src/composables/trade/useSor.ts
+++ b/src/composables/trade/useSor.ts
@@ -191,8 +191,12 @@ export default function useSor({
     if (sorReturn.value.hasSwaps) {
       const { result } = sorReturn.value;
 
+      const swapType: SwapType = exactIn.value
+        ? SwapType.SwapExactIn
+        : SwapType.SwapExactOut;
+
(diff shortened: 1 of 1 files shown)


evgenyboxer (approved):
Thanks for this - yeah this makes sense.

PR #1603, opened 2022-03-29
Fix relayApproval and minter errors on Polygon

timjrobinson:
# Description

## Heisenbug on Polygon

I've come across a heisenbug on Polygon - in lido.ts the `isStETH` function is returning true for all trades, even though the value of `stEthAddress` is an empty string, so it should return false from the first if statement. When running development using Polygon chain the bug doesn't happen, it returns false from `isStETH` as expected. It may be something to do with source code compilation and `stEthAddress` being brought in from a closure instead of passed into the function. 

This bug causes the lidoRelayerApproval to trigger which calls useRelayerApproval with a relayer of an empty string (because the lido relayer is not set on Polygon). Which is causing this error: https://sentry.io/organizations/balancer-labs/issues/2862170362/?project=5725878&referrer=weekly-email because `vaultContract.value.hasApprovedRelayer` is attempting to check if the vault contract has an approved relayer of empty string. 

I've added a check in the relayer to make sure it has a value, and a log in isStETH to diagnose further what's on in production. 

Here's a loom of the weirdness:
https://www.loom.com/share/fc65118fb2ba49fbac85a0cdb83d8d20

## Balancer Minter

I removed the console.error because there is no minter on Polygon/Arbitrum so it always triggers upon loading the page. 

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- Load Polygon
- Add a breakpoint in `isStEth` in `lido.ts`
- Start a new trade.
- The breakpoint should be hit, see if it progresses past the if statement at the beginning. 

## Visual context

## Checklist:

### DIFF
--- a/src/composables/queries/useRelayerApprovalQuery.ts
+++ b/src/composables/queries/useRelayerApprovalQuery.ts
@@ -48,6 +48,10 @@ export default function useRelayerApprovalQuery(
   );
 
   const queryFn = async (): Promise<boolean> => {
+    if (!relayer.value) {
+      return true;
+    }
+
     const approved = await vaultContract.value.hasApprovedRelayer(
       account.value,
       relayer.value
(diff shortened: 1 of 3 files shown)


garethfuller (approved):
Yeah, this is a bizarre bug.

evgenyboxer (approved):

PR #1620, opened 2022-03-31
Add preview of next period votes

markusbkoch:
# Description

Until gauge type weights are set on Tuesday, queries to the gauge controller for gauges' relative weights return zero, so we can't show the current status of the voting process in the vebal page.

But since we only have mainnet gauges for now, we can query the weight of each gauge (`get_gauge_weight`) and the sum of weights of all type 2 gauges (`get_weights_sum_per_type(2)`) and compute the relative value from that. 

This is only a temporary measure that allows us to display next period votes before Tuesday. These changes should be reverted after the type weights are set. In the meantime, anyone please feel free to make any necessary improvements to the code (legibility,  performance w/e)

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

vebal page should display next period votes

## Visual context

image

## Checklist:

### DIFF
--- a/src/services/balancer/gauges/gauge-controller.decorator.ts
+++ b/src/services/balancer/gauges/gauge-controller.decorator.ts
@@ -16,6 +16,8 @@ export interface UserVotesData {
 
 export interface RawVotesData {
   votes: number;
+  absVotes: number;
+  totalVotes: number;
   votesNextPeriod: number;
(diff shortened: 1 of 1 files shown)


timjrobinson:
I've combined this code with the PR I was working on that gets votes for this week and last in #1625.

PR #1625, opened 2022-04-01
Hotfix: Calculate gauge weights manually

timjrobinson:
# Description

Because the total_weight value is not yet set gauge voting weights cannot be calculated using it. @markusbkoch realized we can use points_sum to calculate relative gauge weights in #1620. This PR improves upon that using points_sum and points_weight to calculate the relative weights of each gauge for this week and next, which should be correct for both this first week and future weeks. 

In Markus's PR he used a type_weight of 0.56 however when adding up the percentages this way they don't add up to a full 100%, so I'm not sure if that is correct or maybe all percents on mainnet should only add up to 56% total? The weight for each chain is expected to change so they all add up to 100% so I've added this value to the config file for each. 

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- Visit the gauges page on mainnet Ethereum
- Ensure gauge weights look correct and add up to 100%

## Visual context

2022-04-01-133031_1590x1082_scrot

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer-labs/frontend-v2",
-  "version": "1.46.11",
+  "version": "1.46.12",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@balancer-labs/frontend-v2",
-      "version": "1.46.11",
+      "version": "1.46.12",
       "license": "MIT",
       "devDependencies": {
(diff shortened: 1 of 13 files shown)


garethfuller (approved):
Looks good to me, but I'm not clear on the math. I'd like Tom or Markus to approve before merging.

markusbkoch (approved):

PR #1683, opened 2022-04-08
Improve initial loading timing

kyzia551:
# Description

While working with the app I noticed that initial loading takes quite a lot of time and many calls are going sequentially.
After some research, I found that most parts of the time were consumed by ```balancerSubgraphService.pools.decorate()``` method.

As the result on local, it dropped from ~7s to ~2s

## Type of change

- [x] migration from JsonRpcProvider to BatchJsonRpcProvider (helped do drop number of RPC calls from ~46 to ~21)
- [x] parallel requests


garethfuller:
Thanks so much, @kyzia551 🙌  This is super helpful! I didn't know about the `JsonRpcBatchProvider` 🙈 

Closing in favour of #1684 so I could do some final linting before merging.

PR #1705, opened 2022-04-12
Add sidebar closing by click empty space on the right

alter-eggo:
# Description

This PR adds close sidebar by click empty space on the right. It has better ux for the users on mobile devices.
Screenshot 2022-04-12 at 22 26 27
## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [x] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:

### DIFF
--- a/src/composables/useSidebar.ts
+++ b/src/composables/useSidebar.ts
@@ -3,18 +3,18 @@ import { ref } from 'vue';
 /**
  * STATE
  */
-const sidebarOpen = ref(false);
+const isSidebarOpen = ref(false);
 
 /**
  * METHODS
  */
-function setSidebarOpen(newVal: boolean) {
-  sidebarOpen.value = newVal;
+function setSidebarState(newVal: boolean): void {
+  isSidebarOpen.value = newVal;
 }
 
 export function useSidebar() {
   return {
-    sidebarOpen,
-    setSidebarOpen
+    isSidebarOpen,
+    setSidebarState
   };
 }
--- a/src/composables/useSidebar.ts
+++ b/src/composables/useSidebar.ts
@@ -3,18 +3,18 @@ import { ref } from 'vue';
 /**
  * STATE
  */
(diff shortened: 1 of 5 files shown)


src/composables/useSidebar.ts
```diff
@@ -3,18 +3,18 @@ import { ref } from 'vue';
...
+function setSidebarState(newVal: boolean): void {
+  isSidebarOpen.value = newVal;
 }
 
 export function useSidebar() {
   return {
-    sidebarOpen,
-    setSidebarOpen
+    isSidebarOpen,
+    setSidebarState
```
garethfuller, line 18: I'm not sure why this variable and function name needed to change. Two reasons:
1. Sometimes adding `is` before variable name doesn't make as much sense. In this case `v-if="sidebarOpen"` (if sidebar open?) makes more sense if you read it out in English than `v-if="isSidebarOpen"` (if is sidebar open?).
2. The `setSidebarOpen` function was specifically about mutating the `sidebarOpen` variable. If we were to have more state variables here then maybe it would make sense to have a general state mutation function. But I don't think there is any need to change the function name here since it makes the meaning more obscure.

garethfuller (changes requested):
This is great! Thanks for contributing. My only comment is that I don't think the function and variable name should have changed.

src/composables/useSidebar.ts
alter-eggo: 1. I see. I made such changes, because I read it as `"Is sidebar open?"`. Now I see other logics.
2. Yeah, `state` is not good here. 
I have return both names.

garethfuller: Thank you! Yeah appending `is` to the variable name could really be either. It's a close call.

garethfuller (approved):
Again thanks for your help here! 🙌

PR #1728, opened 2022-04-19
Hotfix - VeBAL Helpers

1saf:
# Description

To fetch APRs correctly for gauges we need to be calling the checkpoint function on the gauge. Nico has kindly given us a helper contract that will do that in the background with a read only call instead of needing to perform a tx.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

APRs should still show as expected.

## Visual context

N/A

## Checklist:


timjrobinson (approved):
Tested and LGTM

src/services/staking/staking-rewards.service.ts
```diff
@@ -66,15 +71,18 @@ export class StakingRewardsService {
...
-      gaugeAddresses,
-      timestamp
-    );
+  private async getRelativeWeightsForGauges(gaugeAddresses: string[]) {
+    const timestamp = getUnixTime(new Date());
+    // the ve bal helpers contract for gauge weights calls
+    // the checkpoint function which is necesary for returning
+    // the correct value.
+    let result = await this.veBALHelpers.getRelativeWeights(gaugeAddresses);
+    if (configService.network.chainId === Network.KOVAN) {
```
timjrobinson, line 80: Rather than re-using result and doing two calls on Kovan you could move this if statement up and put a return inside it.

1saf: done

PR #1730, opened 2022-04-19
Main page loading

1saf:
# Description

Improves the loading of the main page!

## Type of change

- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## How should this be tested?

Load up the main investments table on the home page, everything should load much faster but all previous behaviour should be retained.

## Visual context

Speedy load.

## Checklist:


src/pages/index.vue
```diff
@@ -92,10 +103,73 @@ watch(showMigrationColumn, () => console.log(showMigrationColumn.value));
...
+//       console.log('dingle', pools.value);
+//       return new Promise(resolve => {
+//         setTimeout(() => {
+//           pools.value = pools.value += 5;
+//           resolve('fetched');
+//         }, 7000);
+//       });
+//     }
+//   }
+// });
```
garethfuller, line 141: Delete

src/pages/index.vue
```diff
@@ -92,10 +103,73 @@ watch(showMigrationColumn, () => console.log(showMigrationColumn.value));
...
+    {{ { dynamicDataLoading, priceQueryLoading } }}
+    {{ dataStates }}
+    {{ Object.keys(prices).length }}
+    <!-- <AnimatePresence
+      :isVisible="
+        dataStates['basic'] === 'success' &&
+          dataStates['liquidity'] === 'success'
+      "
+    > -->
+    bing {{ isLoadingMore }}
```
garethfuller, line 155: Remove

src/services/rpc-provider/rpc-provider.service.ts
```diff
@@ -40,6 +44,34 @@ export default class RpcProviderService {
...
+        return 2;
+      case Network.ARBITRUM:
+        return 3;
+      case Network.KOVAN:
+        // Should be ~4s but this causes subgraph to return with unindexed block error.
+        return 1;
+      default:
+        return 13;
+    }
+  }
```
garethfuller, line 74: I get why you've put this in here, but the rpc service should be exclusively about dealing with rpcs. This logic should either go in the web3 service, a new block service or maybe just a composable.

garethfuller (changes requested):
Just submitting a first pass look at the code, I haven't tested it yet. Big blocker atm is generating tokenlists on build and putting them in the appropriate place.

Also, please clean up linting issues.

garethfuller:
@1saf It doesn't seem to be caching queries like we were before. So if you load the home page list then click into a pool, then click back to the home page, it has to reload everything I think. Can we not cache like we were before? Not a huge problem if not.

garethfuller:
It also breaks the pagination if you go into a pool page and then come back. As in, if you had loaded a bunch of pages, click into a pool and then come back, it just resets to loading the first page. Again, not a deal-breaker but it would be nice if it worked as it currently does and maintains that page state.

garethfuller (approved):

PR #1758, opened 2022-04-21
Hotfix: Fetch gauge weights in a more accurate way

timjrobinson:
# Description

Change fetching of gauge weights to use the function `gauge_relative_weight_write` which does a checkpoint on the gauge first to ensure it's weight is accurate for the week. Without the checkpoint weights aren't updated until a user votes on that specific gauge. 

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

View the veBAL gauges page, it should be similar to production but with many more of the gauge weights filled in instead of being 0. 

## Visual context

Production on left, local testing against mainnet on right

2022-04-21-230727_2363x1858_scrot

## Checklist:

### DIFF
--- a/src/lib/abi/GaugeController.json
+++ b/src/lib/abi/GaugeController.json
@@ -298,23 +298,6 @@
       {
         "name": "addr",
         "type": "address"
-      }
-    ],
-    "outputs": [
-      {
-        "name": "",
-        "type": "uint256"
-      }
-    ]
-  },
-  {
-    "stateMutability": "nonpayable",
-    "type": "function",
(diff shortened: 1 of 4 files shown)


src/lib/abi/GaugeController.json
```diff
@@ -290,23 +290,6 @@
       }
     ]
   },
-  {
```
timjrobinson, line 293: This needs to be removed because Ether's cant deal with duplicate functions in ABI's

garethfuller (approved):

PR #1779, opened 2022-04-26
Hotfix - Investments banner on layer 2's

1saf:
# Description

Investments banner was not loading value on layer 2

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

Total investments banner should load on layer 2's

## Visual context

N/A

## Checklist:

### DIFF
--- a/src/composables/queries/useVeBalLockInfoQuery.ts
+++ b/src/composables/queries/useVeBalLockInfoQuery.ts
@@ -2,7 +2,6 @@ import { UseQueryOptions } from 'react-query/types';
 import { computed, reactive } from 'vue';
 import { useQuery } from 'vue-query';
 
-import QUERY_KEYS from '@/constants/queryKeys';
 import { balancerContractsService } from '@/services/balancer/contracts/balancer-contracts.service';
 import { VeBalLockInfo } from '@/services/balancer/contracts/contracts/veBAL';
 import useWeb3 from '@/services/web3/useWeb3';
@@ -29,11 +28,6 @@ export default function useVeBalQuery(
    */
   const enabled = computed(() => isWalletReady.value && isVeBalSupported.value);
 
-  /**
-   * QUERY INPUTS
-   */
-  const queryKey = reactive(QUERY_KEYS.Tokens.VeBAL(networkId, account));
-
   const queryFn = () =>
     balancerContractsService.veBAL.getLockInfo(account.value);
 
@@ -42,5 +36,9 @@ export default function useVeBalQuery(
     ...options
   });
 
-  return useQuery<QueryResponse>(queryKey, queryFn, queryOptions);
+  return useQuery<QueryResponse>(
+    reactive(['tokens', 'veBAL', { networkId, account }]),
+    queryFn,
+    queryOptions
+  );
 }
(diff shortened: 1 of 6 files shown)


src/composables/queries/useVeBalLockInfoQuery.ts
```diff
@@ -29,11 +28,6 @@ export default function useVeBalQuery(
    */
   const enabled = computed(() => isWalletReady.value && isVeBalSupported.value);
 
-  /**
-   * QUERY INPUTS
-   */
-  const queryKey = reactive(QUERY_KEYS.Tokens.VeBAL(networkId, account));
-
```
1saf, line 36: These query keys are not actually reactive, since it is a function they lose reactivity åfter the initial return. I do not recommend this methodology anymore

src/composables/useVeBAL.ts
```diff
@@ -18,7 +23,7 @@ const showRedirectModal = ref(false);
  * COMPUTED
  */
 export const isVeBalSupported = computed(
-  () => isMainnet.value || isKovan.value
+  () => isMainnet.value || isKovan.value || isArbitrum.value || isPolygon.value
```
1saf, line 26: I've added the other networks instead of just a blanket true in case we need to fine tune it later on

src/composables/queries/useVeBalLockInfoQuery.ts
timjrobinson: True, makes sense.

src/composables/useVeBAL.ts
timjrobinson: I thought users can't vote on gauges or lock veBAL on L2? This value is used for those pages.

1saf: yeah damn, I'll see another way

src/composables/queries/useVeBalLockInfoQuery.ts
evgenyboxer: What if the `reactive` will be inside the return of the function? That should solve it no?

1saf: Ill test that out

src/composables/useVeBAL.ts
1saf: @timjrobinson fixed :)

src/composables/queries/useVeBalLockInfoQuery.ts
1saf: @evgenyboxer nah doesn't work

src/composables/useLock.ts
```diff
@@ -21,11 +20,7 @@ export function useLock() {
   /**
    * QUERIES
    */
-  const shouldFetchLockPool = computed((): boolean => !isL2.value);
-  const lockPoolQuery = usePoolQuery(
-    lockablePoolId.value as string,
-    shouldFetchLockPool
-  );
+  const lockPoolQuery = usePoolQuery(lockablePoolId.value as string);
```
timjrobinson, line 23: Shouldn't this still be disabled on L2? As the lockable pool is only on Mainnet

timjrobinson (approved):

PR #1791, opened 2022-04-27
Update L2 incentives alert copy

markusbkoch:
Now that staking is live, make the alert more informative


pkattera:
@markusbkoch Regarding the suggestions copy changes, I'm happy with it.

pkattera (approved):
Only reviewed the copy changes (not the difference in package-lock.json)

Issue #1812, opened 2022-04-28
Values get messed up when swapping input/output


mikebmikeb:
**Describe the bug**
When using the "swap input/output" button on the trade ui 
<img width="101" alt="image" src="https://user-images.githubusercontent.com/90106964/165842536-f00a8f8f-1086-43c8-8c29-5544fd358273.png">
the values get buggy when the above button is clicked while there are inputs filled in.

**To Reproduce**
1. go to the trade interface on polygon
2. put input as usdc, output as link
3. input 20 link as output value
4. click the "swap input/output" button

**Expected behavior**
the amounts should not go into the 

**Screenshots**
Prior to clicking the button 
<img width="520" alt="image" src="https://user-images.githubusercontent.com/90106964/165843974-e43650c9-e2f7-411a-939c-3187834bb4c8.png">

clicking the button twice
<img width="459" alt="image" src="https://user-images.githubusercontent.com/90106964/165844048-a712ddb5-ec99-4c3f-b653-95ed00a2dd68.png">

clicking it a third time 
<img width="487" alt="image" src="https://user-images.githubusercontent.com/90106964/165844083-b5dd4953-1975-41d6-a08b-0eab13047a00.png">

**Additional context**
Seems to not be 100% consistent, but if you click the swap input/output button a couple times it will definitely trigger it.

I theorize this has something to do with usdc's 6 decimal places since i can't get similar behavior from other tokens

ThannDev:
Talk to Polygon Support on the [Live support page](https://forummaticnetwork.org/) so that your issue can be looked into.

timjrobinson:
Fixed in #1819

Issue #1811, opened 2022-04-28
Output value disappearing when output gets too large on polygon trade ui


mikebmikeb:
**Describe the bug**
A bit hard for me to explain this in words, video below.  the output value disappears under certain conditions, perhaps when it has to route through multiple pools?  (Can't see routing prior to trading anymore) 
This does not happen if you manually type the output.  

**To Reproduce**
1. Go to polygon trade interface
2. input: matic output: usdc
3. type in 2000 matic as input, the trade will route properly.  input 20000, the trade will route correctly for a brief moment, then the output will disappear

**Screenshots**
video:
https://user-images.githubusercontent.com/90106964/165843233-a116d28d-cda1-48f1-9b5d-10c3dff3f34f.mov

with input as 2000
<img width="490" alt="image" src="https://user-images.githubusercontent.com/90106964/165843672-897856de-8aba-4ed4-b024-809a16f99e37.png">

with input as 20000
<img width="502" alt="image" src="https://user-images.githubusercontent.com/90106964/165843708-1237871c-1ca3-4f71-9814-6d31dee129da.png">

**Expected behavior**
The output value should not disappear.

timjrobinson:
Fixed in 1819

PR #1819, opened 2022-05-03
Hotfix: Fix trade value snapping

timjrobinson:
# Description

On the trading screen when changing the input value sometimes the output was snapping to either 0 or the return value for the previous inputted amount. See: https://www.loom.com/share/639dc0d752d44fc7b65304704e41a2f2 /cc @zekraken-bot 

This is a race condition between:
- A new block comes in, which calls https://github.com/balancer-labs/frontend-v2/blob/develop/src/composables/trade/useSor.ts#L196
- While that async call is processing the user changes the value. This causes... 
- SOR recalculates the swap amount here: https://github.com/balancer-labs/frontend-v2/blob/develop/src/composables/trade/useSor.ts#L335 
- This sets the output value to the new amount. 
- Then the async call finishes and sets the value to the old sor return here: https://github.com/balancer-labs/frontend-v2/blob/develop/src/composables/trade/useSor.ts#L215

So I added a check for this race condition and it aborts if it happens. 

The value becoming 0 is happening because SOR is returning the address list (of all tokens involved in the swap) sorted, and when that list is passed into `batchQuerySwap` the first and last items are no longer the input/output tokens. So they correct indexes of those tokens must be found instead. This is also fixed. 

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- Open trade UI on Polygon
- Set two tokens
- Continually change the input value and watch the output value change, ensure it isn't snapping between any values or snapping back to a previously calculated output value. 

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer-labs/frontend-v2",
-  "version": "1.51.6",
+  "version": "1.51.7",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@balancer-labs/frontend-v2",
-      "version": "1.51.6",
+      "version": "1.51.7",
       "license": "MIT",
       "devDependencies": {
         "@aave/protocol-js": "^4.3.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@balancer-labs/frontend-v2",
-  "version": "1.51.6",
+  "version": "1.51.7",
   "engines": {
     "node": "14.x",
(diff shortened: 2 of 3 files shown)


timjrobinson:
Also discovered this fixes #1812

timjrobinson:
Added a patch for the results becoming 0 bug. It was due to the addresses returned from SOR being sorted alphabetically and so when they were passed into queryBatchSwap the values were in the wrong order. 

Fixes: #1811

johngrantuk (approved):

garethfuller (approved):

PR #1889, opened 2022-05-16
Hotfix - fix steth trades

garethfuller:
# Description

Currently if you were trading from or too stETH, after inputting amounts when the next block was mined the stETH amount would reset to zero. This is because for the initial amounts we're using the SOR and then to update amounts on each block we're using the queryBatchSwap, since it's a more efficient call. However, the return from the query batch swap only includes a reference to wstETH in the path addresses and so it's impossible to map the tokenOut address which is stETH to the query batch swap amount out index.

This PR adds a crude hack that checks if the token in/out is stETH and in this specific function for updating amounts on each block, replaces it with wstETH for finding the mapped amount values from the query batch swap results.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test swaps to or from stETH. Input amounts and then wait for about a minute for new blocks to come in. The amount for stETH should not reset to 0.

## Visual context

n/a

## Checklist:

### DIFF
--- a/src/composables/trade/useSor.ts
+++ b/src/composables/trade/useSor.ts
@@ -222,6 +222,10 @@ export default function useSor({
             ? AddressZero
             : tokenOutAddressInput.value;
 
+        // If the token in/out is stETH then finding the token position
+        // below doesn't work because result.tokenAddresses only includes
+        // wstETH. This is a crude hack to replace token in/out address
+        // with wstETH so the index mapping works.
         if (isStEthAddress(tokenInAddressInput.value))
           tokenInAddress = configService.network.addresses.wstETH;
         if (isStEthAddress(tokenOutAddressInput.value))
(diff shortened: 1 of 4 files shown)


src/composables/trade/useSor.ts
```diff
@@ -248,17 +257,15 @@ export default function useSor({
...
         }
 
         if (swapType === SwapType.SwapExactIn) {
-          tokenOutAmountInput.value =
-            tokenOutAmountNormalised.toNumber() > 0
-              ? formatAmount(tokenOutAmountNormalised.toString())
-              : '';
+          tokenOutAmountInput.value = tokenOutAmountNormalised.gt(0)
+            ? formatAmount(tokenOutAmountNormalised.toString())
+            : '';
```
garethfuller, line 268: This is just a formatting change, unrelated to bug.

Issue #1830, opened 2022-05-05
APR popup on staked pools table is slightly wrong


timjrobinson:
**Describe the bug**

When hovering over the APR of a pool in the staked pools table it is showing the wrong value for staking APR. It is showing the full amount as staking APR. See screenshots for more information. 

**To Reproduce**

View account: 0x94fbC7cecB1D7C9B6070077268Dc8239A10f3b8e

Look at `Staking APR` in the popup in the staking table, it differs from the APR in the main pools table. 

**Screenshots**

Correct values shown in main pools table. I don't have any veBAL so it is the minimum:
![2022-05-05 18_37_54-Balancer](https://user-images.githubusercontent.com/525534/166889322-11d2eeae-0c17-480f-85d9-0ab45b06e3b7.png)

Staking table shows staking APR of 8.27% in popup when it should be 7.05%. Then that plus the 1.22% swap fees APR will add up to the correct value:
![2022-05-05 18_37_46-Balancer](https://user-images.githubusercontent.com/525534/166889298-ab1620ce-0adc-4bbd-8fb5-ec2ab7664053.png)

PR #1897, opened 2022-05-17
Refactor APR decoration and components

garethfuller:
# Description

Refactors pool decoration and in particular APR calculations to be more isolated. I've also refactored the APR display components.
- Had to remove all references to 'Potential weekly yield' due to the fact that we only have boost values when a user is already staked. More work needs to be done to calculate potential boost numbers assuming fully staking your BPT.
- Simpified Pool types by removing the `DecoratedPool` type and instead adding optional properties to the `Pool` type, e.g. `volumeSnapshot` `feesSnapshot` and `apr` rather than being nested in the `dynamic` property.

Note: Branch name is redundant. I was originally going to include veBAL APRs but a major refactor was required first.

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Check all APR tooltips to make sure they make sense and include the right numbers.
- [ ] Check investment and staking flows to make sure no errors after removal of weekly yield stuff.

## Visual context

n/a

## Checklist:


src/services/pool/concerns/apr.concern.ts
```diff
@@ -0,0 +1,161 @@
...
+    const yieldAPR = await this.calcYieldAPR(
+      prices,
+      currency,
+      protocolFeePercentage
+    );
+    const baseTotalAPR = bnSum([
+      swapFeeAPR,
+      yieldAPR.total,
+      stakingRewardApr
+    ]).toString();
```
garethfuller, line 41: Note to self, baseTotalAPR probably shouldn't include staking rewards since we will need to reference totals with and without staking. Adjust the schema to make these two scenarios easier to handle.

garethfuller:
The Lint and test failures are fake news. They don't seem to be running using the latest commit code for some reason. Looks like a Github bug.

src/services/pool/concerns/apr.concern.ts
garethfuller: This has been handled

src/composables/useNumbers.ts
```diff
@@ -82,101 +83,105 @@ export function fNum(
   );
 }
 
-export default function useNumbers() {
-  const { currency } = useUserSettings();
-  const { priceFor } = useTokens();
+export function numF(
+  number: number | string,
+  options: FNumOptions | undefined = {},
+  currency: FiatCurrency = FiatCurrency.usd
```
timjrobinson, line 89: Nice, I like moving it out of the composable with the currency as an option.

src/composables/useNumbers.ts
```diff
@@ -82,101 +83,105 @@ export function fNum(
...
+
+  function fNum2(
+    number: number | string,
+    options: FNumOptions | undefined = {}
+  ): string {
+    const _currency = currency?.value || FiatCurrency.usd;
+    return numF(number, options, _currency);
   }
 
   return { fNum, fNum2, toFiat };
```
timjrobinson, line 187: I think we can get rid of fNum now.

src/services/pool/concerns/apr.concern.ts
```diff
@@ -0,0 +1,178 @@
...
+        unstaked: unstakedTotalAPR,
+        staked: {
+          calc: aprGivenBoost,
+          ...stakedAprRange
+        }
+      },
+      swap: swapFeeAPR,
+      yield: yieldAPR,
+      staking: {
+        bal: stakingBalApr,
```
timjrobinson, line 58: Is this only for the veBAL pool?

timjrobinson (approved):
Looks great, nice work! I like the new APR layout more than before. 

I tested it out on Polygon and Mainnet and it all appears to work.

timjrobinson:
Fixes #1830

src/composables/useNumbers.ts
garethfuller: Yeah, that composable has got a bit messy, but I needed to do this to make the number formatted function pure.

src/services/pool/concerns/apr.concern.ts
garethfuller: No, this is the BAL emissions APR for any pool that receives them.

PR #1899, opened 2022-05-17
UX improvements including improved gasless trading experience

pkattera:
# Description

1. Improved gasless trading toggle experience
2. Prevent background scrolling when the modal is present
3. Better hover and focus states for BalInput component

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other: ui/ux improvements

## How should this be tested?

**1. Improved gasless trading toggle experience**

The toggle currently is incorrect set to 'On' while disabled for the default starting trading pair of ETH -> DAI. This is confusing because it should be turned off and disabled. However, if it is turned off, when ERC-20 tokens are set to be traded, they would also be turned off by default unless the system automatically changes this (which is not ideal since it is bad UX to change an input which a user may have explicitly changed in the past).

A better solution is to remove the toggle completely when it is not relevant, and provide better descriptions:

gasless-trading-toggle

Test this by checking the states match up to the diagram above.

**2. Prevent background scrolling on modals**

Test by opening a modal (e.g. Connect wallet) on any page and try scrolling downwards. Best to try it on longer pages like the home page or the veBAL page.

**3. Better hover and focus states for BalInput component**

New hover states and focus states when using the BalInput component. 

Screen Shot 2022-05-23 at 2 21 34 PM

Test by hovering and focusing into BalInputs, like on the Trade UI or while Adding liquidity, 

## Checklist:


src/components/_global/BalModal/BalModal.vue
```diff
@@ -1,5 +1,5 @@
 <script lang="ts" setup>
```
pkattera, line 1: Changes to this file are to prevent background scrolling while modals are present

garethfuller (approved):
Looks great, thanks! 🙌

PR #1936, opened 2022-06-01
Refactor types

garethfuller:
# Description

General refactor of types:
- Moved pool types from `/services/balancer/subgraph/types.ts` to `/services/pool/types.ts`
- Removed `FullPool` type.
- Separated subgraph implementation-specific types from pool types.
- Removed duplicate pool type definitions.
- Moved duplicate implementations of snapshot logic to composable `useSnapshots.ts`

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test general functionality hasn't been broken. Now that some Pool attributes are conditional I had to add a bunch of conditional method chaining and default values to fall back on.

## Visual context

n/a

## Checklist:


src/components/contextual/pages/pool/MyPoolBalancesCard.vue
```diff
@@ -119,7 +119,7 @@ const showMigrateButton = computed(
  * METHODS
  */
 function weightLabelFor(address: string): string {
-  return fNum2(props.pool.onchain.tokens[address].weight, {
+  return fNum2(props.pool?.onchain?.tokens?.[address]?.weight || '0', {
```
timjrobinson, line 122: Is this just null while it's loading? If so I think it should return '-' instead of '0' like `weightFor` in DefaultPool.vue does. Or at least these two functions should return the same value when it doesn't exist.

PR #1950, opened 2022-06-06
[WIP] Use sdk for liquidity calculations

timjrobinson:
# Description

This branch is for testing integration with the new pool liquidity calculations added in https://github.com/balancer-labs/balancer-sdk/pull/78

It does the following:
- Updates types to be compatible with SDK types
- Changes TokenPrice type to store prices as strings
- Runs SDK liquidity calculations alongside the current frontend ones to ensure they return the same result. 

This PR shouldn't be merged. It is to ensure all liquidity calculations in the SDK return the exact same results as the frontend currently does, and to discuss any type changes that would be needed to allow the frontend to use this SDK for liquidity calculation. 

In the future the frontend will pull liquidity values from the Balancer API but it always needs to be able to use this as a fallback in case the API is un-accessible, unusable on a network, or someone forks the frontend and doesn't want to use our API.


timjrobinson:
This has been merged into the SDK and so this PR is no longer neccessary as it was just a demo.

PR #1956, opened 2022-06-07
Change pool chart

alter-eggo:
# Description

This pr changes pool chart. On pool page are added 3 charts with volume, tvl and fees historical data.
Also echarts library size is reduced by 20% (153 -> 124 kB gZipped), unnecessary imports were deleted.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:


src/components/contextual/pages/pool/PoolChart.vue
```diff
@@ -21,198 +23,309 @@ type Props = {
...
 
-  const pricesTimestamps = Object.keys(props.historicalPrices);
-  const snapshotsTimestamps = Object.keys(props.snapshots);
+const periodOptions = computed(() => {
+  const maxPeriodLengh = snapshotValues.value.length;
+  const arr = [{ text: t('poolChart.period.all'), value: maxPeriodLengh }];
 
-  if (snapshotsTimestamps.length === 0) {
-    return [];
+  if (maxPeriodLengh > 365) {
```
timjrobinson, line 85: I think we should show all periods no longer how old the pool is and just only show the data we have. Otherwise it may seem the menu is broken / missing options.

src/components/contextual/pages/pool/PoolChart.vue
```diff
@@ -21,198 +23,309 @@ type Props = {
...
+  const periodSnapshots =
+    currentPeriod.value === snapshotValues.value.length
+      ? snapshotValues.value
+      : snapshotValues.value.slice(0, currentPeriod.value - 1);
+  const isAllTimeSelected =
+    periodSnapshots.length === snapshotValues.value.length;
+  const pariodLastSnapshotIdx = periodSnapshots.length - 1;
+
+  if (activeTab.value === PoolChartTab.TVL) {
+    const tvlValues = periodSnapshots.map((snapshot, idx) => ({
```
timjrobinson, line 114: Each of these if statements could be put into a separate function.

src/components/_global/BalChart/BalChart.vue
```diff
@@ -69,7 +76,12 @@ type PeriodOption = {
 };
 
 export default defineComponent({
```
timjrobinson, line 78: Would be good to convert this component into using script setup instead of defineComponent. We've been slowly converting components over time.

src/components/contextual/pages/pool/PoolChart.vue
```diff
@@ -21,198 +23,309 @@ type Props = {
...
-      const price = prices[index];
+function setCurrentPeriod(period: string) {
+  currentPeriod.value = Number(period);
+}
 
-      return price * parseFloat(amount);
-    })
-    .reduce((total, value) => total + value, 0);
+function setCurrentChartValue(value: {
+  value: [string, number];
```
timjrobinson, line 254: could have a more descriptive name so that you're not doing `value.value`. Also it could be an Object instead of array to describe what items 0 and 1 are.

src/composables/queries/usePoolSnapshotsQuery.ts
```diff
@@ -52,9 +52,16 @@ export default function usePoolSnapshotsQuery(
     let prices: HistoricalPrices = {};
 
     const isStablePhantomPool = isStablePhantom(pool.value.poolType);
+    const daysFromPoolCreation = differenceInDays(
```
timjrobinson, line 55: Rather than modifying this function would be better to have a `usePoolSnapshotSinceCreationQuery` that calls this function with the days from pool creation. So that we can still re-use this function in the future with an arbitrary number of days.

src/composables/useNumbers.ts
```diff
@@ -14,7 +14,8 @@ interface Options {
 export interface FNumOptions extends Intl.NumberFormatOptions {
   fixedFormat?: boolean; // If true, don't auto-adjust based on number magnitde
   abbreviate?: boolean; // If true, reduce number size and add k/M/B to end
-  dontAdjustLarge?: boolean; // If true, don't auto-adjust if the number is large
+  dontAdjustLarge?: boolean; // If true, don't auto-adjust if the number is large,
+  fractionDigits?: number; // set fraction digits
```
timjrobinson, line 18: You can just pass `maximumFractionDigits` into fNum2 instead of adding this, as it extends `Intl.NumberFormatOptions`

timjrobinson (changes requested):
A few code style comments. Also during testing found:

When loading pool https://beta-polygon-git-pool-charts-balancer.vercel.app/#/pool/0xc17636e36398602dd37bb5d1b3a9008c7629005f0002000000000000000004c4 it:
- When initially loading no time period is selected. 
- When graph is not active it says 'undefined volume'. This seems to be because of the above. 
- TVL is showing $NaN for a few graphs on Polygon

I think all pools should show all time periods no matter how new they are, then if the time period is longer than the total time it's been alive it just shows max.

src/components/contextual/pages/pool/PoolChart.vue
alter-eggo: fixed, now all periods are shown

src/components/contextual/pages/pool/PoolChart.vue
alter-eggo: fixed

src/components/_global/BalChart/BalChart.vue
alter-eggo: fixed

src/components/contextual/pages/pool/PoolChart.vue
alter-eggo: fixed, I changed the format

src/composables/useNumbers.ts
alter-eggo: fixed

src/composables/queries/usePoolSnapshotsQuery.ts
alter-eggo: I made it in different way. now function use days if they are provided in args, if not - days from pool creation. 
so we still can reuse this function in the future

pkattera (approved):
I've tested this across a few browsers across various screensizes and it works as intended. Great work Anastas!

src/components/_global/BalChart/BalChart.vue
```diff
@@ -315,203 +205,228 @@ export default defineComponent({
...
-      return style;
-    });
+        color: '#FFF',
+        fontSize: 10
+      },
+      data: props.isLastValueChipVisible
+        ? [
+            {
+              name: 'Latest',
+              yAxis: (last(props.data[i]?.values) || [])[1]
```
timjrobinson, line 257: This will error if props.data[i] is an empty array or null, as it'll try and read the 2nd value of an empty array. It does look like this was in the old code though, so maybe it's not possible for props.data[i] to be empty.

src/services/coingecko/api/price.service.ts
```diff
@@ -189,12 +189,12 @@ export class PriceService {
           for (const key in result) {
             const value = result[key];
             const [timestamp, price] = value;
-            if (timestamp > dayTimestamp * 1000) {
-              prices[dayTimestamp * 1000] = price;
-              dayTimestamp += twentyFourHoursInSecs;
+            if (timestamp > start * 1000) {
```
timjrobinson, line 192: Are the timestamps always divisable by 86400 / 3600? As it looks like that's what the dayTimestamp variable was doing previously, making sure that the prices array had an entry every 86400.

timjrobinson (commented):
Looking good, a few small issues:

- wstETH/ETH stable pool and the boosted pool are both showing no data. 
	- It seems to be because Coingecko is returning 429 errors. This could just be my internet being bad but is it possible you're performing too many token data requests causing this? I'm seeing this a lot on the boosted pool too.
- On boosted pool I'm seeing this error in console: 
		- ```TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at U.serialize (index.ts:45:14)
    at U.get (index.ts:40:17)
    at async Object.i [as queryFn] (usePoolSnapshotsQuery.ts:68:7)```
- Time period selection dropdown is unstyled.

src/components/_global/BalChart/BalChart.vue
alter-eggo: yeah, seems that it's not possible for props.data[i] to be empty. nevertheless, fixed it

alter-eggo:
- fixed error with boosted pools
- yeah, coingecko is returning 429 errors sometimes. there are two token data requests as it is on production now. the difference is in request args. may be that is the problem. now much more data is requested, because we need prices for timestamps since pool creation day. actually yet don't know what to do with it(
- @pkattera how can we style time selection period? or should I use `BalDropdown` component?

alter-eggo:
@timjrobinson Time period selection dropdown is styled now

timjrobinson (changes requested):
Looking great! Just one more issue - the TVL for the boosted pool (https://beta-app-git-pool-charts-balancer.vercel.app/#/pool/0x7b50775383d3d6f0215a8f290f2c9e2eebbeceb20000000000000000000000fe) isn't showing for me. Also there's a merge conflict.

src/components/contextual/pages/pool/PoolChart.vue
```diff
@@ -117,33 +120,66 @@ const timestamps = computed(() =>
...
-    if (idx === 0) {
-      tvlValues.push(
-        Object.freeze([timestamp, Number(props.totalLiquidity || 0)])
-      );
-      return;
-    }
 
-    const prices = props.historicalPrices[snapshot.timestamp];
+  // temporary statement until we start get prices from coingecko for
+  if (props.poolAddress === BALANCER_BOOSTED_POOL_ADDRESS) {
```
timjrobinson, line 125: It would be better do to `if (isStablePhantom(props.pool.poolType)` rather than hard coding the boosted pool address, just in case we never get data in coingecko and we will probably see this pop up for future boosted pools.

timjrobinson (changes requested):

timjrobinson (approved):
Just added a check if PoolType is StablePhantomPool instead of having the hard coded address. This is good to merge.

PR #1963, opened 2022-06-08
Refactor pool queries

garethfuller:
# Description

Various refactorings of coupled and messy code in pool queries,  i.e. usePoolQuery, usePoolsQuery & useUserPoolsQuery.

TODO before merge:
- Test changes to stream query doesn't break any functionality on home page.

Next steps for refactoring pool decoration/schema (in a new PR)
- Deep dive into all the additional fetching we're doing around linear pools and generalise. i.e. fetch additional sub-pool data for any pool token that happens to be a pool, not just linear pools of StablePhantom pools.
- Fix weird cyclic dependency in [staking-rewards.service.ts](https://github.com/balancer-labs/frontend-v2/pull/1963/files#diff-ec494d6f45c6897a66b89ac170e96d4aac13d3cc8f71e52133e4ef226475478dR183-R184) where we need accurate total liquidities to calc those APRs, but also we call this function in the [decorator](https://github.com/balancer-labs/frontend-v2/blob/develop/src/services/pool/decorators/pool.decorator.ts#L77) as part of the calls to get supporting data for decoration. So fetching onchain data and calculating total liquidities ends up being done twice. Needs to be re-thought and refactored.

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test pools lists load as expected
- [ ] Test pool pages load as expected

## Visual context

n/a

## Checklist:

### DIFF
--- a/src/components/tooltips/APRTooltip/components/YieldBreakdown.vue
+++ b/src/components/tooltips/APRTooltip/components/YieldBreakdown.vue
@@ -8,6 +8,7 @@ import { isStablePhantom } from '@/composables/usePool';
 import useTokens from '@/composables/useTokens';
 import { includesWstEth } from '@/lib/utils/balancer/lido';
 import { PoolAPRs } from '@/services/pool/types';
+import { getAddress } from '@ethersproject/address';
 
 /**
  * TYPES
@@ -33,7 +34,7 @@ const { t } = useI18n();
 /**
  * COMPUTED
  */
-const yieldAPRTokens = computed(() =>
+const yieldAPRTokens = computed(() => 
   getTokens(Object.keys(props.yieldAPR.breakdown))
 );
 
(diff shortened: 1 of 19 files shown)


garethfuller:
This useStreamedPoolsQuery is not quite fixed. It should load in the basic list like it currently does and then as the last step just use the pool decorator to fetch all the rest of the data rather than doing the aprs and liquidity separately.

timjrobinson (approved):
Made some minor fixes for the portfolio page and everything seems to be working fine in my testing. Two minor issues:

- The staking APR for the boosted pool is slightly higher in this PR than in prod. I think this may be because of the additional on-chain data fetching. 
- Volume is showing `-` until onchain data is loaded, while in production it shows the volume from the graph initially before switching to onchain data. This is a minor issue, I'm looking into it further to see if it can be changed.

alter-eggo (approved):

PR #1973, opened 2022-06-14
Release 1.58.0

timjrobinson:
Weekly Release


Dockerfile.dev
```diff
@@ -1,16 +1,7 @@
-FROM node:14 AS base
+FROM node:14-alpine
```
1saf, line 1: have you verified that the alpine image works?

1saf (approved):

Dockerfile.dev
timjrobinson: Yea @gmbronco has been using it for development. I just tried it then and it worked for me.

PR #1981, opened 2022-06-14
TRM Sanctioned Wallets

1saf:
# Description

Checks if a wallet is sanctioned against the TRM labs sanctions API. If sanctioned the wallet connector will disconnect the wallet immediately while also prompting the user as to why.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

Connect with non sanctioned wallet, should be able to use app.
Connect with sanctioned wallet - should not be able to transact with app.

## Visual context

N/A

## Checklist:


src/services/web3/web3.plugin.ts
```diff
@@ -176,6 +190,12 @@ export default {
         // for when user reloads the app on an already connected wallet
         // need to store address to pre-load that connection
         if (account.value) {
+          // fetch sanctioned status
+          const _isSanctioned = await isSanctionedAddress(account.value);
+          isSanctioned.value = _isSanctioned;
+          if (_isSanctioned) {
+            disconnectWallet();
```
alter-eggo, line 197: may be add here `return statement`? 
now function disconnects users and then set all statements to `connected`

PR #1999, opened 2022-06-19
whitelist tetuBAL-80BAL-20WETH V2

belbix:
# Description

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## How should this be tested?

- [] the pool should be visible at https://polygon.balancer.fi/#/pool/0xb797adfb7b268faeaa90cadbfed464c76ee599cd0002000000000000000005ba

## Visual context

-

## Checklist:


timjrobinson (approved):
Do we still need the old pool in the whitelist? When should it be removed? We don't want users to get confused. 

Compared create transactions:

https://polygonscan.com/tx/0xd3caae1e09324fd3d25694095e05483cd1a6be0114b24cb483b195a4b123955f
https://polygonscan.com/tx/0x85aabe2964e1968587c2ab711c09b6c6d77d42e2f8e9737dc56a7752c35eb14b

And it was created with same parameters from same contract just with a different name, so it looks correct.

belbix:
> Do we still need the old pool in the whitelist? When should it be removed? We don't want users to get confused.
> 
> Compared create transactions:
> 
> https://polygonscan.com/tx/0xd3caae1e09324fd3d25694095e05483cd1a6be0114b24cb483b195a4b123955f https://polygonscan.com/tx/0x85aabe2964e1968587c2ab711c09b6c6d77d42e2f8e9737dc56a7752c35eb14b
> 
> And it was created with same parameters from same contract just with a different name, so it looks correct.

We need a new pool with a shorter name for the reason the gauge factory has restrictions for pool names.
The old pool needs to keep temporally - users will need to migrate, without UI it is pretty complex.

PR #2017, opened 2022-06-27
TRM Sanctioned Wallets

1saf:
# Description

#1981


timjrobinson:
Two minor issues:
- On my first connect with coinbase wallet the modal didn't close and the close button wouldn't work. I couldn't reproduce again though so could be a race condition that is also present in prod. 
- The Balancer grants link isn't work. It does to https://balancer.community/balancer-community-grants which fails to load.

timjrobinson (approved):

PR #2022, opened 2022-06-28
Chore/Tests for GaugesTable and generic unit testing setup

sssmi:
# Description

- Add tests for GaugesTable
- Make jest setup files per environment (jsdom or node)
- Use real locale data by removing vue-i18n mock, and add i18n as a Vue plugin for `@vue/test-utils`.  When testing composables, i18n needs still to be mocked, because composables are tested with `vue-composable-tester`.
- Add some global components to `@vue/test-utils` config, so there is no need to import them to every test file individually.
- Add global stubs for RouterLink and Jazzicon components
- Use `useI18n` composable in `web3.plugin.ts` instead of i18n library directly.

## Type of change

- [x] Code refactor / cleanup

## Checklist:


alter-eggo (approved):

src/composables/pools/usePoolCreation.spec.ts
```diff
@@ -6,7 +6,14 @@ import usePoolCreation, { PoolSeedToken } from './usePoolCreation';
 
 const tokens: Record<string, PoolSeedToken> = {};
 
-jest.mock('vue-i18n');
+jest.mock('vue-i18n', () => ({
```
timjrobinson, line 9: You could still keep the mock file and stop it from auto-mocking in other files with `jest.unmock('vue-i18n');` in the main jest file, as node_modules are always mocked by default.

timjrobinson (approved):
LGTM. One minor change.

src/composables/pools/usePoolCreation.spec.ts
sssmi: 👍 Fixed

PR #1898, opened 2022-05-17
use the decentralized graph endpoint

gtaschuk:
# Description

This uses the decentralized graph as the subgraph endpoint

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Documentation or wording changes
- [ ] Other

## How should this be tested?

- [ ] Verify data flows from graphql

## Visual context

Please provide any relevant visual context for UI changes or additions. This could be static screenshots or a loom screencast.

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have requested at least 2 reviews (If the PR is significant enough, use best judgement here)
- [ ] I have commented my code where relevant, particularly in hard-to-understand areas
- [ ] If package-lock.json has changes, it was intentional.
- [ ] The base of this PR is `master` if hotfix, `develop` if not

PR #2023, opened 2022-06-28
Use the decentralized graph endpoint

timjrobinson:
Original #1898 
Revert #2021 

This is the PR to re-add the decentralized graph endpoint. It broke a few hours ago so not merging again until we've determined the cause. 

## Changes Required

- [ ] Add Fallback to hosted graph endpoint if request fails
- [ ] Add logging to sentry if request falis


timjrobinson:
I think this would work better as an array of endpoints that it iterates through. Could add it to that existing `subgraphs` config item as `main` which has an array of URL's. Then it recursively calls the request function from each, logging an error and falling back to the next as they fail, rather than having that duplicate `getByFallbackUrl` function. 

This way if we add an ability for users to set their own graph endpoint in the future it can just be prepended to the beginning of the array. Or we could implement a system where it finds the endpoint with the lowest latency and uses that. It'd just make it more flexible for future usage by having an array of graph endpoints. 

If we could do this in a way that we can easily re-use it for rpc fallbacks that would also be awesome. As we want to allow having user injected RPC, infura, and alchemy and it falls back from user -> infura -> alchemy (order may change) if they fail. 

There's also this spot on the claims data page: https://github.com/balancer-labs/frontend-v2/blob/develop/src/composables/useClaimsData.ts#L51 that needs to use the fallback somehow.

src/services/balancer/subgraph/balancer-subgraph.client.ts
```diff
@@ -1,29 +1,25 @@
...
-        data: { data }
-      } = await axios.post(this.url, payload);
-      return data;
+      const response = await subgraphFallbackService.get(payload);
+
+      if (!response) {
+        return;
+      }
+
+      return response.data.data;
```
garethfuller, line 15:
```suggestion
      return response?.data?.data;
```

src/services/balancer/subgraph/subgraph-fallback.service.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (this.isSingleSubgraphUrl.value) {
+      return this.url;
+    }
+    return this.subgraphUrls[this.subgraphUrlIndex.value];
+  });
+
+  constructor(
+    private readonly url = configService.subgraph,
+    private readonly subgraphUrls = configService.subgraphUrls || []
+  ) {}
```
garethfuller, line 21: Can we not simplify this to 
```suggestion
  constructor(
    private readonly urls = [configService.subgraph, ...configService.subgraphUrls]
  ) {}
```

src/services/balancer/subgraph/subgraph-fallback.service.ts
```diff
@@ -0,0 +1,57 @@
...
+class SubgraphFallbackService {
+  private subgraphUrlIndex = ref(0);
+
+  public isSingleSubgraphUrl = computed(() => this.subgraphUrls.length === 0);
+  public currentSubGraphUrl = computed(() => {
+    if (this.isSingleSubgraphUrl.value) {
+      return this.url;
+    }
+    return this.subgraphUrls[this.subgraphUrlIndex.value];
+  });
```
garethfuller, line 16: Then this just becomes:
```suggestion
  public currentURL = computed(() => {
    return this.urls[this.subgraphUrlIndex.value];
  });
```

src/services/balancer/subgraph/subgraph-fallback.service.ts
```diff
@@ -0,0 +1,57 @@
...
+        }] failed with message: ${
+          (error as Error).message
+        }. Payload: ${payload}`
+      );
+      this.changeSubgraphUrlIndex();
+
+      if (this.subgraphUrlIndex.value + 1 > this.subgraphUrls.length) {
+        throw error;
+      }
+      this.get(payload);
```
garethfuller, line 48: And this can be simplified too:
```suggestion
      captureException(
        `GraphQL request to [${
          this.currentSubGraphUrl.value
        }] failed with message: ${
          (error as Error).message
        }. Payload: ${payload}`
      );
  
      if (this.urlIndex.value + 1 > this.urls.length) {
        throw error;
      }
 
      this.incrementUrlIndex();
      this.get(payload);
```

src/services/balancer/subgraph/subgraph-fallback.service.ts
```diff
@@ -0,0 +1,57 @@
+import { captureException } from '@sentry/browser';
+import axios, { AxiosResponse } from 'axios';
+import { computed, ref } from 'vue';
+
+import { configService } from '@/services/config/config.service';
+
+class SubgraphFallbackService {
+  private subgraphUrlIndex = ref(0);
```
garethfuller, line 8: Simplify naming.
```suggestion
  private urlIndex = ref(0);
```

src/services/balancer/subgraph/subgraph-fallback.service.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  public isSingleSubgraphUrl = computed(() => this.subgraphUrls.length === 0);
```
garethfuller, line 10: We don't need this anymore.

src/services/balancer/subgraph/subgraph-fallback.service.ts
```diff
@@ -0,0 +1,57 @@
...
+      if (this.subgraphUrlIndex.value + 1 > this.subgraphUrls.length) {
+        throw error;
+      }
+      this.get(payload);
+    }
+  }
+
+  private changeSubgraphUrlIndex(): void {
+    this.subgraphUrlIndex.value += 1;
+  }
```
garethfuller, line 54:
```suggestion
  private incrementUrlIndex(): void {
    this.urlIndex.value += 1;
  }
```

src/services/config/config.service.ts
```diff
@@ -57,6 +62,23 @@ export default class ConfigService {
     });
   }
 
+  public get subgraph(): string {
+    return template(this.network.subgraph, {
+      GRAPH_KEY: this.env.GRAPH_KEY
+    });
+  }
```
garethfuller, line 69: I think we could actually just get rid of the single 'subgraph' key from the configs and only have an array.

garethfuller (changes requested):
This needs a bit more work.

markusbkoch:
Quick update on this from the subgraph side: we plan to deploy the latest version of our mainnet subgraph to the decentralized network once the new Linear and ComposableStable factories come out and we have those working on the hosted service (ETA early next week). 

Until then this PR should not be merged, because the recent schema changes haven't been applied to the decentralized deployment, meaning the app would partially break where it relies on the SDK.

Are the changes requested by Gareth all that's left to do? Can we expect this to be finished by EOW?

alter-eggo:
changes that were requested by Gareth will be done by EOW, but I want to double check how it works, I am not happy with current implementation

alter-eggo:
@garethfuller fixed

src/services/balancer/subgraph/subgraph-fallback.service.ts
```diff
@@ -0,0 +1,45 @@
...
+          (error as Error).message
+        }. Payload: ${payload}`
+      );
+
+      if (this.urlIndex.value + 1 > this.urls.length) {
+        throw error;
+      }
+
+      this.incrementUrlIndex();
+      this.get(payload);
```
timjrobinson, line 36: This should have a return on it. Also this file should have tests to confirm it works.

markusbkoch:
> Until then this PR should not be merged, because the recent schema changes haven't been applied to the decentralized deployment, meaning the app would partially break where it relies on the SDK.

Due to the delays to Linear/Composable, we decided to move forward with deploying the same version of the subgraph that we have in production on the hosted service to the decentralized network. Sync is now complete. Notice the IDs are the same in the hosted service and the network deployments
image
image

We've also set up domain allowlisting for `*.vercel.app` and `*.balancer.fi` and query budgets.

If all other concerns have been addressed, especially wrt to falling back to the hosted service in case the decentralized endpoint returns an error, then I believe this can be merged.

markusbkoch:
As discussed in the roadmap call today, I believe this needs to be updated to also account for passing the valid subgraph URL to the SDK/SOR when instantiating it

garethfuller (changes requested):
@alter-eggo  A search still needs to be done for `configService.network.subgraph` and replaced with the fallback service.

garethfuller:
@markusbkoch I think this is pretty much ready to go. Do you have any concerns before we merge (maybe tomorrow)?

src/lib/scripts/voting-gauge.generator.ts
```diff
@@ -193,7 +194,7 @@ async function getPoolInfo(
   retries = 5
 ): Promise<VotingGauge['pool']> {
   log(`getPoolInfo. poolId: network: ${network} poolId: ${poolId}`);
-  const subgraphEndpoint = config[network].subgraph;
+  const subgraphEndpoint = subgraphFallbackService.url.value;
```
markusbkoch, line 197: I think this one needs to be kept as before because it won't be possible to query the decentralized subgraph from the CLI when running the generator script.

markusbkoch (changes requested):

src/lib/scripts/voting-gauge.generator.ts
alter-eggo: fixed

PR #2037, opened 2022-07-01
Pool page optimisation

alter-eggo:
# Description
This pr optimises pool page. 
- information about pools from home page is saved and used on pool page
- aprs are downloading separately from pool (should be actually)
- intersection observer was added to prevent first render of investment and trades tables
- removed refetching of snapshots and prices

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:


timjrobinson:
Just testing it out and comparing results to prod. I loads faster which is nice, and most of the data is the same. There is just inconsistencies with the APR's ad TVL for some:

- wstETH/ETH pool is about 1% diff in rewards
- Boosted pool is similar rewards but slightly different from prod
- veBAL pool only shows one number instead of a range and when hovering it shows 0% for locking APR
	- Would be good if this could break that APR into the percent in BAL and percent in bb-a-USD too
- The auraBAL/veBAL pool (0x3dd0843a028c86e0b760b1a76929d1c5ef93a2dd000200000000000000000249) is not showing liquidity or APR's, but it is showing them in prod. 

Could you work with @gmbronco to see why those APR's are different and if they're incorrect in the frontend or in the SDK or the reasons they they differ.

src/composables/queries/usePoolQuery.ts
```diff
@@ -47,21 +53,25 @@ export default function usePoolQuery(
...
+      });
+    }
 
     if (isBlocked(pool, account.value)) throw new Error('Pool not allowed');
 
     // Decorate subgraph data with additional data
     const poolDecorator = new PoolDecorator([pool]);
-    const [decoratedPool] = await poolDecorator.decorate(
-      subgraphGauges.value || [],
+    const decoratedPool = await poolDecorator.decorateSinglePool(
```
garethfuller, line 74: Why do we need a separate function for this?

src/pages/pool/_id.vue
```diff
@@ -3,116 +3,20 @@
...
+        :poolTypeLabel="poolTypeLabel"
+        :loadingApr="loadingApr"
+        :pool="pool"
+        :poolApr="poolApr"
+        :isStableLikePool="isStableLikePool"
+        :noInitLiquidity="noInitLiquidity"
+        :titleTokens="titleTokens"
+        :missingPrices="missingPrices"
+        :isLiquidityBootstrappingPool="isLiquidityBootstrappingPool"
+        :isStablePhantomPool="isLiquidityBootstrappingPool"
```
garethfuller, line 17: Can a bunch of these props not be removed by moving the computed properties into the PoolPageHeader component?

src/pages/pool/_id.vue
```diff
@@ -210,106 +127,120 @@ interface PoolPageData {
...
      * STATE
      */
     const data = reactive<PoolPageData>({
       id: route.params.id as string
     });
 
-    /**
-     * COMPUTED
-     */
+    //#region pool query
```
garethfuller, line 154: What are these comments for?

src/pages/pool/_id.vue
```diff
@@ -210,106 +127,120 @@ interface PoolPageData {
...
+
+    function addIntersectionObserver(): void {
+      if (
+        !('IntersectionObserver' in window) ||
+        !('IntersectionObserverEntry' in window) ||
+        !intersectionSentinel.value
+      ) {
+        isSentinelIntersected.value = true;
+        return;
+      }
```
garethfuller, line 214: Where is this pattern coming from? I'd like to understand it better.

src/services/pool/decorators/pool.decorator.ts
```diff
@@ -35,41 +35,73 @@ export class PoolDecorator {
...
+    poolService.removePreMintedBPT();
+    await poolService.setLinearPools();
+    poolService.setOnchainData(rawOnchainDataMap[pool.id], tokens);
+    poolService.setTotalLiquidity(prices, currency, tokens);
+    poolService.setFeesSnapshot(poolSnapshot);
+    poolService.setVolumeSnapshot(poolSnapshot);
+    poolService.setUnwrappedTokens();
+
+    return poolService.pool;
+  }
```
garethfuller, line 103: Why is this needed? Is it just decoration without APR? If it is, maybe we should just add a flag to the decorate function like `excludeAPR: boolean` or something. Then skip that part in the function. Having two functions that repeat 90% of the code is not very dry.

src/services/pool/decorators/pool.decorator.ts
```diff
@@ -78,14 +110,15 @@ export class PoolDecorator {
...
+    const blockNumber = getTimeTravelBlock(currentBlock);
     const block = { number: blockNumber };
     const isInPoolIds = { id_in: this.pools.map(pool => pool.id) };
     try {
-      return await this.poolSubgraph.pools.get({
+      const data = await this.poolSubgraph.pools.get({
         where: isInPoolIds,
         block
       });
+      return data;
```
garethfuller, line 121: Looks like this might have been added for logging? Unnecessary const.

garethfuller (changes requested):
In general, looks great and the pool pages are loading much faster now! ⚡ Just a few minor comments. Nice work! 🙌

src/pages/pool/_id.vue
alter-eggo: https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API
in this case Intersection Observer helps to render Investments and Trades tables on scrolling to them. it increases the speed of first page render.

src/pages/pool/_id.vue
alter-eggo: This region comments helps to separate code to logical blocks. It improves readability  
+ it helps to collapse and hide sections of code in ide

src/services/pool/decorators/pool.decorator.ts
alter-eggo: yeah, fixed

src/services/pool/decorators/pool.decorator.ts
alter-eggo: added logic with `includeAPR` argument

src/pages/pool/_id.vue
alter-eggo: moved one to `PoolPageHeader`, others are used in `_id.vue` file

garethfuller (approved):
I think this is ready to merge. The decorate function isn't ideal but it works well and should only be a temp fix until we have the API.

PR #2085, opened 2022-07-22
Fix disabled inputs and remove custom styling

garethfuller:
# Description

We had custom input styling in the gauge vote form which looked a bit wonky. In general, we should not be custom styling inputs anywhere so we can maintain consistency across the app. This PR reverts the custom styling and adds some global styling for disabled inputs.

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

n/a

## Visual context

Before:
Screenshot 2022-07-22 at 15 41 36

After:
Screenshot 2022-07-22 at 15 48 16

## Checklist:


pkattera:
@garethfuller Agree with using global sizes and styles and removing custom styling.

One thing – the disabled input still feels like it can allow user input.

Is it possible at a global level to remove the shadows and change the background color a bit more (for disabled inputs `bg-50` feels too light and `bg-100` probably feels too dark but probably makes it more clear that no input is possible)

disabled-buttons

sssmi:
> @garethfuller Agree with using global sizes and styles and removing custom styling.
> 
> One thing – the disabled input still feels like it can allow user input.
> 
> Is it possible at a global level to remove the shadows and change the background color a bit more (for disabled inputs `bg-50` feels too light and `bg-100` probably feels too dark but probably makes it more clear that no input is possible)

To add to @pkattera suggestion on disabled styles, could the suffix "%" have a lighter color as well when the input is disabled, so it would fit better to the background?

garethfuller:
The problem with making the disabled input more obvious is the effect it has on the trade UI inputs. They are disabled initially until the SOR is loaded which results in this flicker. When the disabled state is harsh it's quite an ugly flicker. @pkattera maybe you could look into this a bit more and take on this PR?

pkattera:
@garethfuller I can take a look at a better global disabled state.

Regarding the Trade UI inputs, I think it's currently a poor experience where the fields are disabled initially but look enabled. I'm always trying to click into the field and think the UI is broken before remembering the field is disabled. Is there a good reason for disabling the Trade UI input fields before the SOR loads? Ideally, the user could start entering amounts, but the preview button could remain disabled as the SOR is loading. In an ideal flow, by the time the user has entered an amount, the SOR would have loaded in the background along with the exchange rates etc.

sssmi (approved):

pkattera (approved):
Looks good to me. At some point we should consider also having a disabled style for any appended items (like the % in this case), but it's already a big improvement, so let's ship it.

PR #2092, opened 2022-07-28
Update eslint

sssmi:
# Description

- Update eslint and its plugins
- Added `eslint-plugin-tailwindcss` plugin
- Tighten up Vue rules from `plugin:vue/vue3-essential` to `plugin:vue/vue3-recommended`
- Use plain eslint plugins instead of vue-cli ones. Running linter on webpack only slows down local development. Running eslint before commit and in CI should suffice.
- Integrate prettier as eslint plugin, so prettier can be used to format Vue components.
- Prettier is now only used to format only .md, .json, .yaml and .html files. Eslint (with prettier integrated) handles .js, .ts and .vue files
- Upgrated husky, and added `lint-staged` to pre-commit hook, so linters are run against staged files.

## Type of change

- [x] Dependency changes
- [x] Code refactor / cleanup

## How should this be tested?

Run `npm run lint:fix`. See if the autofixed issues make sense. Also look at the issues that require manual fixing. Do they make sense?

## Visual context

## Checklist:


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npm run lint:fix
+# npm run lint:fix -- --cache
```
sssmi, line 4: I will uncomment this before merge

.eslintrc.js
```diff
@@ -5,21 +5,10 @@ module.exports = {
     node: true
   },
 
-  globals: {
```
sssmi, line 8: All these globals are not needed anymore

.eslintrc.js
```diff
@@ -31,17 +20,15 @@ module.exports = {
 
   rules: {
     'no-console': 'off',
-    'no-debugger': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
+    'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off',
     '@typescript-eslint/no-explicit-any': 'off',
     '@typescript-eslint/ban-ts-ignore': 'off',
-    '@typescript-eslint/camelcase': 'off',
```
sssmi, line 37: Removed some rules, that don't do anything anymore

.eslintrc.js
```diff
@@ -31,17 +20,15 @@ module.exports = {
...
     '@typescript-eslint/no-var-requires': 'off',
-    '@typescript-eslint/no-use-before-define': 'off',
-    '@typescript-eslint/explicit-module-boundary-types': 'off',
     '@typescript-eslint/ban-ts-comment': 'off',
-    'simple-import-sort/imports': 'warn',
-    'simple-import-sort/exports': 'warn'
+    
+    // New rules
+    'vue/component-name-in-template-casing': ['error', 'PascalCase'],
+    'vue/multi-word-component-names': 'off'
```
sssmi, line 31: Vue recommends multi word component names, so they don't get mixed with regular html elements, but many components are already single word, so had to turn this off.

garethfuller (commented):
Thanks, looks good. Will you run it now? I can approve after that.

timjrobinson (approved):
LGTM. What's the reasoning behind passing component properties as dash case rather than camelCase? I thought camelCase made more sense becuase that's how they're defined as properties in the Props object so it's easier to find where they're used, but I'm not too fussed either way.

sssmi:
> LGTM. What's the reasoning behind passing component properties as dash case rather than camelCase? I thought camelCase made more sense becuase that's how they're defined as properties in the Props object so it's easier to find where they're used, but I'm not too fussed either way.

Good point. kebab-case props is the default rule in recommended settings in [vue/attribute-hyphenation](https://eslint.vuejs.org/rules/attribute-hyphenation.html)

Setting the `vue/attribute-hyphenation` to `never` might make more sense.

garethfuller (approved):
Looks good, I went through some basic testing in the previews and nothing seems to have been affected, nice work! 🙌

timjrobinson (approved):

insulineru:
Good job, @sssmi 👍🏻

PR #2093, opened 2022-07-28
Use Balancer API to request pools

timjrobinson:
This is a demo of using the new Balancer GraphQL API to load pool data as added in PR: https://github.com/balancer-labs/balancer-api/pull/30

This shouldn't be merged.


garethfuller:
@timjrobinson should we close this?

PR #2100, opened 2022-08-01
Feature - Pool locking module

1saf:
# Description

Adds a module to the 80/20 WETH/BAL lockable pool similar to the staking pool, which allows users to get a quick glance at their lockable status.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

Open up. the 80/20 WETH/BAL pool and then see if you can see the module. On connected wallet should show locked and unlocked LP tokens as well as test expiry of lock.

## Visual context

image

## Checklist:


src/components/_global/BalAccordion/BalAccordion.vue
```diff
@@ -173,7 +173,7 @@ watch(
 
 <template>
   <div ref="wrapperElement">
-    <BalCard hFull no-pad shadow="none" class="rounded-xl">
```
1saf, line 176: Need to test if this breaks other parts of app

timjrobinson:
When the lock has expired could we make the lock button say "Re-lock" and it does a BalActionSteps of redeem -> lock? Doesn't need to be in this PR but seems like something users would want.

timjrobinson:
When I first opened the 80/20 pool it didn't load the graph or the bit below "Lock LP for veBAL". It had the error: `instrument.js:109 TypeError: Cannot read properties of null (reading 'insertBefore')` in the console without a useful stack trace.

After a hard refresh the graph loads but the bit below "Lock LP for veBAL" is still loading and when I click on it it puts a cursor in it like it's an input field. 

I don't have a wallet connected, using chrome.

1saf:
> 

Will investigate. Thanks for raising

1saf:
> When the lock has expired could we make the lock button say "Re-lock" and it does a BalActionSteps of redeem -

I think best suited for another issue - @pkattera thoughts?

pkattera:
> > When the lock has expired could we make the lock button say "Re-lock" and it does a BalActionSteps of redeem -
> 
> I think best suited for another issue - @pkattera thoughts?

Yeah, I'm happy to create another ticket for this. Ideally, you'd tap a button either called 'Re-lock' or 'Extend lock' and there would be a two step process to redeem and then lock.

pkattera:
I may also be experiencing the same error as Tim – there is a forever loading section below the new module, and I'm not able to click into it:
https://www.loom.com/share/77a7feaa3c4a4b3aaac52b3a7125406c

1saf:
@pkattera @timjrobinson Done deal!

pkattera (commented):
The bugs from last time have been fixed. 

One minor issue – right now both the 'lock' button and 'redeem' button for expired locks, just link through to the /vebal page. Ideally:
- The lock button should link to: /#/get-vebal?returnRoute=vebal

Screen Shot 2022-08-12 at 3 45 44 PM

- The Redeem button should pop up this modal:
Screen Shot 2022-08-12 at 3 43 54 PM

pkattera:
The new changes work fine for me. As discussed with @1saf – the only small thing is to properly disable the 'Lock' button link which still allows people to click through, even if they have no lockable BPT.

garethfuller (approved):

Issue #2097, opened 2022-08-01
Friendly pool names


pkattera:
**Is your feature request related to a problem? Please describe.**
As more pools with 8+ tokens are introduced, it becomes harder for users to parse pool names. Upcoming Managed Pools also need friendly names for marketing purposes. 

**Describe the solution you'd like**

**Home page in the Investment Pools table**
- If a pool has a friendly name, replace the current pool symbols with the friendly name.
- If a pool has it's own pool icon, replace the underlying assets with the pool icon.

**Pool page**
- If a pool has a friendly name, replace the currently title (which displays the pool type) with the friendly name. 
- If a pool has a friendly name, add new text in the description to state what the pool type is. 

**Additional context**

<img width="2268" alt="friendly-pool-names" src="https://user-images.githubusercontent.com/1121139/182113477-72be1f88-1240-49d0-9db9-da1da796c113.png">

pkattera:
@garethfuller Can you please add some context around any implementation solutions you recommend on how to maintain a list of friendly pool names?

PR #2103, opened 2022-08-02
Feature: Friendly pool names

Matthews3301:
# Description

Use "friendly names" for pools - that is a custom name that is different to pool composition or type.

Closes https://github.com/balancer-labs/frontend-v2/issues/2097

## Type of change

In pools table and pool page header, certain pools should now have a custom name instead of default pool composition/type. Also, in pool table, a custom icons are being used.

## How should this be tested?

Check pool id "0x7b50775383d3d6f0215a8f290f2c9e2eebbeceb20000000000000000000000fe" which now has a name "Balancer Boosted Aave USD".

## Visual context

Custom icons in pools table.

## Checklist:


src/constants/pools.ts
```diff
@@ -484,6 +484,26 @@ const POOLS_MAP = {
...
+  '0x5c6ee304399dbdb9c8ef030ab642b10820db8f56000200000000000000000014':
+    'Balancer Protocol Liquidity',
+  '0xdc2ecfdf2688f92c85064be0b929693acc6dbca6000200000000000000000701':
+    'Balancer Protocol Liquidity',
+  '0xf8a0623ab66f985effc1c69d05f1af4badb01b00000200000000000000000060':
+    'Balancer Protocol Liquidity',
+  '0x3dd0843a028c86e0b760b1a76929d1c5ef93a2dd000200000000000000000249':
+    'AuraBAL Stable Pool',
+};
+
```
garethfuller, line 506: Please make this network specific. i.e. update the `Pools` type to have a `Names` attribute. Then in each of the networks add the Names attribute along with any relevant mappings.

garethfuller (changes requested):

src/constants/pools.ts
Matthews3301: Updated

src/components/tables/PoolsTable/PoolsTable.vue
```diff
@@ -273,19 +274,31 @@ function lockedUntil(lockEndDate?: number) {
       </template>
       <template #iconColumnCell="pool">
         <div v-if="!isLoading" class="py-4 px-6">
-          <BalAssetSet :addresses="orderedTokenAddresses(pool)" :width="100" />
+          <BalAssetSet
+            :addresses="
+              POOLS.Names[pool.id]
+                ? [pool.address]
+                : orderedTokenAddresses(pool)
+            "
```
garethfuller, line 282: Can we maybe make these addresses a computed property and do a check to make sure we have an icon for a pool.address first before we use it, otherwise just default to token addresses.

garethfuller (changes requested):

src/components/tables/PoolsTable/PoolsTable.vue
Matthews3301: Do you mean add new field `iconAddresses` to `Pool` type? Or move this logic to a method? It's not possible to have this line as a computed property within `PoolsTable`.

garethfuller: Ahh right, yeah it can't be a computed property but we could have a method like:
```ts
function iconAddresses(pool) {
  // Check if we have a logoURI for pool address, if so return pool address
  // else use pool token addresses
}
Again, let's not add this to the Pool type just keep it contained in the PoolsTable.

garethfuller: You will need to use the useToken composable again to make this function work.

Matthews3301: It's a non-trivial task to check if we have the icon for the `pool.address` - as token information sometimes includes `logoURI` which returns 404 - e.g. for AuraBAL: "https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0x3dd0843A028C86e0b760b1A76929d1C5Ef93a2dd/logo.png"

Maybe instead, we should try to ensure that all the pools with friendly names definitely have the correct logoURIs?

Matthews3301:
Continued in https://github.com/balancer-labs/frontend-v2/pull/2107

PR #2129, opened 2022-08-10
feature: Single url, all networks

Matthews3301:
# Description

Moving away from using subdomains for separate networks. More info here: https://www.notion.so/One-app-many-networks-0ec226994c1542c4af7a1d4be6be943b

1. Change network by using dropdown - it reloads the page, however, the url doesn't require subdomain anymore.
2. Open specific network by using url query string, e.g. `?networkId=137`.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

Change network using networks dropdown.

## Visual context

No changes.

## Checklist:


garethfuller:
> Open specific network by using url query string, e.g. ?networkId=137.
Could we not just use URL params? I've kind of suggested that in the URL structure table in notion. Also, lets stick to friendly network names instead of ids in the URLs please.

garethfuller:
On switching networks, it looks like tokenlists aren't being changed? I haven't looked at the code yet, but we'll also have to essentially reset the token registry which is controlled from tokens.provider.ts.
Screenshot 2022-08-17 at 15 31 20

garethfuller:
When switching network on a pool page and redirecting back to the home page of the new network, please use the notification system like Pon suggested to say something like, Pool doesn't exist on <old-network>.

garethfuller:
When switching network on pages like /:network/trade, please don't redirect back to the home page. Instead please switch to the same page but with the new network. The only place we should need a redirect is on the pool pages.

Matthews3301:
@garethfuller thank you, updated. The pool notifcation uses `useAlerts()` - do we use any other notification components? Or should I add one?

timjrobinson:
@garethfuller why do the trade page or homepage need the network included in the url? Pools do because there can be different pools of the same ID on different networks. Trade and homepage could use whatever is selected in the corner, or whatever network the users wallet is connect to. Uniswap doesn't include the network in the URL on the trade page.

garethfuller:
> @garethfuller thank you, updated. The pool notifcation uses `useAlerts()` - do we use any other notification components? Or should I add one?

@Matthews3301 `useNotifications()` is what Pon meant for you to use I think.

Matthews3301:
@garethfuller Updated to `useNotifications()`

src/plugins/router.ts
```diff
@@ -59,103 +65,108 @@ const routes: RouteRecordRaw[] = [
...
-    meta: { layout: 'ContentLayout' },
-  },
-  {
-    path: '/:pathMatch(.*)*',
-    name: 'not-found',
-    redirect: '/',
-  },
-  {
-    path: '/vebal',
+    path: '/:networkSlug/vebal',
```
timjrobinson, line 126: This page is only on mainnet, same with `get-vebal` and `unlock`

garethfuller:
@Matthews3301 https://www.loom.com/share/64a14377a1e246199e2ac7f25e5b69e4

src/composables/useNetwork.ts
```diff
@@ -1,18 +1,30 @@
...
-  process.env.VUE_APP_NETWORK != null
-    ? (Number(process.env.VUE_APP_NETWORK) as Network)
-    : Network.MAINNET;
+const windowAvailable = typeof window !== 'undefined';
+const localStorageNetworkId: Network | undefined =
+  windowAvailable && localStorage.getItem('networkId')
+    ? (Number(localStorage.getItem('networkId')) as Network)
+    : undefined;
+
+const NETWORK_ID =
```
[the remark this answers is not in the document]
Matthews3301: Are you sure? This scenario happens in pretty much only case - when user opens `/` route without a network selected in url. Should we not show Ethereum by default to highlight our main pools like like 80-20 BAT-ETH and Aave boosted?

src/plugins/router.ts
Matthews3301: Yes, I was thinking about this - I see what you mean. However, would it not be easier for users to understand this via the existing modal which explains that it's eth only? Might be confusing that all other urls start with the network name. 

Also, this is more future-proof if there is veBal on other networks.

src/composables/useNetwork.ts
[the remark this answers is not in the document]
garethfuller: imo I don't think we should be detecting the user's wallet's network and auto redirecting. I think that could be especially confusing if they follow a link to a pool on a specific network, for example: A user who's wallet is connected to mainnet clicks on an external URL linking to a polygon pool, on landing on the page we detect their network is mainnet and throw them back to the home page.

src/plugins/router.ts
garethfuller: Yeah, I think it's fine to maintain the network in the URL for the verbal pages for the same points that Matthew made.

src/components/navs/AppNav/AppNavNetworkSelect.vue
```diff
@@ -1,66 +1,106 @@
...
+  )
+    return `/#/${network.networkSlug}?poolNetworkAlert=${networkConfig.name}`;
+
+  const currentRoute = router.currentRoute.value;
+  return router.resolve({
+    name: currentRoute.name ?? 'home',
+    params: { ...currentRoute.params, networkSlug: network.networkSlug },
+    query: currentRoute.query,
+  }).href;
 }
```
garethfuller, line 99: I imagine this function is actually no longer relevant? It was only used before to get the app domain for a given network, not the path.

src/composables/useNetwork.ts
```diff
@@ -1,18 +1,30 @@
...
+  localStorageNetworkId ||
+  Number(process.env.VUE_APP_NETWORK) ||
```
garethfuller, line 18: VUE_APP_NETWORK should be deprecated with these changes. It shouldn't be required anymore.

src/composables/useNetwork.ts
```diff
@@ -26,9 +38,6 @@ export const isTestnet = computed(() => isKovan.value || isGoerli.value);
 /**
  * METHODS
  */
-export function setNetworkId(id: Network) {
-  networkId.value = id;
-}
```
garethfuller, line 31: Should we not keep the possibilty to change the networkId programmatically? Ideally we have a single source of truth for what the current app network is, e.g. `networkId`. When a user lands on a route it's set to the relevant network, when a user switches network with the network switcher it's set to the relevant network. Then eventually we can transition everything to be reactive so we don't need refreshes.

src/providers/token-lists.provider.ts
```diff
@@ -34,6 +35,7 @@ export interface TokenListsProviderResponse {
 
 /** SETUP */
 const { uris } = tokenListService;
+const { NETWORK_ID } = useNetwork();
```
garethfuller, line 38: We should always be accessing a reactive property for the network. e.g. networkID.

src/providers/token-lists.provider.ts
```diff
@@ -50,7 +52,7 @@ export default {
 
     let allTokenLists = {};
     try {
-      allTokenLists = require<TokenListMap>('/public/data/tokenlists.json');
+      allTokenLists = require<TokenListMap>(`/public/data/tokenlists/tokens-${NETWORK_ID}.json`);
```
garethfuller, line 55: This will be why some token icons are not displaying. `allTokenLists` now needs to be a reactive property that updates when `networkId.value` updates.

garethfuller (changes requested):
Good first pass.

src/components/navs/AppNav/AppNavNetworkSelect.vue
Matthews3301: It performs a different role now, for pool routes it redirects to "home" and for all other routes it updates the current url with new network.

src/providers/token-lists.provider.ts
Matthews3301: The change in this line should fix the token icon problem. Does it still persist?

src/composables/useNetwork.ts
Matthews3301: I think we still need it for unit tests... Not sure if possible to run some of the tests which require kovan without it.

src/components/navs/AppNav/AppNavNetworkSelect.vue
garethfuller: Let's rename it then. Maybe something like `transformUrl` or `handleNetworkChange`?

src/composables/useNetwork.ts
Matthews3301: Do you mean should we keep this in preparation for future changes? Currently, there's no point for networkId to be reactive as network is always constant after an app is loaded.

garethfuller: Yeah I guess that's what I was getting at, but actually, I think it's fine to remove it for now. We should focus on getting this working with a page reload first.

garethfuller:
@Matthews3301 next week, please click around the app and perform actions whilst watching the console. Any errors that come up lets look into them and make sure they're not caused by network switching.

garethfuller:
Also, if the environment variable `VUE_APP_ENV` / `APP_ENV` when fetched from configService is either 'development' or 'staging' lets add Goerli to the network switcher.

garethfuller:
@Matthews3301 Load the app, then change network, then go to trade and try and perform a trade, it throws an error because of an incorrect network somewhere.

Matthews3301:
@garethfuller By any chance, is this not this error (`underlying network changed`): https://docs.google.com/document/d/1l5cTqDaI7pt1QdNU7qL4j_cQtRD0HosFH6qa_suRabU/edit

This seems to be unrelated to this merge request but is a problem we should fix. I'll look into this more.

Matthews3301:
@pkattera @garethfuller I added network change watcher so that if user has already connected their wallet and changed then changed their network through MetaMask, the page will be redirected to that new network. Ready for review.

Matthews3301:
@garethfuller There is now a redirect which checks for a subdomain and redirects to the new format if necessary.

src/composables/useNetwork.ts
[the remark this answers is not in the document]
timjrobinson: @garethfuller The users wallet would only come after the URL in priority. Priority should be URL flag -> local storage network -> users wallet currently selected network -> mainnet. 

@Matthews3301 A lot of people only use Polygon (or L2's) avoid mainnet because it's too expensive for them. It would make more sense for them to just be shown the Polygon homepage as their default landing page. 

It's also useful for repeat Balancer users - so that they don't have to continually select their favorite network every time they open Balancer. Though maybe this is covered by localStorageNetworkId.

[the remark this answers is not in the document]
garethfuller: I still don't agree that we should auto-redirect based on a user's wallet (except in the case that we talked about Matthew, where a user is already connected to the correct network and then switches).

I do agree that we should store their last used network in localStorage and use that, but I'm not sure how. One issue for example, similar to what I described above: what if the user last used polygon and then left Balancer, they then click on a link to a mainnet pool page or to the arbitrum pools list from an external app, do we redirect them? We shouldn't. So in what cases would we use the localStorage variable to redirect on landing?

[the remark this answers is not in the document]
timjrobinson: I think the localStorage network or using users wallet network only makes sense on the main pools list or trade page without any network in the URL. Which seems to be the logic here, having the network in the URL overrides all.

src/composables/useNavigationGuards.ts
```diff
@@ -13,6 +17,53 @@ export default function useNavigationGuards() {
...
+      const networkFromUrl = networkFromSlug(to.params.networkSlug?.toString());
+      localStorage.setItem(
+        'networkId',
+        (networkFromUrl || networkFromSubdomain).toString()
+      );
+      window.location.href = networkFromUrl
+        ? `${appUrl()}${to.fullPath}`
+        : `${appUrl()}/${config[networkFromUrl || networkFromSubdomain].slug}${
+            to.fullPath
+          }`;
```
garethfuller, line 39: This should be a testable function that returns a URL. Please write some tests so we can be sure a variety of different input URLs will be redirected to the right place.

src/composables/useNavigationGuards.ts
```diff
@@ -13,6 +17,53 @@ export default function useNavigationGuards() {
...
+        } else if (localStorageNetwork === networkFromUrl) {
+          // if on the correct network -> next()
+          next();
+        } else {
+          // if on different network -> update localstorage and reload
+          document.write('');
+          localStorage.setItem('networkId', networkFromUrl.toString());
+          window.location.href = `/#${to.fullPath}`;
+          router.go(0);
+        }
```
garethfuller, line 60: Again, I think this should be extracted into a function and write tests to be sure it does as intended based on some test URLs.

garethfuller (changes requested):

src/composables/useNavigationGuards.ts
Matthews3301: See above.

src/composables/useNavigationGuards.ts
Matthews3301: 👍 Refactored the code to split into separate functions and added unit tests. 

Separately, I don't think file `/src/lib/config/test.json` is used for anything - can I delete it?

src/composables/useNavigationGuards.ts
```diff
@@ -1,18 +1,98 @@
...
+) {
+  const networkFromUrl = networkFromSlug(networkSlug ?? '');
+  localStorage.setItem(
+    'networkId',
+    (networkFromUrl || networkFromSubdomain).toString()
+  );
+  return networkFromUrl
+    ? `${appUrl()}${fullPath}`
+    : `${appUrl()}/${config[networkFromSubdomain].slug}${fullPath}`;
+}
```
garethfuller, line 37: Again, not super clear from the code what's going on here. I feel like the inputs here are over-complicating things. We should have a single function which takes the current URL and outputs the redirect URL if required

```suggestion
/**
 * Given current URL returns redirect URL, if no redirect required returns input URL.
 *
 * @param {string} url - Current URL.
 * @returns {string} URL to redirect too or current URL if no redirect required.
 */
export function getSubdomainNetworkRedirect(url: string): string {
  const subdomainNetwork = getSubdomainNetwork(url);
  if (subdomainNetwork) {
    // Construct redirect URL
  } else {
    return url;
  }
}
```
With a function like this we can more easily and clearly test; given input URL I expect this output URL.

Matthews3301: Thanks. How would we get fullPath and networkSlug from a url? By using regex?

I changed the function to this format but it requires more parameters.

src/composables/useNetwork.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+      },
+      {
+        url: 'https://goerli.balancer.fi/#/pool/0x5a6a8cffb4347ff7fc484bf5f0f8a2e234d34255000200000000000000000275',
+        result:
+          'https://localhost:8080/#/goerli/pool/0x5a6a8cffb4347ff7fc484bf5f0f8a2e234d34255000200000000000000000275',
+      },
+      {
+        url: 'https://ethereum.balancer.fi/#/portfolio',
+        result: 'https://localhost:8080/#/ethereum/portfolio',
+      },
```
garethfuller, line 41: The ethereum subdomain doesn't exist. Mainnet is on app.balancer.fi. I'd like to see a test case that insures `app.balancer.fi/#/some/path` basically doesn't redirect. Or redirects to `app.balancer.fi/#/ethereum/some/path` if it needs to.

I also noticed that `app.balancer.fi/#/trade`, for example, doesn't work. It doesn't redirect and it just loads the home page. That means any external link to out current mainnet trade page will be broken. We need to handle stuff like this.

Note: I've referenced `app.balancer.fi` as the domain here instead of the preview URL to keep it simple.

garethfuller (changes requested):

src/composables/useNetwork.ts
```diff
@@ -47,38 +63,94 @@ export function networkFor(key: string | number): Network {
...
+ * Get subdomain, excluding 'beta'
+ *
+ * @param {string} url - Host url - e.g. "polygon.balancer.fi/".
+ * @returns {string} Subdomain.
+ */
+export function getSubdomain(url: string) {
+  const subdomain = url.split('.')[0];
+  if (subdomain === 'beta') {
+    return url.split('.')[1];
+  } else {
```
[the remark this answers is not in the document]
Matthews3301: Updated

src/composables/useNavigationGuards.ts
```diff
@@ -1,18 +1,90 @@
...
+      if (to.redirectedFrom || !nonNetworkedRoutes.includes(to.fullPath)) {
+        localStorage.setItem('networkId', networkId.value.toString());
+        router.push({
+          path: `/${networkSlug}${to.redirectedFrom?.fullPath ?? to.fullPath}`,
+        });
+      } else {
+        next();
+      }
+    }
+  });
```
garethfuller, line 86: @Matthews3301 Could you explain specifically what redirect cases this is handling? Just in the comments below. I'm trying to understand it but it's not super clear.

Matthews3301: 1. E.g. `xxx.fi/polygon/trade` -> if current network != polygon, reload page with right network.
2. E.g. `xxx.fi/trade`(previous format) -> unless it's one of nonNetworkedRoutes, redirect to `xxx.fi/<current_network>/trade`.

PR #2135, opened 2022-08-12
Single interface for constructing transactions

garethfuller:
# Description

Problem:
- We had several `sendTransaction` and `callStatic` methods spread out across the codebase.
- Gas settings should be the same for all transactions
- Up until now all transactions have been contract calls
- With pool migration and SDK development we need to start submitting raw transactions given pre-constructed data.

Solution:
This PR adds a [TransactionBuilder](https://github.com/balancer-labs/frontend-v2/blob/tx-builder/src/services/web3/transactions/transaction.builder.ts) class that should be used everywhere we submit transactions. This builder has two concerns:
- [Raw](https://github.com/balancer-labs/frontend-v2/blob/tx-builder/src/services/web3/transactions/concerns/raw.concern.ts) - for submitting raw transactions given a `to` and `data` param, typically returned from an SDK function.
- [Contract](https://github.com/balancer-labs/frontend-v2/blob/tx-builder/src/services/web3/transactions/concerns/contract.concern.ts) - for submitting contract call transactions.

The TransactionBuilder ensures that all transactions are constructed and handled in the same way including a retry with legacy gas settings when a tx fails because of EIP1559 settings.

TransactionBuilder interface should be used like:
```ts
const txBuilder = new TransactionBuilder(signer)

// For contract calls.
const tx = await txBuilder.contract.sendTransaction({
    contractAddress,
    abi,
    action,
    params,
});

// For raw transactions
const tx = await txBuilder.raw.sendTransaction({ to, data });
```

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test all transaction types - Swaps, investments, withdrawals, stake, unstake, vote, lock.

## Visual context
n/a

## Checklist:


src/services/gas-price/gas-price.service.ts
```diff
@@ -32,42 +33,85 @@ export default class GasPriceService {
...
-          gasSettings.gasPrice = gasPrice.price;
-        }
+    );
+  }
+
+  private async setGasPriceSettings(
+    gasSettings: GasSettings,
+    forceLegacyTxType: boolean
+  ): Promise<GasSettings> {
+    const gasPrice = await this.getLatest();
```
alter-eggo, line 103: maybe name here `getLatestGasPrice`?

alter-eggo (commented):
lgtm

src/services/gas-price/gas-price.service.ts
garethfuller: Good shout, done.

alter-eggo (approved):

src/composables/pools/useTokenApprovals.ts
```diff
@@ -90,13 +90,13 @@ export default function useTokenApprovals(
...
-      const tx = await sendTransaction(
-        getProvider(),
-        address,
-        ERC20ABI,
-        'approve',
-        [spender, amount]
-      );
+      const txBuilder = new TransactionBuilder(getSigner());
+      const tx = await txBuilder.contract.sendTransaction({
+        contractAddress: address,
```
timjrobinson, line 95: This API is nice, I like specifying each argument by name much more than the list.

src/services/web3/web3.service.ts
```diff
@@ -42,6 +35,11 @@ export default class Web3Service {
 
   public setUserProvider(provider: ComputedRef<Web3Provider>) {
     this.userProvider = provider;
+    this.setTxBuilder(provider.value.getSigner());
+  }
+
+  public setTxBuilder(signer: JsonRpcSigner) {
```
timjrobinson, line 41: I don't like the name of this function, as it's inconsistent with setUserProvider above, I would assume the parameter is the TransactionBuilder.

src/services/web3/transactions/concerns/transaction.concern.ts
```diff
@@ -0,0 +1,30 @@
...
+  public shouldRetryAsLegacy(error: WalletError): boolean {
+    return (
+      error.code === RPC_INVALID_PARAMS_ERROR_CODE &&
+      EIP1559_UNSUPPORTED_REGEX.test(error.message)
+    );
+  }
+
+  public shouldLogFailure(error: WalletError): boolean {
+    return (
+      error.code === ErrorCode.UNPREDICTABLE_GAS_LIMIT &&
```
timjrobinson, line 26: We could be accidentally hiding legitimate errors here. I'd prefer to have a ignore list of error codes so that we know about new ones.

PR #2144, opened 2022-08-16
feat: New env variable - VUE_APP_WALLET_SCREENING

Matthews3301:
# Description

TRM sanction check will only be performed if there is `VUE_APP_WALLET_SCREENING` variable set to `true` in the environment.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

Set `VUE_APP_WALLET_SCREENING` to true
Set `VUE_APP_WALLET_SCREENING` to false
Dont set `VUE_APP_WALLET_SCREENING`

## Visual context

N/A

## Checklist:


src/services/web3/web3.plugin.ts
```diff
@@ -67,6 +67,11 @@ type PluginState = {
 };
 
 async function isSanctionedAddress(address: string): Promise<boolean | null> {
+  if (
+    !process.env.VUE_APP_WALLET_SCREENING ||
+    process.env.VUE_APP_WALLET_SCREENING !== 'true'
+  )
```
garethfuller, line 73: Could you move add this ENV var to config.service.ts `env` attribute and resolve the variable to a boolean type as expected. So in the config service something like:
```ts
WALLET_SCREENING: (process.env.VUE_APP_WALLET_SCREENING && process.env.VUE_APP_WALLET_SCREENING === 'true') || false
```
Then above you should just be able to do:
```suggestion
  if (configService.env.WALLET_SCREENING)
```

Matthews3301: Updated

src/services/web3/web3.plugin.ts
```diff
@@ -67,15 +68,19 @@ type PluginState = {
...
+        {
+          address: address.toLowerCase(),
+        },
+      ]);
+      const isSanctioned = response.data[0].isSanctioned;
+      return isSanctioned;
+    } catch {
+      return null;
+    }
+  } else {
```
garethfuller, line 83: Last question, unrelated to your work, why is the return type `Promise<boolean | null>` here? Could we not default to `false` if it fails? Is there any indication from the code why that decision was made?

Matthews3301: It's because it's an `async` function and therefore it returns a promise which must be resolved with `await`. It's an `async` function because it makes `axios.post` call.

garethfuller: Yeah but that doesn't stop it returning false instead of null. It could be like this:
```ts
async function isSanctionedAddress(address: string): Promise<boolean> {
  if (configService.env.WALLET_SCREENING) {
    try {
      const response = await axios.post(SANCTIONS_ENDPOINT, [
        {
          address: address.toLowerCase(),
        },
      ]);
      const isSanctioned = response.data[0].isSanctioned;
      return isSanctioned;
    } catch {
      return false;
    }
  } else {
    return false;
  }
}
```

garethfuller: I'm just wondering why it was returning null in the first place.

Matthews3301: Oh, sorry, I misunderstood. Perhaps the response `isSanctioned` is not always boolean and can be empty?

garethfuller: Nonetheless, can we fix this function so that it always returns a boolean please.

Matthews3301: Updated 🙂

garethfuller (approved):

src/services/config/config.service.ts
```diff
@@ -37,6 +38,10 @@ export default class ConfigService {
         'MISSING_KEY',
       ENABLE_STABLE_POOLS: process.env.VUE_APP_ENABLE_STABLE_POOLS === 'true',
       PORTIS_DAPP_ID: process.env.VUE_APP_PORTIS_DAPP_ID || 'MISSING_KEY',
+      WALLET_SCREENING:
+        (process.env.VUE_APP_WALLET_SCREENING &&
```
timjrobinson, line 42: This can just be `process.env.VUE_APP_WALLET_SCREENING === 'true'` as if `VUE_APP_WALLET_SCREENING` is not set it will be undefined and thus be false

src/services/web3/web3.plugin.ts
```diff
@@ -66,17 +67,20 @@ type PluginState = {
...
-      {
-        address: address.toLowerCase(),
-      },
-    ]);
-    const isSanctioned = response.data[0].isSanctioned;
-    return isSanctioned;
-  } catch {
-    return null;
+async function isSanctionedAddress(address: string): Promise<boolean> {
+  if (configService.env.WALLET_SCREENING) {
```
timjrobinson, line 71: Try and avoid else statements as they're almost never needed. This can be replaced with `if (!configService.env.WALLET_SCREENING) return false;`

timjrobinson (changes requested):

src/services/config/config.service.ts
Matthews3301: [👍](https://emojipedia.org/thumbs-up/)

src/services/web3/web3.plugin.ts
Matthews3301: It was specifically requested to be like this by @garethfuller

timjrobinson: Yea he was just saying to return a boolean instead of null. I'm saying we can clarify the code by removing the else portion and just fail quickly, as it makes the code neater.

Matthews3301: 👍

timjrobinson (approved):
LGTM!

PR #2146, opened 2022-08-17
Hotfix: Use an interval for new blocks instead of constantly listening to the websocket

timjrobinson:
# Description

Currently most of our alchemy bill is from listening for new blocks through the websocket connection. The only thing we use from these new blocks is the block number. 

This PR switches this to only listen for one block, then set an interval of the chains cadence to update the current block number, which should lower our costs while providing essentially the same service to the rest of the app. 

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- Test the app on Polygon and Arbitrum and ensure all functionality is working as expected.
- Test trading and ensure SOR is pulling in new data and updating prices on an interval. 
- After merging, ensure our Alchemy usage has dropped. 

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer-labs/frontend-v2",
-  "version": "1.67.0",
+  "version": "1.67.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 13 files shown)


garethfuller (approved):
Nice fix!

PR #2158, opened 2022-08-19
Fetch a more accurate time travel block for fees/volume

timjrobinson:
# Description

Currently to calculate fees/volume we find a block 24 hours in the past by doing (secondsInDay / 13). This isn't very accurate and is currently off by about 300 blocks from the actual block 24 hours ago. 

This PR uses a subgraph request to find the exactly block number from 24 hours ago instead of doing this rough calculation. This is what the SDK is also using in the [APRs Update](https://github.com/balancer-labs/balancer-sdk/pull/98) so it will be easier to compare Prod APR/fees/volume numbers to that PR. 

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- View pools page and individual pools, they should show volume / fees correctly. 

## Checklist:

### DIFF
--- a/src/composables/useSnapshots.ts
+++ b/src/composables/useSnapshots.ts
@@ -1,6 +1,5 @@
-import { configService } from '@/services/config/config.service';
 import { twentyFourHoursInSecs } from './useTime';
-import { BlockNumberResponse } from '@/types';
+import { blockService } from '@/services/block/block.service';
 
 export type TimeTravelPeriod = '24h';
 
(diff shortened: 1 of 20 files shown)


src/composables/useSnapshots.ts
```diff
@@ -1,38 +1,45 @@
...
   }
 }
+
+const query = (timestamp: string) => `{
+  blocks(first: 1, orderBy: number, orderDirection: asc, where: { timestamp_gt: ${timestamp} }) {
+    number
+  }
+}`;
+
+const fetchBlockByTime = async (timestamp: string): Promise<number> => {
```
sssmi, line 26: Could `useGraphQuery` composable be used here to execute the query instead of `fetch`?

timjrobinson: Oh yea true, I just copied this from the SDK. Will fix.

src/types/index.ts
```diff
@@ -67,3 +67,13 @@ export type BaseContent = {
...
+
+export interface BlockNumberResponse {
+  data: {
+    blocks: [
+      {
+        number: string;
+      }
+    ];
+  };
+}
```
garethfuller, line 79: Just define this in the useSnapshots composable so it's in the same place.

garethfuller (approved):

Issue #1989, opened 2022-06-15
Remove 'load more' bottom table row button when less than 10 results


pkattera:
**Describe the bug**
The table row containing the 'load more' button should be removed when there are less than 10 results in paginated tables, like the 'Investment pools' table. 

**To Reproduce**
1. Go to the home page
2. Filter by tokens / select a few popular tokens until there are less than 10 results
3. See the load more table row still visible.

**Expected behavior**
The table row should be completely removed.

**Screenshots**

See load more table row still present with less than 10 results.

![load-more](https://user-images.githubusercontent.com/1121139/173782051-38bc330b-389a-4076-8c63-3a27ecbe699c.png)

vishesh0123:
can I work on this issue..

garethfuller:
@vishesh0123 yeah sure!

nayane95:
May I also try to fix this?

PR #2173, opened 2022-08-24
Hotfix: Fix load more not appearing for trade / investment tables

timjrobinson:
# Description

@zekraken-bot reported that since the last weekly release the investments / trades list on a single pool page is no longer showing the "load more" button. This is because of the change in #2156 that made all BalTable's not show load more with <10 results. 

I've moved the logic to the main pools page only for now, to solve the original issue (#1989) while not changing the logic of BalTable. 

This does have an edge case if there are exactly 10 pools that match a token, the load more button will show but not do anything. I'll be fixing this in #2093 when we rework pool fetching. I'll make this variable be based on if the vue-query reports there is more to load. 

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- On the homepage, ensure load more works
- When filtering by a token on the homepage, load more should only show if there are >= 10 pools with that token 
- On an individual pool page the investments and swaps tables should have load more buttons. 

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer-labs/frontend-v2",
-  "version": "1.68.1",
+  "version": "1.68.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@balancer-labs/frontend-v2",
-      "version": "1.68.1",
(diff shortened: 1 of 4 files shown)


sssmi (approved):
Works fine now 👍

PR #2188, opened 2022-08-29
Safe friendly pool creation recovery

TomAFrench:
# Description

The pool creation form recovery mechanism relies on decoding the transaction data to recover the arguments sent to the pool factory (and so the config of the pool). This works from EOAs but if the pool is deployed using a contract wallet (such as a Safe) then this will fail as the call to the factory is wrapped in a call to the Safe. Instead we should read the pool's address through the emitted event and query the pool for it's config.

I've improved the `retrievePoolIdAndAddress` function to a method which uses the factory's interface to decode the event logs so we don't have to look for a hardcoded topic hash anymore. There was also a type issue in `PoolSeedToken` which I fixed.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

Connect to the polygon app and attempt to recover the pool creation form using tx hash [0xd206199ac88b0bcaf46efc7465cec6a3e6e9b3886c9fcf139061f87b0b4e5ee7](https://polygonscan.com/tx/0xd206199ac88b0bcaf46efc7465cec6a3e6e9b3886c9fcf139061f87b0b4e5ee7). This is a deployment of a pool through a gnosis safe.

Expect that the pool creation form cannot recover on the production site but can on this branch.

## Visual context

N/A

## Checklist:

### DIFF
--- a/src/composables/pools/usePoolCreation.spec.ts
+++ b/src/composables/pools/usePoolCreation.spec.ts
@@ -24,28 +24,28 @@ describe('usePoolCreation', () => {
       tokenAddress: '0x9f8F72aA9304c8B593d555F12eF6589cC3A579A2',
       weight: 70,
       isLocked: false,
-      id: 0,
+      id: '0',
       amount: '0',
     };
     tokens.WETH = {
       tokenAddress: '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2',
       weight: 20,
       isLocked: false,
-      id: 1,
+      id: '1',
       amount: '0',
     };
     tokens.USDT = {
       tokenAddress: '0xdac17f958d2ee523a2206206994597c13d831ec7',
       weight: 10,
       isLocked: false,
-      id: 2,
+      id: '2',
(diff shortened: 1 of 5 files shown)


garethfuller:
Hey @TomAFrench could you give more context around the `PoolSeedToken` issue? I think I might change it back because it breaks the implementation here: https://github.com/balancer-labs/frontend-v2/blob/safe-friendly-pool-creation-recovery/src/components/cards/CreatePool/ChooseWeights.vue#L246 where `uniqueId` returns a string.

TomAFrench:
I was running into issues [here](https://github.com/balancer-labs/frontend-v2/pull/2188/files#diff-a2465e98876ce6d939980d0bcb35fbea4a973d519dd1506b6bb7c6b7821eca9dR565) where we were trying to set a number as the id. I've reverted my changes and just explicitly converted this number to a string.

garethfuller (approved):
Thanks again for fixing this!

PR #2206, opened 2022-09-06
fix: Make sure SOR is used as soon as pools have been loaded

Matthews3301:
# Description

On page load, if I enter a value before the SOR loads I get a not enough liquidity error. It should refetch the swap. This is because SOR needs to be loaded to calculate trade amounts and if both are not present, not enough liquidity is assumed.

## Type of change

## How should this be tested?

Load Trade page and enter amount before SOR has been loaded. If there is liquidity, it should allow you to preview/trade.

## Visual context

N/A

## Checklist:


garethfuller (approved):
Nice catch, weird that this was set to false. 🤔

PR #2319, opened 2022-10-24
Deep pool composition view

agualis:
# Description

Refactor old `PoolBalanceCard` to new `PoolCompositionCard` (after merging old `BoostedPool` and `DefaultPool` into one single component).

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist:


garethfuller:
@agualis I've merged `develop` into `feature/generalized-boosted-pools` now so it's up to date. Maybe if you merge in `feature/generalized-boosted-pools` here it will remove the unnecessary diffs.

agualis:
> @agualis I've merged `develop` into `feature/generalized-boosted-pools` now so it's up to date. Maybe if you merge in `feature/generalized-boosted-pools` here it will remove the unnecessary diffs.

@garethfuller Now it looks ok again and checks are green ✅ We still have the commit squash problem but I will fix it later.

garethfuller (changes requested):

src/components/contextual/pages/pool/PoolCompositionCard/PoolCompositionCard.vue
```diff
@@ -0,0 +1,160 @@
...
+const { upToLargeBreakpoint } = useBreakpoints();
+const { explorerLinks } = useWeb3();
+const { fNum2 } = useNumbers();
+const { priceFor } = useTokens();
+
+/**
+ * METHODS
+ */
+function getTokenShare(address: string) {
+  const linearPools = pool.value.onchain?.linearPools;
```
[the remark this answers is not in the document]
garethfuller: I'm not 100% sure where to get the correct data for this yet either btw. We need to work through it. But we've included the balance with the token attribute in the token tree here: https://github.com/balancer-labs/frontend-v2/blob/develop/src/services/balancer/subgraph/entities/pools/query.ts#L17 can we not use that balance?

src/components/contextual/pages/pool/PoolCompositionCard/PoolCompositionCard.vue
```diff
@@ -0,0 +1,160 @@
...
+
+function hasNestedTokens(address: string) {
+  return getUnderlyingTokens(pool.value, address).length;
+}
+
+function getRootTokenStyle(address) {
+  // When there are nested tokens, the root token will have full width (just breakdown)
+  if (hasNestedTokens(address)) return 'py-4';
+  // When NO nested tokens, the root token will have 3 cols (breakdown + balance + value)
+  if (isWeighted) return 'grid grid-cols-4 py-4';
```
garethfuller, line 64: There are many other pool types that don't have nesting. This should be checking if it's a nested pool rather than for a specific pool type, e.g.
```suggestion
  if (!isDeep(pool)) return 'grid grid-cols-4 py-4';
```

src/components/contextual/pages/pool/PoolCompositionCard/nested-token-helpers.ts
```diff
@@ -0,0 +1,14 @@
...
+}
+
+export function getUnderlyingTokens(pool: Pool, address: string) {
+  const token = findTokenByAddress(pool, address);
+
+  const underlyingTokens = token?.token.pool?.tokens || [];
+  return underlyingTokens.filter(
+    token => !pool.tokensList.includes(token.address)
+  );
+}
```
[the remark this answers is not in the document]
agualis: > 1. This isn't a great location for these functions, they are pool helpers, they should probably go in usePool.ts

+1 I just created this file to start this conversation. My only concern about including this helpers inside `usePool.ts` is that every file starting with `use` prefix should be a composable and these helpers are not really composables (having Vue dependencies like ref or reactive) but pure TS functions. Maybe that's not a big deal but, for example, when testing a `useSomething.ts` file I would expect to be using `vue-composable-tester` while testing pure functions I would not. 

Anyways, maybe it is not such a strong argument, so I can follow your suggestion and keep on thinking about other trade-offs.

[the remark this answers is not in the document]
garethfuller: A composable can include pure functions which are part of the scope of that composable. These functions themselves are not composables but fall under the scope of using a pool. We could of course separate out pure functions from composable files into a poolHelpers.ts or something. But we haven't established a pattern around that yet.

src/components/contextual/pages/pool/PoolCompositionCard/PoolCompositionCard.vue
agualis: I don't understand the suggestion. I was considering 3 cases: 
1. hasNestedTokens() like **Balancer Boosted Aave USD**
nestedPool
2. isWeighted which I understood (probably wrongly) as pools which have weight atribute like **Oracle Weighted Pool**:
weighted
3. The rest of the pools (no nested, no weighted) like this **stable pool**:
Screenshot 2022-11-04 at 15 38 06

I still miss a lot of domain background 😅 Maybe we can have a quick call to clarify this?

garethfuller: ahh my bad, was skim-reading this part and didn't fully understand it. You are right here.

agualis: Maybe my new helper: 
```ts
function hasNestedTokens(address: string) {
  return getUnderlyingTokens(pool.value, address).length;
}
```
can be avoided and replaced by `isDeep`?

src/components/contextual/pages/pool/PoolCompositionCard/PoolCompositionCard.vue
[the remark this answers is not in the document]
agualis: Last change to address, I will need help with this one as I don't understand how to recreate the old "formula":
```ts
  return bnum(token?.balance || '0')
    .div(linearPools[address].totalSupply)
```

with the new `balances` that you pointed.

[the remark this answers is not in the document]
garethfuller: We should add `totalShares` to this part of the query: https://github.com/balancer-labs/frontend-v2/blob/develop/src/services/balancer/subgraph/entities/pools/query.ts#L24-L28

Where `totalShares` is effectively the same thing as `totalSupply`, then you will have to write some code to figure out which pool the token you're looking at belongs to and use the `totalShares` value in that calculation.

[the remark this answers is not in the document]
agualis: This is the way we are currently using shares prop (the one that passes linearPool totalSupply value):

https://github.com/balancer-labs/frontend-v2/blob/d2263dc668a81b09b6eaa69e48db6b2775478885/src/components/contextual/pages/pool/PoolBalancesCard/components/AssetRow.vue#L40

As you can see, we multiply by the shares which is a `percentage` as you can see in this screenshot where I'm rendering the shares prop: 
Screenshot 2022-11-07 at 12 24 00

src/composables/usePool.ts
```diff
@@ -357,6 +357,30 @@ export function fiatValueOf(pool: Pool, shares: string): string {
   return bnum(shares).times(bptPriceFor(pool)).toString();
 }
 
+export function findTokenByAddress(pool: Pool, address: string) {
+  return pool.tokens.find(token => isSameAddress(token.address, address));
+}
```
garethfuller, line 362: This doesn't really work. It only finds tokens in layer 1 of the token tree. You need to search through all nested layers. I added a function for this, in this PR: https://github.com/balancer-labs/frontend-v2/pull/2404/files#diff-c7a3eec563f69ad2832ed59c9eda57a3a98acab763483eaa6ef08e401bd8f605R356-R370

It will be merged to the feature base soon so you should be able to use it.

src/composables/usePool.ts
```diff
@@ -357,6 +357,30 @@ export function fiatValueOf(pool: Pool, shares: string): string {
...
+
+export function calculateTokenBPTShareByAddress(
+  pool: Pool,
+  address: string
+): string {
+  const token = findTokenByAddress(pool, address);
+  if (!token) return '0';
+  return bnum(token?.balance || '0')
+    .div(token.token.pool?.totalShares || 1)
```
garethfuller, line 380: I think this is incorrect, this is getting the `totalShares` if the token itself is a pool. You instead need to find the parent pool of the current token.
```
- Pool 1
-- Token 1
-- Token 2 (pool 2)
---- Token 3
---- Token 4
```
Let's say you want the bpt share of Token 2. Your calculation is returning `totalShars` of Pool 2, but what you want is the `totalShares` of Pool 1 because that is the parent pool of Token 2.

agualis: I changed this by passing `totalShares` as a prop in `TokenBreakdown` (recursive component)

garethfuller:
@agualis Could we keep padding and sizes similar to current production please?

This branch:
Screenshot 2022-11-21 at 11 32 44

Production:
Screenshot 2022-11-21 at 11 32 47

agualis:
Left-light is production, Right-dark is current branch: 

Screenshot 2022-11-21 at 15 55 40

Screenshot 2022-11-21 at 15 55 13

garethfuller (approved):

PR #2323, opened 2022-10-25
Initialize the pool tokens inputs as pool tokens with wallet balance

sssmi:
# Description

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:

### DIFF
--- a/src/components/cards/MyWallet/MyWallet.vue
+++ b/src/components/cards/MyWallet/MyWallet.vue
@@ -1,19 +1,18 @@
 <script setup lang="ts">
-import { take } from 'lodash';
 import { computed, toRef } from 'vue';
 import { useI18n } from 'vue-i18n';
 
 import useBreakpoints from '@/composables/useBreakpoints';
 import { isMainnet } from '@/composables/useNetwork';
 import useTokens from '@/composables/useTokens';
-import { isSameAddress, includesAddress } from '@/lib/utils';
 import { configService } from '@/services/config/config.service';
 import useWeb3 from '@/services/web3/useWeb3';
 import { Address } from '@/types';
-import { AnyPool, PoolToken } from '@/services/pool/types';
+import { AnyPool } from '@/services/pool/types';
 import MyWalletSubheader from './MyWalletSubheader.vue';
 import useNativeBalance from '@/composables/useNativeBalance';
 import { usePool } from '@/composables/usePool';
+import useMyWallet from '@/composables/useMyWallet';
 
 type Props = {
   excludedTokens?: string[];
@@ -29,17 +28,22 @@ const props = withDefaults(defineProps<Props>(), {
   includeNativeAsset: false,
 });
 
-const { appNetworkConfig, isWalletReady, startConnectWithInjectedProvider } =
-  useWeb3();
+const { isWalletReady, startConnectWithInjectedProvider } = useWeb3();
 const { upToLargeBreakpoint } = useBreakpoints();
-const {
-  balances,
-  dynamicDataLoading: isLoadingBalances,
-  nativeAsset,
-} = useTokens();
+const { dynamicDataLoading: isLoadingBalances } = useTokens();
 const networkName = configService.network.name;
 const { t } = useI18n();
-const { isWethPool, isDeepPool } = usePool(toRef(props, 'pool'));
+const { isDeepPool } = usePool(toRef(props, 'pool'));
+const {
+  tokensWithBalance,
+  poolTokensWithBalance,
+  poolTokensWithoutBalance,
+  notPoolTokensWithBalance,
+} = useMyWallet({
+  excludedTokens: props.excludedTokens,
+  pool: props.pool,
+  includeNativeAsset: props.includeNativeAsset,
+});
 
 const noNativeCurrencyMessage = computed(() => {
   return t('noNativeCurrency', [nativeCurrency, networkName]);
@@ -54,91 +58,6 @@ const noTokensMessage = computed(() => {
 });
 
 const { hasNativeBalance, nativeBalance, nativeCurrency } = useNativeBalance();
-
-function isExcludedToken(tokenAddress: Address) {
-  return props.excludedTokens.some(excludedAddress =>
-    isSameAddress(excludedAddress, tokenAddress)
-  );
-}
-
-const tokensWithBalance = computed(() => {
-  return take(
-    Object.keys(balances.value).filter(tokenAddress => {
-      const _includeNativeAsset = props.includeNativeAsset
-        ? true
-        : !isSameAddress(tokenAddress, appNetworkConfig.nativeAsset.address);
-      return (
-        Number(balances.value[tokenAddress]) > 0 &&
-        _includeNativeAsset &&
-        !isSameAddress(tokenAddress, appNetworkConfig.addresses.veBAL) &&
-        !isExcludedToken(tokenAddress)
-      );
-    }),
-    21
-  );
-});
-
-function deepPoolTokenAddressReducer(
-  acc: Set<string>,
-  poolToken: PoolToken
-): Set<string> {
-  // Exclude BPT node
-  if (isSameAddress(poolToken.address, props.pool.address)) {
-    return acc;
-  }
-  // If current node has children, recursively look through the nested token tree
-  if (poolToken.token.pool?.tokens?.length) {
-    const nestedTokenAddresses = poolToken.token.pool?.tokens.reduce<
-      Set<string>
-    >(deepPoolTokenAddressReducer, acc);
-    return nestedTokenAddresses;
-  }
-
-  // Add the final token address to set
-  const { address } = poolToken;
-  return acc.add(address);
-}
-
-function getDeepPoolTokenAddresses(pool: AnyPool): string[] {
-  const tokensSet = pool.tokens.reduce<Set<string>>(
-    deepPoolTokenAddressReducer,
-    new Set<string>()
-  );
-
-  return Array.from(tokensSet);
-}
-const poolTokenAddresses = computed((): string[] => {
-  if (isDeepPool.value) {
-    return getDeepPoolTokenAddresses(props.pool);
-  }
-  if (isWethPool.value) {
-    return [nativeAsset.address, ...props.pool.tokensList];
-  }
-  return props.pool.tokensList;
-});
-
-const poolTokensWithBalance = computed<string[]>(() => {
-  return (
-    poolTokenAddresses.value.filter(poolToken =>
-      includesAddress(tokensWithBalance.value, poolToken)
-    ) || []
-  );
-});
-const poolTokensWithoutBalance = computed<string[]>(() => {
-  return (
-    poolTokenAddresses.value.filter(
-      poolToken => !includesAddress(tokensWithBalance.value, poolToken)
-    ) || []
-  );
-});
-const notPoolTokensWithBalance = computed<string[]>(() => {
-  if (!poolTokenAddresses.value) return tokensWithBalance.value;
-  return (
-    tokensWithBalance.value.filter(
-      token => !includesAddress(poolTokenAddresses.value || [], token)
-    ) || []
-  );
-});
 
 const emit = defineEmits<{
   (e: 'click:asset', tokenAddress: Address): void;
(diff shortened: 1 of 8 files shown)


src/services/balancer/pools/joins/handlers/deep-pool-join.handler.ts
```diff
@@ -1,40 +1,105 @@
...
+    const bptOut = formatFixed(
+      this.lastGeneralisedJoinRes.expectedOut,
+      this.pool.value.onchain?.decimals || 18
+    );
+    const priceImpact: number = bnum(
+      formatFixed(
+        // @ts-ignore-next-line -- priceImpact is part of the response, but type is missing
+        this.lastGeneralisedJoinRes.priceImpact,
+        this.pool.value.onchain?.decimals || 18
+      )
```
garethfuller, line 96: If price impact is being returned in evm number format it will always be scaled with 18. You only use the decimals of the pool or token when you're dealing with amounts of the relevant token.

garethfuller: Make sure this is the case with Bruno, but I'm pretty sure you can just use 18 here.

src/components/forms/pool_actions/InvestForm/components/InvestPreviewModal/components/TokenAmounts.vue
```diff
@@ -73,14 +79,17 @@ function amountShare(address: string): string {
...
+        <div class="flex flex-col self-center mr-3">
           <div class="font-medium">
             <span class="font-numeric">
               {{ fNum2(token.amount, FNumFormats.token) }}
             </span>
             {{ tokenMap[token.address]?.symbol }}
           </div>
-          <div class="text-sm text-secondary font-numeric">
+          <div
+            v-if="Number(token.fiatAmount) > 0"
```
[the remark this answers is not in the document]
sssmi: This conditional just hides the fiat value text, if fiat value is 0.

src/composables/useMyWalletTokens.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    const tokensList = pool?.tokensList || [];
+    if (isWethPool.value) {
+      return [nativeAsset.address, ...tokensList];
+    }
+    return tokensList;
+  });
+
+  const poolTokensWithBalance = computed<string[]>(() => {
```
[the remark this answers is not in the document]
sssmi: No reason, `poolTokenAddresses` above is probably just refactor, so I kept the same return type

src/composables/useSignRelayerApproval.ts
```diff
@@ -0,0 +1,59 @@
...
+  const { t } = useI18n();
+
+  /**
+   * METHODS
+   */
+  async function signRelayerApproval(): Promise<void> {
+    const signer = getSigner();
+    const signerAddress = await signer.getAddress();
+    const signature = await Relayer.signRelayerApproval(
+      configService.network.addresses.batchRelayerV4,
```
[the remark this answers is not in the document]
sssmi: I agree

src/lib/config/goerli.json
```diff
@@ -47,7 +47,8 @@
     "wstETH": "0x6320cD32aA674d2898A68ec82e869385Fc5f7E2f",
     "lidoRelayer": "0xdc10e8f47B6858E32218E01db224795235872B19",
     "balancerHelpers": "0x5aDDCCa35b7A0D07C74063c48700C8590E87864E",
-    "batchRelayer": "0x2Fb8aD26faDC971464A17407e9213D1A22befc8a",
+    "batchRelayer": "0x7b9B6f094DC2Bd1c12024b0D9CC63d6993Be1888",
```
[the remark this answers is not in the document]
sssmi: Changes are against `feature/generalized-boosted-pools`. I had previously changed the `batchRelayer` to use Relayer v4 here, and this is just an undo.

`0x7b9B6f094DC2Bd1c12024b0D9CC63d6993Be1888` is the right address for Relayer v3

src/providers/local/join-pool.provider.ts
```diff
@@ -50,13 +55,24 @@ type Props = {
...
+  /**
+   * COMPOSABLES
+   */
+  const { getTokens, prices, injectTokens } = useTokens();
+  const { toFiat } = useNumbers();
+  const { slippageBsp } = useUserSettings();
+  const { getSigner } = useWeb3();
+  const { txState, txInProgress } = useTxState();
+  const relayerApproval = useRelayerApproval(Relayer.BATCH_V4);
+  const { relayerSignature, signRelayerAction } = useSignRelayerApproval();
```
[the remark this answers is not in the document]
sssmi: I think at some point I had to pass a state variable to one of the composables. It's not the case any more, so I'll change the order back.

src/providers/local/join-pool.provider.ts
```diff
@@ -70,23 +86,16 @@ const provider = (props: Props) => {
...
-  const { getTokens, prices, injectTokens } = useTokens();
-  const { toFiat } = useNumbers();
-  const { slippageBsp } = useUserSettings();
-  const { getSigner } = useWeb3();
-  const { txState, txInProgress } = useTxState();
-
   /**
    * COMPUTED
    */
+  const isDeepPool = computed<boolean>(() => isDeep(pool.value));
```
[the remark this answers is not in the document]
sssmi: I generally prefer to use type arguments instead of setting the callback function type. But it doesn't make much difference to me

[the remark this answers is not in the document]
garethfuller: I don't mind either way, but would be nice if we could keep it consistent across the code base. Maybe there is a linter rule?

src/providers/local/join-pool.provider.ts
```diff
@@ -173,17 +196,40 @@ const provider = (props: Props) => {
...
+    // If form is empty, clear the price impact and
+    // return early
+    if (!hasAmountsIn.value) {
+      priceImpact.value = 0;
+      return;
+    }
+    // Early return if relayer not yet approved
+    if (shouldSignRelayer.value) {
+      return;
+    }
```
garethfuller, line 221:
```suggestion
    if (shouldSignRelayer.value) return;
```

src/services/balancer/pools/joins/handlers/deep-pool-join.handler.ts
```diff
@@ -1,40 +1,105 @@
...
+        signerAddress,
+        wrapLeafTokens,
+        slippage,
+        relayerSignature
+      )
+      .catch(err => {
+        console.error(err);
+        throw new Error(err);
+      });
+    console.log({ lastGeneralisedJoinRes: this.lastGeneralisedJoinRes });
```
garethfuller, line 83: Please remember to remove logging.

src/services/balancer/pools/joins/handlers/deep-pool-join.handler.ts
```diff
@@ -1,40 +1,105 @@
...
+
+    if (!this.lastGeneralisedJoinRes) throw new Error('Not enough liquidity.');
```
garethfuller, line 85: This is a different error case than in the swap handler. If a swap returns a zero amount then it's usually because of liquidity, but not here. I'd change this to something like "Failed to fetch expected output.".

src/providers/local/join-pool.provider.ts
```diff
@@ -173,17 +196,38 @@ const provider = (props: Props) => {
...
    */
   async function queryJoin() {
+    // If form is empty, clear the price impact and
+    // return early
+    if (!hasAmountsIn.value) {
+      priceImpact.value = 0;
+      return;
+    }
+    // Early return if relayer not yet approved
+    if (shouldSignRelayer.value) return;
```
garethfuller, line 219: Actually, I don't think we should have this block here. I'm not sure where we should have it but this means for all pool types regardless of whether you need the relayer signature or not, it will block queries. I guess the block should happen based on the join handler type.

Maybe we should have a static property on the JoinHandler interface called `isRelayerSigRequired`. Then for all pool handlers, you have to specify whether it is or not. Then this conditional would be `joinPoolService.isRelayerSigRequired && !hasSignedRelayer.value`

src/services/balancer/pools/joins/handlers/deep-pool-join.handler.ts
```diff
@@ -1,40 +1,105 @@
...
-  }
+    relayerSignature,
+  }: JoinParams): Promise<QueryOutput> {
+    const parsedAmountsIn: string[] = amountsIn.map(({ address, value }) => {
+      // Get the address in right casing style
+      const realAddress = getAddress(address);
+      const token = tokensIn[realAddress];
+      const parsedAmount = parseFixed(
+        value || '0',
+        token?.decimals ?? 18
```
garethfuller, line 58: I don't think we should assume decimals for joins/exits, it's critical for the correct maths. If we don't have the token decimals we should throw an error.

src/services/balancer/pools/joins/handlers/deep-pool-join.handler.ts
```diff
@@ -1,40 +1,105 @@
...
+        console.error(err);
+        throw new Error(err);
+      });
+
+    if (!this.lastGeneralisedJoinRes) {
+      throw new Error('Failed to fetch expected output.');
+    }
+    const bptOut = formatFixed(
+      this.lastGeneralisedJoinRes.expectedOut,
+      this.pool.value.onchain?.decimals || 18
```
garethfuller, line 89: Pool decimals is the only thing we can assume because I believe all balancer pools are 18 decimals.

src/services/balancer/pools/joins/handlers/deep-pool-join.handler.ts
```diff
@@ -1,40 +1,105 @@
...
+      const realAddress = getAddress(address);
+      const token = tokensIn[realAddress];
+      const parsedAmount = parseFixed(
+        value || '0',
+        token?.decimals ?? 18
+      ).toString();
+      return parsedAmount;
+    });
+
+    const tokenAddresses: string[] = amountsIn.map(({ address }) => address);
```
garethfuller, line 63: Are you 100% sure that `.map` results in the same order every time? If not `parsedAmountsIn` and `tokenAddresses` could end up not lining up and causing issues.

sssmi:
> @sssmi Clicked max on USDT and this happened:
> 
> ```
> TypeError: Cannot convert undefined or null to object
>     at Function.keys (<anonymous>)
>     at eval (index.umd.js?3041:474:1)
>     at Array.some (<anonymous>)
>     at js.encodeBalanceAndAllowanceOverrides (index.umd.js?3041:474:1)
>     at async js.simulateMulticall (index.umd.js?3041:474:1)
>     at async Hs.amountsOutByJoinPath (index.umd.js?3041:474:1)
>     at async Hs.joinPool (index.umd.js?3041:474:1)
>     at async DeepPoolJoinHandler.queryJoin (deep-pool-join.handler.ts?85ac:69:1)
>     at async eval (join-pool.provider.ts?d9d4:201:1)
>     at async trackLoading (index.ts?f745:196:1)
> ```
> 
> Screenshot 2022-11-15 at 14 07 32

Most recent beta release of SDK must be used, so did you run `npm i` before? Also, is this mainnet, or polygon?

src/services/balancer/pools/joins/handlers/deep-pool-join.handler.ts
sssmi: `parsedAmountsIn` and `tokenAddresses` are both generated from the `amountsIn` array, so the order is always same as well

src/services/balancer/pools/joins/handlers/deep-pool-join.handler.ts
sssmi: Done

src/services/balancer/pools/joins/handlers/deep-pool-join.handler.ts
sssmi: Can I leave this as it is?

garethfuller:
@sssmi https://www.loom.com/share/a4a75c9cb3ce49b488a7934e515430d3

garethfuller (approved):

PR #2327, opened 2022-10-25
Fix: make sure restaking modal request token approval if needed

Matthews3301:
# Description

When restaking from old gauge to new gauge, the user needs to approve spend on the new gauge address. This is now included in the restaking modal.

## Type of change

## How should this be tested?

Have a look at a pool where you have balance staked in an old gauge. Remove all approvals for this pool in e.g. https://revoke.cash - then go to restake modal. You should see approval step.

## Visual context

N/A

## Checklist:

- [x]  I have requested at least 2 reviews (If the PR is significant enough, use best judgement here)

### DIFF
--- a/src/components/contextual/stake/StakePreview.vue
+++ b/src/components/contextual/stake/StakePreview.vue
@@ -51,7 +51,12 @@ const {
 } = useStaking();
 const { getTokenApprovalActionsForSpender } = useTokenApprovalActions(
   [props.pool.address],
-  ref([balanceFor(props.pool.address).toString()])
+  ref([
+    (props.action === 'restake'
+      ? stakedSharesForProvidedPool.value
+      : balanceFor(props.pool.address)
+    ).toString(),
+  ])
 );
 
 const stakeAction = {
--- a/public/data/tokenlists/tokens-1.json
+++ b/public/data/tokenlists/tokens-1.json
@@ -1,0 +1,13639 @@
+{
+  "https://raw.githubusercontent.com/balancer-labs/assets/master/generated/listed.tokenlist.json": {
+    "name": "Balancer",
+    "timestamp": "2022-10-22T00:00:00.000Z",
+    "logoURI": "https://raw.githubusercontent.com/balancer-labs/pebbles/master/images/pebbles-pad.256w.png",
+    "keywords": ["balancer", "listed"],
+    "version": { "major": 1, "minor": 0, "patch": 0 },
+    "tokens": [
+      {
+        "address": "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
+        "chainId": 1,
+        "name": "Wrapped Ether",
+        "symbol": "WETH",
+        "decimals": 18,
+        "logoURI": "https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2/logo.png"
+      },
+      {
+        "address": "0xba100000625a3754423978a60c9317c58a424e3D",
+        "chainId": 1,
+        "name": "Balancer",
+        "symbol": "BAL",
+        "decimals": 18,
+        "logoURI": "https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0xba100000625a3754423978a60c9317c58a424e3D/logo.png"
+      },
+      {
+        "address": "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
+        "chainId": 1,
+        "name": "USD Coin",
+        "symbol": "USDC",
+        "decimals": 6,
+        "logoURI": "https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48/logo.png"
+      },
+      {
+        "address": "0x6B175474E89094C44Da98b954EedeAC495271d0F",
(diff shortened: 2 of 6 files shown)


src/components/contextual/stake/StakePreview.vue
```diff
@@ -51,7 +51,12 @@ const {
 } = useStaking();
 const { getTokenApprovalActionsForSpender } = useTokenApprovalActions(
   [props.pool.address],
-  ref([balanceFor(props.pool.address).toString()])
+  ref([
+    (props.action === 'restake'
+      ? stakedSharesForProvidedPool.value
+      : balanceFor(props.pool.address)
+    ).toString(),
+  ])
```
garethfuller, line 59: Not a massive fan of having conditional logic within function calls like this. Could we extract the ref and call it something readable?

Matthews3301: Updated

src/components/contextual/stake/StakePreview.vue
```diff
@@ -49,14 +49,16 @@ const {
...
+]);
 const { getTokenApprovalActionsForSpender } = useTokenApprovalActions(
   [props.pool.address],
-  ref([
-    (props.action === 'restake'
-      ? stakedSharesForProvidedPool.value
-      : balanceFor(props.pool.address)
-    ).toString(),
-  ])
+  approvalAmounts
```
garethfuller, line 61: Ok, one more small refactor, I think this ref should be a computed property and only provide the single approval value relevant for the pool address. Something like this:
```ts
const approvalAmount = computed((): string => {
  return (props.action === 'restake'
    ? stakedSharesForProvidedPool.value
    : balanceFor(props.pool.address)
  ).toString()
});

const { getTokenApprovalActionsForSpender } = useTokenApprovalActions(
  [props.pool.address],
  [approvalAmount.value]
);

Comment (Matthews3301) [reply src/components/contextual/stake/StakePreview.vue]: Ok, we can reuse `numSharesToModify` for that.

File: src/components/contextual/stake/StakePreview.vue [hunk 3]
```diff
@@ -49,9 +49,15 @@ const {
   stakeBPT,
   unstakeBPT,
 } = useStaking();
+
+const numSharesToModify = computed(() =>
```
alter-eggo, line 53: are you sure it should be computed? 🤔

alter-eggo: may be just
```
const numSharesToModify = props.action === 'stake'
    ? balanceFor(getAddress(props.pool.address))
    : stakedSharesForProvidedPool.value
```

Matthews3301: Good suggestion 👍

garethfuller: What if `stakedSharesForProvidedPool` changes?

garethfuller: Probably best to leave it in a computed property if any of the underlying values are reactive. Unless I'm missing something?

alter-eggo: hm, there is not computed property in code below
```
const shareBalanceToDisplay = ref(
  props.action === 'unstake'
    ? stakedSharesForProvidedPool.value
    : balanceFor(props.pool.address)
);
```

garethfuller: I don't understand why you're not referencing this piece of code? We were discussing `numSharesToModify`.  Regardless I wasn't specifically talking about underlying computed properties, but reactive. `props.action` is reactive and `stakedSharesForProvidedPool.value` is reactive in the code snippet you just shared. `props.action` and `stakedSharesForProvidedPool.value` is reactive in the property in question, which means it probably shouldn't be a static variable.

src/components/contextual/stake/StakePreview.vue
```diff
@@ -49,9 +49,14 @@ const {
   stakeBPT,
   unstakeBPT,
 } = useStaking();
+
+const numSharesToModify =
+  props.action === 'stake'
+    ? balanceFor(getAddress(props.pool.address))
+    : stakedSharesForProvidedPool.value;
```
garethfuller, line 56: `props.action` and `stakedSharesForProvidedPool.value` are reactive properties, so I think this should be a computed property.

src/components/contextual/stake/StakePreview.vue
```diff
@@ -81,7 +86,7 @@ const isActionConfirmed = ref(false);
 const confirmationReceipt = ref<TransactionReceipt>();
 const stakeActions = ref<TransactionActionInfo[]>([]);
 const shareBalanceToDisplay = ref(
-  props.action === 'unstake'
+  props.action === 'unstake' || props.action === 'restake'
     ? stakedSharesForProvidedPool.value
     : balanceFor(props.pool.address)
 );
```
garethfuller, line 92: I understand that this was already set as a `ref` but this also looks like it should be a computed property. My understanding is that `ref`s are for properties that you need to be mutable, i.e. be able to modify elsewhere in the code. I'm pretty sure `shareBalanceToDisplay` does not need to be modified but it does need to be reactive based on the internal properties it depends on. Therefore it should be a computed property.

garethfuller (changes requested):

src/components/contextual/stake/StakePreview.vue
alter-eggo: as I understand the result of this code is exactly the same as it is in `numSharesToModify`, but written in other way. I mean previously  there was not need in reactive property there. maybe I am wrong, but I think there is not now as 
`numSharesToModify`  is used only on approval stage.

PR #2336, opened 2022-10-27
Refactor StakePreview proposal

garethfuller:
# Description

Removes redundant variable and sets currentShares as a static property.

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test stake preview works as expected.

## Visual context

n/a

## Checklist:

### DIFF
--- a/src/components/contextual/stake/StakePreview.vue
+++ b/src/components/contextual/stake/StakePreview.vue
@@ -50,13 +50,15 @@ const {
   unstakeBPT,
 } = useStaking();
 
-const numSharesToModify =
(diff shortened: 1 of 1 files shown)


garethfuller:
@alter-eggo sorry Anastas, I think you were right about keeping it static. It's not ideal that this is the case, but it works for this component. I think it's important that the value doesn't update once the transaction goes through and that's why it was originally set as a `ref` rather than a computed property.

alter-eggo (approved):

PR #2337, opened 2022-10-27
Hotfix: Add possibility to invest in mai4 pool 

alter-eggo:
# Description
Add possibility to invest in mai4 pool and remove warning on pools table.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:

### DIFF
--- a/src/components/forms/pool_actions/MigrateForm/constants.ts
+++ b/src/components/forms/pool_actions/MigrateForm/constants.ts
@@ -87,6 +87,8 @@ const POLYGON_POOL_MIGRATIONS: PoolMigrationInfo[] = [
   },
 ];
 
+export const migratablePoolsInvestExceptions = [POOLS.IdsMap.mai4?.mai4];
+
 const MIGRATIONS_BY_NETWORK = {
   [Network.GOERLI]: GOERLI_POOL_MIGRATIONS,
   [Network.MAINNET]: MAINNET_POOL_MIGRATIONS,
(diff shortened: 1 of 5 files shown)


src/components/forms/pool_actions/MigrateForm/constants.ts
```diff
@@ -87,6 +87,8 @@ const POLYGON_POOL_MIGRATIONS: PoolMigrationInfo[] = [
   },
 ];
 
+export const migratablePoolsInvestExceptions = [POOLS.IdsMap.mai4?.mai4];
```
garethfuller, line 90: This is a difficult one to name, because it's not just about the investment but also about not showing specific warnings. I'd suggest this is a class of migration that we could call 'soft'. It's soft in that we still present users with the option to migrate liquidity and keep the warning on the pool page, but we don't highlight this to the user in the pools list or block them from investing in the pool. So maybe we could change the name of this variable and a comment to describe this, something like:
```suggestion
/**
 * Checks if a pool is classified as a soft migration, a soft migration meaning that you can still invest in the 
 *  current pool and we don't highlight a forced migration for current LPs.
 */
export function isSoftMigratablePool(id: string) {
...
}
```

garethfuller (changes requested):

src/components/forms/pool_actions/MigrateForm/constants.ts
alter-eggo: hm, yeah, think it is better. thanks! 
fixed

garethfuller (approved):

PR #2340, opened 2022-10-28
Add ExitPoolProvider

garethfuller:
# Description

Adds the ExitPoolProvider. Note that it doesn't really work yet, this PR is just to get the files in place.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

n/a

## Visual context

n/a

## Checklist:

### DIFF
--- a/src/providers/local/exit-pool.provider.ts
+++ b/src/providers/local/exit-pool.provider.ts
@@ -1,0 +1,195 @@
+import { isDeep, tokenTreeNodes } from '@/composables/usePool';
+import useTokens from '@/composables/useTokens';
+import { useTxState } from '@/composables/useTxState';
+import {
+  HIGH_PRICE_IMPACT,
+  REKT_PRICE_IMPACT,
+} from '@/constants/poolLiquidity';
+import symbolKeys from '@/constants/symbol.keys';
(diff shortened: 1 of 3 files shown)


src/providers/local/exit-pool.provider.ts
```diff
@@ -0,0 +1,195 @@
...
+/**
+ * ExitPoolProvider
+ *
+ * Handles pool exiting state and transaction execution.
+ */
+const provider = (props: Props) => {
+  /**
+   * STATE
+   */
+  const pool = toRef(props, 'pool');
```
sssmi, line 44: Could this be a computed property so it stays in sync with the `pool` prop?

sssmi (changes requested):

src/providers/local/exit-pool.provider.ts
garethfuller: This is how toRef works: https://vuejs.org/api/reactivity-utilities.html#toref

"The created ref is synced with its source property: mutating the source property will update the ref, and vice-versa."

But either way is fine.

sssmi: Ah, I somehow missed that. All good then 👍

sssmi (approved):

PR #2357, opened 2022-11-01
Refactor network nav gaurds

garethfuller:
# Description

Moves logic around and extracts into functions to be more understandable.

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test redirect logic of routes for single app all networks.

## Visual context

n/a

## Checklist:

### DIFF
--- a/src/composables/useNavigationGuards.ts
+++ b/src/composables/useNavigationGuards.ts
@@ -9,7 +9,7 @@ NProgress.configure({ showSpinner: false });
 let delayedStartProgressBar;
 
 /**
- * Navigation gaurds that require Vue app context.
+ * Navigation guards that require Vue app context.
  */
 export default function useNavigationGuards() {
   const router = useRouter();
--- a/src/composables/useNavigationGuards.ts
+++ b/src/composables/useNavigationGuards.ts
@@ -1,89 +1,19 @@
 import NProgress from 'nprogress';
-import { Router, useRouter } from 'vue-router';
-
-import { Network } from '@balancer-labs/sdk';
-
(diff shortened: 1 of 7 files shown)


src/composables/useNavigationGuards.ts
```diff
@@ -1,91 +1,24 @@
 import NProgress from 'nprogress';
```
garethfuller, line 1: I've moved this logic into a higher-level nav-gaurds.ts file that is applied directly to the router since these redirects don't require app context.

src/composables/useNavigationGuards.ts
```diff
@@ -1,91 +1,24 @@
...
-        subdomainNetwork.toString(),
-        `${appUrl().replace(subdomain, 'app')}${
-          to.params.networkSlug ? '' : networkSlug
-        }${to.fullPath}`,
-        router
-      );
-    }
-    next();
-  });
+    localStorage.setItem('networkId', networkId.value.toString());
```
garethfuller, line 20: I've moved networkId localstorage setting here because it didn't feel right to have to do it in multiple places in nav-gaurds.ts. It also made some of those functions more awkward to write which made me think it was the wrong place to be setting it.

garethfuller: I also think it makes more sense to sync the localstorage with the app's current state in the app context based nav guards, which is what this file is for. I'm not 100% sure if this will break expected behaviour or not. Let me know if you think it will.

Matthews3301 (approved):

PR #2369, opened 2022-11-02
Fix: always redirect pool urls to mainnet

Matthews3301:
# Description

If url requested doesn't have networkSlug, and it is using pool route, then always redirect to ethereum mainnet url.

Also, on main page, show network name in pools heading.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:

### DIFF
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -28,7 +28,7 @@ const {
 } = useStreamedPoolsQuery(selectedTokens);
 const { upToMediumBreakpoint } = useBreakpoints();
 const { priceQueryLoading } = useTokens();
-const { networkSlug } = useNetwork();
+const { networkSlug, networkConfig } = useNetwork();
 
 const isInvestmentPoolsTableLoading = computed(
   () => dataStates.value['basic'] === 'loading' || priceQueryLoading.value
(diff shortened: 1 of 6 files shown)


src/pages/index.vue
```diff
@@ -50,9 +50,7 @@ function navigateToCreatePool() {
     <BalStack vertical>
       <div class="px-4 xl:px-0">
         <div class="flex justify-between items-end mb-8">
-          <h3>
-            {{ $t('investmentPools') }}
-          </h3>
+          <h3>{{ networkConfig.chainName }} {{ $t('pools') }}</h3>
```
garethfuller, line 53: Only issue with this is for Arbitrum it ends up being "Arbitrum One Pools". Maybe we can just change the chain name in the config? I don't think there should be any issue caused by that.

garethfuller (changes requested):

src/pages/index.vue
Matthews3301: Updated

garethfuller (approved):

PR #2375, opened 2022-11-03
Spike: simplify providers

agualis:
# Description

We currently use [Roots.ts](https://github.com/balancer-labs/frontend-v2/blob/develop/src/Root.ts) to provide/inject global providers so that they can be used by any component. This solution works but it makes testing setup challenging, as it is really difficult to instantiate a component under test with access those injected providers.
The main goal of those providers is providing access to global state and methods/computed-props to interact with that state. That can be also done with plain composables as described in [this post](https://vueschool.io/articles/vuejs-tutorials/state-management-with-composition-api/). The benefits would be: 
- The new code is easier to understand cause there's no need for provide/inject or for the recursive calls in Root.ts.
- The new code is easier to test because the components using the composables simply import and call them (the don't need a higher level component to provide anything for them).
- Some TS interfaces like https://github.com/balancer-labs/frontend-v2/blob/d0dde3375f97fd39ba97146dad2cf6d58bafb037/src/providers/tokens.provider.ts#L51 could be deleted because the response type would be automatically inferred by TS.
- The relationships between "provider" modules become more explicit which helps detecting and fixing potential cyclic dependency problems. 

For the sake of a clearer review I kept the original file structure. For example: 
- `tokens-provider.ts` defines the global state and contains a public method that I called `getTokensProvision` which exports _state + computed + methods_.
- `useTokens.ts` simply export the previous method as `useTokens()`.

In a final refactor, I would merge those 2 files into one. 

IMO the biggest advantage of this change is that it allows us to test components without mocking the providers (we could still do that if we want) because we just need to mock the api calls with MSW (most of them could be mocked once globally to be reused for all tests). That would align with testing library mindset: 

> The more your tests resemble the way your software is used, the more confidence they can give you.

2 examples that illustrate how much we would simplify testing: 

https://github.com/balancer-labs/frontend-v2/pull/2375/files#diff-7b7c438140e37f8fdb2a851b9e0645c2983b6b91687870ee2f9f3f13df9ab782L8-L14 

https://github.com/balancer-labs/frontend-v2/pull/2375/files#diff-854b90a15a0fd9c299c4d8264d5fe9749d0efb15870c3491beaee364c2a90b6c

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:


src/providers/token-lists.provider.ts
```diff
@@ -21,121 +12,101 @@ export interface TokenListsState {
...
+    activeTokenLists,
+    defaultTokenList,
+    balancerTokenLists,
+    approvedTokenLists,
+    vettedTokenList,
+    // methods
+    toggleTokenList,
+    isActiveList,
+  };
+}
```
garethfuller, line 112: Why does this need to be a function? Also the Provision part of `getTokensListProvision` doesn't really make sense. What about:
```suggestion
export const tokenListsProvider = {
    // state
    ...toRefs(state),
    // computed
    allTokenLists,
    activeTokenLists,
    defaultTokenList,
    balancerTokenLists,
    approvedTokenLists,
    vettedTokenList,
    // methods
    toggleTokenList,
    isActiveList,
}
```

garethfuller (changes requested):
You removed Root.ts which was injecting these providers, where are they all instantiated now?

src/providers/token-lists.provider.ts
agualis: Yeah, I kept the two original files (`foo.provider.ts` and `useFoo.ts`) to make the code review simpler but the idea would be deleting `foo.provider.ts` with a const as you suggest.

agualis:
> You removed Root.ts which was injecting these providers, where are they all instantiated now?

If we refactor the "providers" into composables, we don't need provide/inject anymore. They are simply imported and used in setup with `useFoo()`.

In the description, I tried to explain the advantages of it, although, there is one big disadvantage:

while the global code outside the `useFoo` function is just called once, the code inside useFoo is run every `useFoo` call.  As you know, that can affect performance in certain situations (like when using `watch`) but, I think, that we will be able to extract most of the code to the global scope in most of the cases.

agualis:
@garethfuller Note that, in the tokens.provider.ts, we could extract more global computed and functions to the global scope so that they are not run every time the composable is used but there is one restriction: 

Functions/computed using vue-query cannot be outside the useProvider* function (global scope) because they need to be executed inside setup method. Because of that, I think it is better to leave everything inside `useProvider*` and move things up (to the global scope) just when performance is heavily affected.

garethfuller:
Closing in favour of #2576

PR #2399, opened 2022-11-07
Remove network specific fathom goals

garethfuller:
# Description

When we used subdomains we needed to track the same Fathom Analytic goals across apps. We now only have one Fathom Analytics app to track so logic can be simplified.

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test app works as expected. 

## Visual context

n/a

## Checklist:


Matthews3301 (approved):
Looks good, the only tiny thing is that `Goals` const could be lower case as it's not a class anymore.

alter-eggo (approved):

PR #2414, opened 2022-11-14
Feat: join/exit Aura swap trade

Matthews3301:
# Description

With the release of new Balanced SDK version, it is now possible to perform a new type of swap, "Join/Exit" swap. This uses batch relayer to perform multiple transactions which join and exit given pool.

Example of how to use this:
https://github.com/balancer-labs/balancer-sdk/blob/develop/balancer-js/examples/swapSor.ts

The swap type priority is this:
1. If user selected "Gasless" AKA "CoW swap" AKA "Gnosis" > use Gnosis trade.
2. If user unselected "Gasless" or it's not available > check if JoinExit is available and use it.
3. If JoinExit trade is not available > use SOR

Please see below on how to test.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

On Mainnet, test different trade types.
On Polygon, BAL -> AVAX swap should use JoinExit.
On Polygon, BAL -> MATIC swap should not use JoinExit.
Check other networks.

## Visual context

N/A

## Checklist:


Matthews3301:
@garethfuller In terms of gas used and transaction fees, this new route seems to be more expensive than "balancer" route. There is also an additional initial step. E.g.:

Tested on Polygon:
BAL -> MATIC (“balancer” route)
1. Approve BAL (52,072 gas)
2. Confirm swap (109,642 gas)

BAL -> AVAX (“joinExit” route)
1. Approve Batch relayer (49,327 gas)
2. Approve BAL (52,072 gas)
3. Confirm swap (162,562 gas)

src/composables/trade/useJoinExit.ts
```diff
@@ -0,0 +1,258 @@
...
+    await injectTokens(unknownAssets);
+    await handleAmountChange();
+  });
+
+  watchEffect(async () => {
+    swapInfo.value = await balancer.sor.getSwaps(
+      tokenInAddressInput.value,
+      tokenOutAddressInput.value,
+      exactIn.value ? SwapTypes.SwapExactIn : SwapTypes.SwapExactOut,
+      parseFixed(tokenInAmountInput.value || tokenOutAmountInput.value, 18),
```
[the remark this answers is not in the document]
Matthews3301: Good spot, thank you.

src/components/modals/TradePreviewModalGP.vue
```diff
@@ -815,21 +891,22 @@ watch(blockNumber, () => {
...
-                trading.isGnosisTrade.value
+                trading.isGnosisTrade.value && !trading.isJoinExitTrade
                   ? $t('tradeSummary.transactionTypesTooltips.sign.content')
                   : $t('tradeSummary.transactionTypesTooltips.trade.content')
               }}
             </div>
           </div>
         </BalTooltip>
       </div>
+      {{ relayerSignature }}
```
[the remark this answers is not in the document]
Matthews3301: Fixed

src/components/modals/TradePreviewModalGP.vue
```diff
@@ -865,6 +942,16 @@ watch(blockNumber, () => {
       >
         {{ $t('approveLidoRelayer') }}
       </BalBtn>
+      <BalBtn
```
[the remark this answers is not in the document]
garethfuller: Agree with Simeon, let's take the opportunity to refactor and use BalActionSteps here, it will take quite a bit of work but if we don't do it as part of this work we might never do it. Also please remove GP from the component name,

src/composables/trade/useTrading.ts
```diff
@@ -103,15 +106,30 @@ export default function useTrading(
...
+        // Currently Relayer only suitable for ExactIn and non-eth swaps
+        exactIn.value &&
+          tokenInAddressInput.value.toLowerCase() !==
+            AddressZero.toLowerCase() &&
+          tokenOutAddressInput.value.toLowerCase() !==
+            AddressZero.toLowerCase() &&
+          joinExit.swapInfo.value?.returnAmount &&
+          !joinExit.swapInfo.value?.returnAmount.isZero()
+          ? 'joinExit'
+          : 'balancer'
```
garethfuller, line 122: This is very difficult to parse, please refactor into more understandable conditional variables.

src/components/modals/TradePreviewModalGP.vue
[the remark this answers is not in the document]
Matthews3301: Good idea, done.

src/composables/trade/useTrading.ts
Matthews3301: 👍

src/components/modals/TradePreviewModal.vue
```diff
@@ -1,305 +1,800 @@
...
+}
+
+function cofirmPriceUpdate() {
+  priceUpdated.value = false;
+  priceUpdateAccepted.value = true;
+  lastQuote.value = props.trading.getQuote();
+}
+
+function handlePriceUpdate() {
+  if (lastQuote.value != null) {
```
alter-eggo, line 453: it might be better:
```
if (lastQuote.value === null) {
      return;
}
```
than put all that code in one more if statement

src/components/modals/TradePreviewModal.vue
```diff
@@ -1,305 +1,800 @@
...
+const gnosisRelayerApproval = useRelayerApprovalTx(
+  RelayerType.GNOSIS,
+  props.trading.isGnosisTrade
+);
 
-    async function approveLidoRelayer(): Promise<void> {
-      await lidoRelayerApproval.approve();
-    }
+const pools = computed<SubgraphPoolBase[]>(() => {
+  return props.trading.sor.pools.value;
```
alter-eggo, line 262: why we need to put this in computed property?

src/components/modals/TradePreviewModal.vue
```diff
@@ -1,305 +1,800 @@
...
-  emits: ['trade', 'close'],
-  setup(props, { emit }) {
-    const { fNum2, toFiat } = useNumbers();
+const PRICE_UPDATE_THRESHOLD = 0.02;
 
-    const { addressIn, amountIn, addressOut } = toRefs(props);
+type Props = {
+  trading: UseTrading;
+  error?: any;
+  warning?: any;
```
alter-eggo, line 31: may be we can provide some specific type here?

alter-eggo (changes requested):

Matthews3301:
@alter-eggo Thanks. Sorry, I should have mentioned this: as part of this merge request, file TradePreviewModalGP.vue has been renamed to TradePreviewModal.vue. This makes the diff seem like there are many changes in this file. However, it should be compared to https://github.com/balancer-labs/frontend-v2/blob/develop/src/components/modals/TradePreviewModalGP.vue

I think if we refactor all the points you've suggested, it should be a separate merge request as they don't relate to joinExit.

garethfuller (changes requested):
Swaps are broken, testing WMATIC > BAL on polygon for instance results in not output amount.

Matthews3301:
@garethfuller hmm I can't seem to reproduce this. It's swapping okay for me. Would you mind recording the screen and console for this please?

Matthews3301:
@garethfuller I tried finding a pair of tokens which would help testing this live, however it only WETH <> AuraBAL pair was the only one I could find. It also depends on the amounts and market conditions.

I swapped using this pair:
https://etherscan.io/tx/0xa74e7983a383fb4e5ac0b6cff102703e6716cbea23c5fb1da938a3365370ed9b

Seems to be using the correct swap route method?

src/composables/trade/useTrading.ts
```diff
@@ -265,6 +315,10 @@ export default function useTrading(
       if (!gnosis.hasValidationError.value) {
         gnosis.handleAmountChange();
       }
+    } else if (isJoinExitTrade.value) {
```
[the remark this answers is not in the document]
garethfuller: I think this refactor is valid. Yeah it's changing from how gnosis above is implemented but it's an obvious cleanup.

src/composables/trade/useJoinExit.ts
```diff
@@ -0,0 +1,318 @@
...
+        balancer.networkConfig.addresses.tokens.wrappedNativeAsset,
+        String(slippageBufferRate.value * 1e4),
+        relayerSignature.value || undefined
+      );
+
+      const signer = getSigner();
+      const relayerContract = await balancer.contracts.relayerV4?.connect(
+        signer
+      );
+      const tx = await relayerContract?.multicall(relayerCallData.rawCalls);
```
garethfuller, line 198: All transactions **must** go through the TxBuilder. We handle a bunch of things there around gas settings etc. Search for uses of `txBuilder.contract.sendTransaction` to see how to construct the call in this scenario.

garethfuller (changes requested):

src/composables/trade/useJoinExit.ts
Matthews3301: Thanks, updated.

src/composables/trade/useJoinExit.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+jest.mock('@/composables/approvals/useRelayerApproval', () => ({
+  __esModule: true,
+  default: () =>
+    jest.fn().mockImplementation(() => ({
+      relayerSignature: '-',
+    })),
+  RelayerType: {
+    BATCH_V4: 'BATCH_V4',
+  },
+}));
```
agualis, line 37: Our current code forces to do a very complex mocking setup. We will simplify this in incoming PRs.

src/composables/trade/__mocks__/mockSorSwapInfo.ts
```diff
@@ -0,0 +1,64 @@
+export default function mockSorSwapInfo() {
```
agualis, line 1: Using this kind of hardcoded untyped mocks is painful cause:
- Tests will get broken whenever the SDK team changes the `SwapInfo` type. 
- The error will wont't be easy to spot 
-  It will require manually fixing this file.  

The alternative is simple: using jest/vitest-mock-extended to automatically match the type.

That is, replacing the whole content of this file with: 

```ts
import { SwapInfo } from '@balancer-labs/sdk';
import { mock } from 'jest-mock-extended';

export default function mockSorSwapInfo() {
  return mock<SwapInfo>();
}
```

agualis: Same goes for `mockSorOutput.json` and `mockTokensOutput.json`, but let's leave that change for incoming PRs where I will suggest a simpler alternative.

garethfuller: Only thing I'd suggest is if we can trim down those mocks, they are HUGE.

Matthews3301: @agualis Sure, thanks. For this test to work, there needs to be a very specific combination of pool, tokens and amounts - will this be preserved?

agualis: Check [this](https://github.com/balancer-labs/frontend-v2/pull/2414/commits/b5665bc0cd7f3d90b4f7f84c13e5ef63ecc2bbed) out. It is far from being my preferred solution but it avoids the huge jsons. 

I have much more ideas to improve this farther (in incoming discussions/PRs).

Please check if the key parts are preserved or I can add them otherwise.

agualis: Agree and, as I said, I would improve this farther with more time. Let's have a chat to double check that it tests the same logic that it was testing before.

agualis: I just pushed a new version that: 
1. Covers useJoinExit as it used to
2. Uses much smaller mocks and moved from json to TS so that tests are less fragile

Again, far from being the best solution but it should work for now.

Matthews3301:
Continued in https://github.com/balancer-labs/frontend-v2/pull/2678

PR #2427, opened 2022-11-15
Hotfix: update event tracking

garethfuller:
# Description

Some events had been dropped in previous refactoring and some were named incorrectly, this PR fixes those things and adds a couple more events.

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test app functions as normal.

## Visual context

n/a

## Checklist:


src/composables/trade/useSor.ts
```diff
@@ -496,7 +496,7 @@ export default function useSor({
       onTxConfirmed: () => {
         trading.value = false;
         latestTxHash.value = tx.hash;
-        trackGoal(Goals.Swapped);
+        trackGoal(Goals.Swapped, Number(tokenInAmountInput.value) || 0);
```
timjrobinson, line 499: Is there a way to add the token symbol or USD value here? as the token amount by itself isn't a very useful metric .

timjrobinson (approved):
One comment, otherwise LGTM

Matthews3301 (approved):
Looks okay, I'm not 100% sure about the purpose of specific tracking events, however, would it not be better to track pages being opened via `beforeMount` on each page rather than link clicks? Some links don't have the tacking behaviour, e.g. in the Footer.

garethfuller:
> Looks okay, I'm not 100% sure about the purpose of specific tracking events, however, would it not be better to track pages being opened via `beforeMount` on each page rather than link clicks? Some links don't have the tacking behaviour, e.g. in the Footer.

@Matthews3301 Yeah good point, but individual page views are tracked, the click events are more just about understanding what users are using to navigate the site. They probably aren't that useful but the more data the better I guess.

PR #2440, opened 2022-11-18
Fetch Pool APR's using SDK

timjrobinson:
# Description

Switch APR fetching from local calculation to using the SDK instead. Update types to use SDK Pool, PoolType and APRBreakdown types. Switch all locations that APR is displayed to work with the new APRBreakdown type.

This requires adding two new PoolType's to the SDK: https://github.com/balancer-labs/balancer-sdk/pull/208, there are some hacks around this which can be removed once that is merged and SDK is updated. Using SDK PoolType was required now that Pool is extended from the SDK Pool type. 

## Type of change

- [x] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- Compare this branch to production, ensure all APR's are roughly equivalent (there are some differences due to updates in the way the SDK calculates APR's, these are listed here: https://github.com/balancer-labs/balancer-sdk/pull/98
- Ensure that all APR tooltips show accurate info and the values add up to the main value displayed
- Ensure APR's show correctly on individual pool pages and they match the APR on the main pools list. 

## Checklist:


src/components/contextual/pages/pool/PoolContractDetails.vue
```diff
@@ -14,6 +14,12 @@ type Props = {
   pool: Pool;
 };
 
+enum LocalPoolTypes {
```
timjrobinson, line 17: This is a hack until the Managed and Linear pool types are added to the SDK.

agualis (changes requested):
Great to remove complexity from the frontend app 😊

Screenshot 2022-11-18 at 07 31 40

I think that this is a good opportunity to add some component tests.  Maybe we could start with `AprBreakdown` component. There is a pending PR to make testing simpler but I think that we can start with some tests already. If you agree, I can support you with this.

timjrobinson:
@agualis good idea! Will add them today

src/components/tooltips/APRTooltip/APRTooltip.spec.ts
```diff
@@ -124,5 +128,35 @@ describe('APRTooltip', () => {
...
+        ...EmptyPoolMock,
+        tokensList: [configService.network.addresses.wstETH],
+      };
+      const { container } = render(APRTooltip, {
+        props: {
+          pool: poolMock,
+          poolApr: aprBreakdown,
+        },
+      });
+      expect(container.getElementsByClassName('total-apr')[0].textContent).toBe(
```
agualis, line 154: These tests are great but I would like to start a little discussion:

[Testing library](https://testing-library.com/docs/queries/about#priority) recommends to use accesible or semantic queries instead of CSS classnames. 

At the same time, in this concrete case, using `getByRole` or `getByText` would not be so specific as `getElementsByClassName('total-apr')` that you used here. 

Because of that, in this case I would use `data-testid` attribute in the template like this: 

```html
 <div
        data-testid="total-apr"
        class="px-3 pt-3 pb-1 bg-gray-50 dark:bg-gray-800 rounded-t total-apr"
 >
```
so that we can replace the expects block by this simpler to read and less brittle version:

```js
expect(screen.getByTestId('total-apr').textContent).toBe(
        'Total APR15.22%'
);

expect(screen.getByTestId('swap-fee-apr').textContent).toBe(
      '15.22% Swap fees APR'
);
```
I agree that it would involve adding a bit of pollution to the html ([related discussion here](https://github.com/testing-library/react-testing-library/issues/683)) just for testing purposes but, I think that it is worthy given the tradeoffs. 

What do you think? Did you already had this conversation in the team?

agualis (changes requested):

src/constants/pools.ts
```diff
@@ -6,7 +6,7 @@ export const MIN_FIAT_VALUE_POOL_MIGRATION = isMainnet.value ? 100_000 : 1; // 1
 
 // Do not display APR values greater than this amount; they are likely to be nonsensical
 // These can arise from pools with extremely low balances (e.g., completed LBPs)
-export const APR_THRESHOLD = 10_000;
+export const APR_THRESHOLD = 1_000_000;
```
garethfuller, line 9: I think we want to prevent anything over 10,000% Is that what this is for basis points?

src/components/tooltips/APRTooltip/APRTooltip.spec.ts
```diff
@@ -53,6 +54,35 @@ describe('APRTooltip', () => {
...
+        swapFees: 78,
+        protocolApr: 117,
+        min: 78,
+        max: 195,
+      };
+      const poolMock: Pool = {
+        ...EmptyPoolMock,
+        id: POOLS.IdsMap?.veBAL || '',
+      };
+      console.log('POOL ID: ', poolMock.id);
```
agualis, line 70: You forgot this console.log

agualis (changes requested):

garethfuller (changes requested):
APRs shouldn't be a range on L2s. Since there is no boost. 
Screenshot 2022-11-21 at 15 53 56

src/components/tooltips/APRTooltip/APRTooltip.spec.ts
timjrobinson: Good idea, implemented.

agualis (approved):
Approved to unblock this PR although I don't have enough context to review the APR calculations. Good job 💪

src/services/pool/decorators/pool.decorator.ts
```diff
@@ -70,15 +70,8 @@ export class PoolDecorator {
...
-      if (setAprCondition) {
-        await poolService.setAPR(
-          poolSnapshot,
-          prices,
-          currency,
-          protocolFeePercentage,
-          gaugeBALAprs[pool.id],
-          gaugeRewardTokenAprs[pool.id]
-        );
+      if (setAprCondition && poolSnapshot) {
```
alter-eggo, line 73: `protocolFeePercentage, gaugeBALAprs, gaugeRewardTokenAprs` were used in calculation on frontend, it seems there is no more need in them so we can remove `getData` function and `setAprCondition`. 

also I am not sure why `poolSnapshot` is added in if statement?

src/composables/queries/usePoolAprQuery.ts
```diff
@@ -84,37 +58,16 @@ export default function usePoolAprQuery(
...
-    const _snaphshot = await getSnapshot(_pool.id);
-    const apr = await new AprConcern(_pool).calc(
-      _snaphshot[0],
-      prices.value,
-      currency.value,
-      protocolFeePercentage,
-      gaugeBALAprs[_pool.id],
-      gaugeRewardTokenAprs[_pool.id]
-    );
+    // const _snapshot = await getSnapshot(_pool.id);
```
alter-eggo, line 63: should be deleted?

alter-eggo (changes requested):

src/services/pool/decorators/pool.decorator.ts
alter-eggo: here should be `if (includeAprs)`

timjrobinson:
In testing I've noticed on rare occasions the APR's will become infinity, this seems to be coming from swapFee APR's becoming infinity. There is probably a divide by zero somewhere because some data hasn't loaded yet.

PR #2495, opened 2022-12-03
Fix issues with APR Tooltip on Polygon

timjrobinson:
# Description

Fixes a few issue with Tooltips on Polygon / L2's since the General Boosted Pools update merged yesterday

- When BAL rewards min/max are the same, show them as one line item instead of having a breakdown.
- If a pool has reward tokens for staking, show them as a line item under staking rewards
- If BAL rewards min/max are the same and the pool has stakign rewards, show two line items of BAL, then rewards.
- Label stMATIC staking rewards

## Type of change

## How should this be tested?

- View the APR Tooltips on Polygon, ensure they look correct, aren't showing ranges where min/max are the same, and data isn't missing

## Visual context

Before: 

2022-12-03-175330_208x244_scrot

After: 

2022-12-03-175304_262x212_scrot

Without any additional staking reward tokens (just BAL):

2022-12-03-180324_177x152_scrot

## Checklist:


timjrobinson:
The 'Rewards APR' is LDO token rewards. @pkattera should we show individual rewards tokens? It could say `LDO APR` instead of `Rewards APR`.

garethfuller (approved):

PR #2515, opened 2022-12-06
Fix BigNumber from method error

alter-eggo:
# Description
BigNumber.from() throws an error when there is more than 15 digits provided, this error falls into sentry. I am not sure that we should process this specifically.
Or may be we can restrict typing more than 15 digits in inputs. what do you think @pkattera @garethfuller ?
or show some hint?

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:

### DIFF
--- a/src/composables/trade/useSor.ts
+++ b/src/composables/trade/useSor.ts
@@ -352,7 +352,7 @@ export default function useSor({
 
       console.log('[SOR Manager] swapExactIn');
 
-      const swapReturn: SorReturn = await sorManager.getBestSwap(
+      const swapReturn = await sorManager.getBestSwap(
         tokenInAddress,
         tokenOutAddress,
         tokenInDecimals,
@@ -360,6 +360,9 @@ export default function useSor({
         SwapTypes.SwapExactIn,
         tokenInAmountScaled
       );
+      if (!swapReturn) {
+        return;
+      }
 
       sorReturn.value = swapReturn; // TO DO - is it needed?
       const tokenOutAmountNormalised = bnum(
@@ -404,7 +407,7 @@ export default function useSor({
 
       console.log('[SOR Manager] swapExactOut');
 
-      const swapReturn: SorReturn = await sorManager.getBestSwap(
+      const swapReturn = await sorManager.getBestSwap(
         tokenInAddress,
         tokenOutAddress,
         tokenInDecimals,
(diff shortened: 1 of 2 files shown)


src/composables/trade/useSor.ts
```diff
@@ -352,14 +352,17 @@ export default function useSor({
...
         tokenInAddress,
         tokenOutAddress,
         tokenInDecimals,
         tokenOutDecimals,
         SwapTypes.SwapExactIn,
         tokenInAmountScaled
       );
+      if (!swapReturn) {
+        return;
+      }
```
garethfuller, line 365: I'm not sure this is a good idea? Thrown errors render as alerts to users and are sent to Sentry, if we just return here no one will know that something failed.

src/lib/utils/balancer/helpers/sor/sorManager.ts
```diff
@@ -171,6 +176,16 @@ export class SorManager {
     };
   }
 
+  getSwapAmount(amountScaled: OldBigNumber): BigNumber | void {
+    try {
+      return BigNumber.from(amountScaled.toString());
+    } catch (e) {
+      // BigNumber.from() throws an error when there is more than 15 digits
+      // not sure that we should process this specifically
+      console.log(e);
```
garethfuller, line 185: Again I don't think this is a good idea, we should let it throw the error so we can handle that. Instead, we should be preventing the user from inputting more than 15 digits.

garethfuller (changes requested):

src/composables/trade/useSor.ts
alter-eggo: I put in try/catch only calculating swapAmount, so  if there will be any error in `sorManager.getBestSwap` except calculating swapAmount, this error will occur in this function as well and error will be sent to sentry

src/lib/utils/balancer/helpers/sor/sorManager.ts
alter-eggo: yeah, input limit seems better here

garethfuller: There should already be an input limit on the TokenInput so could you check why that isn't working in this case? Maybe it's a copy and past issue or something like that.

alter-eggo:
this pr is no more actual. error is fixed by https://github.com/balancer-labs/frontend-v2/pull/2522

PR #2518, opened 2022-12-06
Hotfix: revert token input changes

garethfuller:
# Description

TokenInput used to have this internal state for amount/address. I know it seems like we can just use prop values here but I seem to remember a specific reason it was like this. I believe this might be causing a lot of the exceptions we are seeing in the swap UI after merging the boosted pool work. The changes to the TokenInput seem to be the only thing that could cause this after looking through the diff. So this PR reverts to using internal state for this component.

We should deploy this and see if the exceptions are fixed in Sentry. If not we can investigate further.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test TokenInputs in swap UI and join/exit UIs.

## Visual context

n/a

## Checklist:

### DIFF
--- a/src/components/inputs/TokenInput/TokenInput.vue
+++ b/src/components/inputs/TokenInput/TokenInput.vue
@@ -203,9 +203,8 @@ const decimalLimit = computed<number>(() => token.value?.decimals || 18);
 /**
  * METHODS
  */
-function changeAmount(amount: InputValue) {
+function handleAmountChange(amount: InputValue) {
   const safeAmount = overflowProtected(amount, decimalLimit.value);
-  _amount.value = safeAmount;
   emit('update:amount', safeAmount);
 }
 
@@ -216,7 +215,7 @@ const setMax = () => {
     ? props.customBalance
     : getMaxBalanceFor(_address.value, props.disableNativeAssetBuffer);
 
-  changeAmount(maxAmount);
+  handleAmountChange(maxAmount);
 };
 
 /**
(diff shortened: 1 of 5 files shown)


src/components/inputs/TokenInput/TokenInput.vue
```diff
@@ -246,14 +255,14 @@ const setMax = () => {
     inputAlignRight
     @blur="emit('blur', $event)"
     @input="emit('input', $event)"
-    @update:model-value="handleAmountChange($event)"
+    @update:model-value="emit('update:amount', $event)"
```
garethfuller, line 257: Overflow protection is handled by BalTextInput itself so we shouldn't go through `changeAmount` function here and instead just emit event.

timjrobinson (approved):
LGTM Also appears to fix this issue we're seeing in prod: https://sentry.io/organizations/balancer-labs/issues/3788923769/?project=5725878&query=is%3Aunresolved+invalid+BigNumber+string&referrer=issue-stream&statsPeriod=14d so will merge now.

src/lib/utils/balancer/helpers/sor/sorManager.ts
```diff
@@ -147,7 +146,7 @@ export class SorManager {
       v2TokenIn.toLowerCase(),
       v2TokenOut.toLowerCase(),
       swapType,
-      BigNumber.from(amountScaled.toString()),
```
timjrobinson, line 150: This is the line that was throwing an exception in prod. `amountScaled` was being passed through as `1.23456e+23` which was coming from the scale function in useSor and then BigNumber didn't know what to do with this value.

PR #2535, opened 2022-12-08
Fix token input errors

alter-eggo:
# Description
This pr fixes the following sentry errors:
https://sentry.io/organizations/balancer-labs/issues/3079646365/?project=5725878&query=is%3Aunresolved&referrer=issue-stream
https://sentry.io/organizations/balancer-labs/issues/2999209081/?project=5725878&query=is%3Aunresolved&referrer=issue-stream

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?
- Go to swap page
- Choose WstEth as token out
- Type token out amount, switch tokens, try to type again
There shouldn't be any errors in console (except possible 429 from coingecko)

## Visual context

## Checklist:


src/composables/trade/useSor.ts
```diff
@@ -402,8 +402,12 @@ export default function useSor({
       if (!sorReturn.value.hasSwaps) {
         priceImpact.value = 0;
       } else {
-        tokenInAmount = await adjustedPiAmount(tokenInAmount, tokenOutAddress);
-
+        if (isMainnet.value) {
+          tokenInAmount = await adjustedPiAmount(
+            tokenInAmount,
+            tokenOutAddress
+          );
```
garethfuller, line 409: Are you sure we don't actually need to do this on L2s like Polygon?

alter-eggo: I am not sure, @markusbkoch can you help here please

markusbkoch: I don't think this conditional should be introduced here. I don't think it would break anything under the current circumstances, but it seems to me like `adjustedPiAmount` is meant to be a catch-all for cases like the stETH/wstETH one, but not limited to this specific case.

src/lib/utils/balancer/wrapper.ts
```diff
@@ -47,7 +49,7 @@ export const getWrapOutput = (
   if (wrapper === wstETH) {
     return wrapType === WrapType.Wrap
       ? getWstETHByStETH(wrapAmount)
```
markusbkoch, line 51: Doesn't this function need to be updated in the same way too?

markusbkoch: It would have to output the inverse of `getRate()`

src/composables/trade/useSor.ts
alter-eggo: I changed the condition to `isMainnet.value || isArbitrum.value` as there are currently rate providers addresses only for these networks

markusbkoch: Right, what I mean is that IIUC this check is redundant with the check that already exists within the `adjustedPiAmount` function. So unless I'm missing something this condition means we have to maintain two pieces of code instead of one when a new rate provider for wstETH comes along on Polygon for example, or when we add support for wrapping a new token that's similar to stETH. Like I said, the conditional here doesn't seem to break anything for the current state of things. But it does seem to add overhead to our work in the future.

alter-eggo: if we remove condition, it will break calculation only on goerli, can we add rate provider there?

alter-eggo: thanks, I will remove conditions

src/lib/utils/balancer/wrapper.ts
alter-eggo: fixed

markusbkoch: I meant don't we have to refactor the original implementation of `getWstETHByStETH` so that the amount is converted base on rateProvider.getRate() instead of based on a call to the wstETH contract?

src/lib/utils/balancer/lido.ts
```diff
@@ -47,15 +55,23 @@ export function getWstETHByStETH(stETHAmount: BigNumberish) {
...
     rpcProviderService.jsonProvider
   );
-  return wstETH.getStETHByWstETH(wstETHAmount);
+
+  const rate = await wstEthRateProvider.getRate();
+  amount = isStEthByWstEth
+    ? amount.mul(rate).div(ONE)
+    : amount.mul(ONE).div(rate);
+
+  return amount;
```
garethfuller, line 76: Let's refactor this function a bit.
1. Let's rename it to be a bit more clear.
2. Since it should only be relevant on mainnet lets throw if not.
3. Let's change the inputs to an object so that it's clear externally what the direction of the conversion is.

```ts
const MAINNET_RATE_PROVIDER_ADDRESS = '0x72d07d7dca67b8a406ad1ec34ce969c90bfee768'

type ConversionParams = {
  amount: BigNumber,
  isWrap: boolean,  // e.g. is stETH to wstETH
}

/**
 * Convert stETH amount to wstETH or vice versa. Only relevant on mainnet when wrapping or unwrapping.
 *
 * @param {BigNumber} amount - The amount to convert, could be stETH or wstETH value.
 * @param {boolean} isWrap - True if wrapping stETH to wstETH, false if unwrapping wstETH to stETH.
 * @returns Converted value for wrap or unwrap, if input is stETH, returns wstETH value and vice versa.
 */
export function convertStEthWrap({ amount, isWrap }: ConversionParams): BigNumber {
  try {
      const rateProvider = new Contract(
        MAINNET_RATE_PROVIDER_ADDRESS,
        ['function getRate() external view returns (uint256)'],
        rpcProviderService.jsonProvider
      );

      const rate = await rateProvider.getRate();

      return isWrap
        ? amount.mul(rate).div(ONE)
        : amount.mul(ONE).div(rate);
   } catch (error) {
      throw new Error(`Failed to convert stETH: ${error}`);
   }
}
```

src/components/inputs/TokenInput/TokenInput.vue
```diff
@@ -203,8 +205,12 @@ const decimalLimit = computed<number>(() => token.value?.decimals || 18);
 /**
  * METHODS
  */
-function handleAmountChange(amount: InputValue) {
+function handleAmountChange(amount: InputValue, isSetMax = false) {
   const safeAmount = overflowProtected(amount, decimalLimit.value);
+  if (isSetMax) {
+    emit('setMax', safeAmount);
+    return;
+  }
```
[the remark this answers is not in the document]
alter-eggo: I implemented this because of `handleTokenSwitch` function. now when user switches tokens, `exactIn` value is set improperly due to the sideEffects of token switching complicated logic (it changes twice)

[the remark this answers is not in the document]
garethfuller: I still don't understand why we would need to add a setMax event because of a token switch event. Could we not fix the underlying issue? Why are things changing twice? If you think it might help, you could record a loom going through the interaction and the code to show what the actual issue is?

mikemcdonald:
What's needed to get this over the finish line?

garethfuller:
> What's needed to get this over the finish line?

@mikemcdonald I'd quite like to see the refactor I suggested above done. However, Anastas is off this week. I might be able to fit it in  tomorrow though.

PR #2547, opened 2022-12-11
Fix portfolio staked pools list

timjrobinson:
# Description

There were two bugs causing the staked pools list to show every single pool instead of only the ones the user had a stake in

- The first bug was there where the query wasn't being updated when poolIds was of length 0
- Second bug was the Balancer API doesn't accept an empty array for pool id in. So I've added a hack for it to just pass a single empty string instead. It also didn't like if both in and not_in were in the same query so I deleted the not_in. Will fix the API correctly soon.

## Type of change

## How should this be tested?

- Visit your portfolio page, it should show pools you are staking in, or an empty list if you're not staking in any pools. 

## Checklist:

### DIFF
--- a/src/components/tooltips/APRTooltip/components/StakingBreakdown.vue
+++ b/src/components/tooltips/APRTooltip/components/StakingBreakdown.vue
@@ -1,7 +1,7 @@
 <script lang="ts" setup>
 import { computed } from 'vue';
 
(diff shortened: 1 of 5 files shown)


garethfuller (approved):
Thanks, looks like it works. Portfolio page taking a long time to load though, we should investigate that.

PR #2549, opened 2022-12-12
Improve + simplify token yield breakdown in APR Tooltip

timjrobinson:
# Description

- When there is a single token giving yield show <symbol> APR
- When there are multiple tokens giving yield show a breakdown with a generic header of 'Token APR'
- Removes a bunch of 'is X token in this array' checks that were getting unruly and replaces them with the token symbol + APR

## Type of change

## How should this be tested?

- Look at the APR tooltips of pools with token yield, ensure they look correct, include all tokens, and make sense. 

### Before:

2022-12-12-192554_310x251_scrot

2022-12-12-192719_288x306_scrot

### After

2022-12-12-192424_238x256_scrot

2022-12-12-192656_262x331_scrot

## Checklist:

### DIFF
--- a/src/components/tooltips/APRTooltip/APRTooltip.spec.ts
+++ b/src/components/tooltips/APRTooltip/APRTooltip.spec.ts
@@ -174,7 +174,7 @@ describe('APRTooltip', () => {
   });
 
   describe('Token Aprs', () => {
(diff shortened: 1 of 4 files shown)


pkattera (approved):
This is much better! Reviewed for functionality.

PR #2552, opened 2022-12-12
Soft page transition animations

pkattera:
# Description

This PR adds page transitions using opacity to soften the appearance as content fades in between pages. This includes the following pages:
- Home page
- Swap page
- Claims page
- Portfolio page
- veBAL page
- Invest flow modal
- Withdraw flow modal
- Migrate flow modal
- Terms of Service
- Privacy Policy
- Cookies Policy

The main change was wrapping content in vue transition tags:
`<Transition appear name="appear"> </Transition>`

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other: Softer page transition animations

## How should this be tested?

Make sure no pages have broken with the changes. Page content should appear with a soft fade in from opacity: 0 to opacity: 1. 

## Checklist:

### DIFF
--- a/src/components/contextual/pages/pool/invest/InvestPage.vue
+++ b/src/components/contextual/pages/pool/invest/InvestPage.vue
@@ -15,21 +15,23 @@ const { upToLargeBreakpoint } = useBreakpoints();
 </script>
 
 <template>
-  <div class="invest-page-layout-grid">
-    <div v-if="!upToLargeBreakpoint" class="col-span-5">
-      <InvestPageMyWallet />
-    </div>
+  <Transition appear name="appear">
+    <div class="invest-page-layout-grid">
+      <div v-if="!upToLargeBreakpoint" class="col-span-5">
+        <InvestPageMyWallet />
+      </div>
 
-    <div class="col-span-7">
-      <InvestPageInvestBlock />
-    </div>
+      <div class="col-span-7">
+        <InvestPageInvestBlock />
+      </div>
 
-    <InvestPageAccordion
-      v-if="upToLargeBreakpoint"
-      class="mt-4"
-      :isDeepPool="isDeepPool"
-    ></InvestPageAccordion>
-  </div>
+      <InvestPageAccordion
+        v-if="upToLargeBreakpoint"
+        class="mt-4"
+        :isDeepPool="isDeepPool"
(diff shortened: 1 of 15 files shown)


src/components/contextual/pages/pool/invest/InvestPage.vue
```diff
@@ -15,21 +15,23 @@ const { upToLargeBreakpoint } = useBreakpoints();
 </script>
 
 <template>
-  <div class="invest-page-layout-grid">
-    <div v-if="!upToLargeBreakpoint" class="col-span-5">
-      <InvestPageMyWallet />
-    </div>
+  <Transition appear name="appear">
```
pkattera, line 18: On pages like this, the only thing that changed was adding the <Transition></Transition> tags.

garethfuller (changes requested):
I believe this should only be wrapped around the components in src/pages and nothing else. I see you've done it around layouts but I don't think that will have any effect when you change pages because the layout mostly stays the same. You have also used in on the InvestPage.vue component which is not an actual page, sorry the naming is misleading.

pkattera:
@garethfuller 
> I believe this should only be wrapped around the components in src/pages and nothing else. I see you've done it around layouts but I don't think that will have any effect when you change pages because the layout mostly stays the same. 

The main reason for it was to apply the same animation for the Navigation. I ended up removing the Transition from defaultLayout and added it to the Navigation component (which lives outside the other pages). 

> You have also used it on the InvestPage.vue component which is not an actual page, sorry the naming is misleading.

I'd like for the first screens of the invest, withdraw and migrate pages/modals to also include the same soft transition. I've applied them to `migrate` and `withdraw` which are actual pages. It looks like the invest page is driven by `InvestPage`. The functionality works as intended using the transition here, but is there a better place for me to add it?

garethfuller:
@pkattera so for example in defaultLayout.vue you would replace `<router-view :key="$route.path" />` with:
```vue
<router-view v-slot="{ Component, route }">
  <transition name="appear">
    <component :is="Component" />
  </transition>
</router-view>
```
And then you can do the same BlankLayout, ContentLayout and FocussedLayout.

pkattera:
Closing this PR in favor of #2571

PR #2579, opened 2022-12-16
Skip balancer api repository when url is not defined

agualis:
# Description

When connected to Goerli, you can see the attached error when loading the index page.

The problem is that [goerli.json](https://github.com/balancer-labs/frontend-v2/blob/870d008f7062e3082cb446c895a8df61f4bab8ba/src/lib/config/goerli.json) does not have an attribute for balancerApi so that the SDK.

This solution will also work for `docker.json`.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:

### DIFF
--- a/src/composables/queries/usePoolsQuery.ts
+++ b/src/composables/queries/usePoolsQuery.ts
@@ -13,6 +13,7 @@ import {
   GraphQLArgs,
   PoolsFallbackRepository,
   PoolsRepositoryFetchOptions,
+  PoolRepository as SDKPoolRepository,
 } from '@balancer-labs/sdk';
 import { PoolDecorator } from '@/services/pool/decorators/pool.decorator';
 import { flatten } from 'lodash';
@@ -21,6 +22,7 @@ import { tokenTreeLeafs } from '../usePool';
 import { balancerSubgraphService } from '@/services/balancer/subgraph/balancer-subgraph.service';
 import { balancerAPIService } from '@/services/balancer/api/balancer-api.service';
 import { poolsStoreService } from '@/services/pool/pools-store.service';
+import { isBalancerApiDefined } from '@/lib/utils/balancer/api';
 
 type PoolsQueryResponse = {
   pools: Pool[];
@@ -45,8 +47,6 @@ export default function usePoolsQuery(
   const { injectTokens, tokens: tokenMeta, dynamicDataLoading } = useTokens();
   const { networkId } = useNetwork();
 
-  let balancerApiRepository = initializeDecoratedAPIRepository();
-  let subgraphRepository = initializeDecoratedSubgraphRepository();
   let poolsRepository = initializePoolsRepository();
 
   /**
@@ -54,12 +54,9 @@ export default function usePoolsQuery(
    */
 
   function initializePoolsRepository(): PoolsFallbackRepository {
-    const fallbackRepository = new PoolsFallbackRepository(
-      [balancerApiRepository, subgraphRepository],
-      {
-        timeout: 30 * 1000,
-      }
-    );
+    const fallbackRepository = new PoolsFallbackRepository(buildProviders(), {
+      timeout: 30 * 1000,
+    });
     return fallbackRepository;
   }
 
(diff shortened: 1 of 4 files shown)


timjrobinson:
@agualis I like the style updates. The balancer API is also used in usePoolsQuery so I updated it there too. I renamed + moved the isBalancerApiDefined check to a util function so it can be shared between them.

Tested this on goerli + mainnet and everything appears to work correctly. Noticed that console.log appears 5 times per individual pool page because usePoolQuery is called from so many different paths, this is something we can optimise in the future.

timjrobinson (approved):
If you think my changes are good and work for you feel free to merge. Your code LGTM.

agualis:
@timjrobinson Thanks Tim, your changes look good. 

I also noticed the 5X console.log and thought that it would remind us that we can do some optimisations in the future 👀

Just one small detail about the naming:  I did not want to use a "boolean function name" like `isBalancerApiDefined` because the function returns a boolean but also contains a side effect (console.log()). We can refactor following the advice in[ this post](https://www.teamten.com/lawrence/programming/keep-if-clauses-side-effect-free.html) or we can also keep it like this cause the side effect is not so important in this scenario. I'm ok with both decisions.

src/lib/utils/balancer/api.ts
```diff
@@ -0,0 +1,10 @@
+import { configService } from '@/services/config/config.service';
+
+export function isBalancerApiDefined() {
+  const defined = configService.network.balancerApi;
+  if (!defined)
+    console.log(
+      `Skipping balancer api provider in your current network (${configService.network.chainName})`
+    );
+  return defined;
+}
```
garethfuller, line 10: I don't think we need to log here and this should return a boolean.
```suggestion
export const isBalancerApiDefined = !!configService.network.balancerApi
```

src/composables/queries/usePoolsQuery.ts
```diff
@@ -104,6 +101,18 @@ export default function usePoolsQuery(
...
+    const providers: SDKPoolRepository[] = [];
+    if (isBalancerApiDefined()) {
+      const balancerApiRepository = initializeDecoratedAPIRepository();
+      providers.push(balancerApiRepository);
+    }
+    const subgraphRepository = initializeDecoratedSubgraphRepository();
+    providers.push(subgraphRepository);
+
+    return providers;
+  }
```
garethfuller, line 114: Lets stick to the naming convention already established.
```suggestion
  function buildRepositories() {
    const repositories: SDKPoolRepository[] = [];
    if (isBalancerApiDefined()) {
      const balancerApiRepository = initializeDecoratedAPIRepository();
      repositories.push(balancerApiRepository);
    }
    const subgraphRepository = initializeDecoratedSubgraphRepository();
    repositories.push(subgraphRepository);

    return repositories;
  }
```

src/services/pool/pool.repository.ts
```diff
@@ -57,4 +57,16 @@ export default class PoolRepository {
...
+    const providers: SDKPoolRepository[] = [];
+    if (isBalancerApiDefined()) {
+      const balancerApiRepository = this.initializeDecoratedAPIRepository();
+      providers.push(balancerApiRepository);
+    }
+    const subgraphRepository = this.initializeDecoratedSubgraphRepository();
+    providers.push(subgraphRepository);
+
+    return providers;
+  }
```
garethfuller, line 71: Can we try to define this function in one place?

garethfuller (changes requested):

src/services/pool/pool.repository.ts
agualis: @garethfuller It is not trivial so I suggest that @timjrobinson or me try to do it later in a separate PR.

garethfuller: Can usePoolsQuery not use `PoolRepository`? Is it doing something different?

agualis: The current code shows that `initializeDecoratedAPIRepository` and `initializeDecoratedSubgraphRepository` are slightly different. Probably refactoring to use PoolRepository in both places would avoid that duplication but I'm not so familiar with this code and that's why I check it carefully check it in a different PR.

timjrobinson: Yea I'd rather just get this merged and do that refactor later.

timjrobinson (approved):

src/services/pool/pool.repository.ts
agualis: @garethfuller If you agree, could you approve this PR so that we can merge?

garethfuller (approved):

PR #2582, opened 2022-12-20
Hotfix: Fallback to Subgraph if pool doesn't exist in Balancer API

timjrobinson:
# Description

Some pools are missing from the Balancer API, I haven't figured out why yet but when this happens the frontend should be falling back to the Subgraph instead of saying they don't exist. 

This adds to changes:
- Throw an error from pool repository if the single pool requested is not found, so that the fallback provider triggers and uses the Subgraph instead.
- Throw an error about the pool not existing if it doesn't exist as previously it would go into is blocked and fail that it "Can't read poolType of undefined" which is a red herring error. 

## Type of change

## How should this be tested?

- Test this pool that is missing from Balancer API but present in the Subgraph, ensure it loads correctly. - `/#/polygon/pool/0x2a60ec04577e0025d177a251d12d39af593e95ae0002000000000000000008cb`
- Test that a common pool that is in both Balancer API and Subgrpah loads correctly
- Test that a pool that doesn't exist throws an error in the console that "Pool doesn't exist" and the red banner appears at the top of the page. 
- 
## Visual context

## Checklist:


src/composables/queries/usePoolQuery.ts
```diff
@@ -79,6 +79,8 @@ export default function usePoolQuery(
       pool = await poolRepository.fetch(getQueryArgs());
     }
 
+    if (!pool) throw new Error('Pool does not exist');
```
alter-eggo, line 82: `await poolRepository.fetch(getQueryArgs());`  returns empty array if there is no pool or undefined?
shouldn't condition be like: `if (!pool?.length)`?

timjrobinson: The pool repository only returns a single pool, the individual fetchers return an array of pools.

alter-eggo: should we change return type of fetch function in poolRepository to
`public async fetch(queryArgs: GraphQLArgs): Promise<Pool | void> {` ?

garethfuller (approved):

PR #2584, opened 2022-12-20
Fix contract call settings error process

alter-eggo:
# Description
Now if there is any error during calling `settingsForContractCall` function, it returns `undefined`, this pr fixes it

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:

### DIFF
--- a/src/services/gas-price/gas-price.service.ts
+++ b/src/services/gas-price/gas-price.service.ts
@@ -66,22 +66,24 @@ export class GasPriceService {
   ): Promise<GasSettings> {
     let gasSettings: GasSettings = {};
 
-    const gasLimit = await contractWithSigner.estimateGas[action](
-      ...params,
-      options
-    ).catch(err => {
-      return Promise.reject(err.error);
-    });
-    gasSettings.gasLimit = this.formatGasLimit(gasLimit.toNumber());
-
-    if (this.shouldSetGasPriceSettings(options)) {
-      gasSettings = await this.setGasPriceSettings(
-        gasSettings,
-        forceLegacyTxType
+    try {
+      const gasLimit = await contractWithSigner.estimateGas[action](
+        ...params,
+        options
       );
-    }
+      gasSettings.gasLimit = this.formatGasLimit(gasLimit.toNumber());
 
-    return gasSettings;
+      if (this.shouldSetGasPriceSettings(options)) {
+        gasSettings = await this.setGasPriceSettings(
+          gasSettings,
+          forceLegacyTxType
+        );
+      }
+
+      return gasSettings;
+    } catch (err) {
+      return Promise.reject(err);
+    }
   }
 
   private formatGasLimit(limit: number): number {
--- a/src/services/gas-price/gas-price.service.ts
+++ b/src/services/gas-price/gas-price.service.ts
@@ -66,24 +66,20 @@ export class GasPriceService {
   ): Promise<GasSettings> {
     let gasSettings: GasSettings = {};
 
-    try {
-      const gasLimit = await contractWithSigner.estimateGas[action](
-        ...params,
-        options
+    const gasLimit = await contractWithSigner.estimateGas[action](
+      ...params,
+      options
+    );
+    gasSettings.gasLimit = this.formatGasLimit(gasLimit.toNumber());
+
+    if (this.shouldSetGasPriceSettings(options)) {
+      gasSettings = await this.setGasPriceSettings(
+        gasSettings,
+        forceLegacyTxType
       );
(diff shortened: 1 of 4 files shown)


src/services/gas-price/gas-price.service.ts
```diff
@@ -66,22 +66,24 @@ export class GasPriceService {
   ): Promise<GasSettings> {
     let gasSettings: GasSettings = {};
 
-    const gasLimit = await contractWithSigner.estimateGas[action](
-      ...params,
-      options
-    ).catch(err => {
-      return Promise.reject(err.error);
```
alter-eggo, line 73: here is the problem. there is no error in err

src/services/gas-price/gas-price.service.ts
```diff
@@ -66,22 +66,24 @@ export class GasPriceService {
...
+      if (this.shouldSetGasPriceSettings(options)) {
+        gasSettings = await this.setGasPriceSettings(
+          gasSettings,
+          forceLegacyTxType
+        );
+      }
+
+      return gasSettings;
+    } catch (err) {
+      return Promise.reject(err);
```
timjrobinson, line 85: Does this work exactly the same as not having a try/catch? As it's just re-throwing. I think the original implementation was trying to return err.message, but this works.

timjrobinson (approved):

src/services/gas-price/gas-price.service.ts
agualis: @alter-eggo I pushed a test to showcase this scenario. You can remove the try-catch and the test will still pass. 

IMO, we should avoid this try-catch until the higher level, the one that is responsible to process the exception to translate it to a human-friendly user message or to trigger a recovery mechanism. I would also expect proper tests to cover those cases. 

For instance, in this concrete case, we have this code in a higher level: 

https://github.com/balancer-labs/frontend-v2/blob/33ff0174b9ec950ca3b525661a4902ed5006a3ba/src/services/web3/transactions/concerns/contract.concern.ts#L51

, where we are assuming that we always get a `WalletError` which is dangerous. That's outside the scope of this bugfix but I would suggest refactoring this kind of error hanndling to [something like this](https://kentcdodds.com/blog/get-a-catch-block-error-message-with-typescript) (together with more new tests).

src/services/gas-price/gas-price.service.ts
garethfuller: Let's avoid changing the structure to try/catch and handle the fact that there might not be an error attribute. Something like:
```ts
return Promise.reject(err?.error || 'Failed to estimate gas');
```

garethfuller (changes requested):

src/services/gas-price/gas-price.service.ts
alter-eggo: what is the purpose of throwing `err?.error`. it seems that in most cases there won't be `error` field in `err` so, we will just throw a general message.

I agree with @agualis and it might be better to remove error handling here until the higher level

garethfuller: I think the reason is specifically to catch and throw the error message that comes from the contract failure. How do you know in most cases that the error attribute isn't there?

alter-eggo: hm, maybe I'm wrong, it just looks weird to me to put `error` field in error 🤔. 

for example in that specific case with RBW / UNIM pool (polygon), in withdraw flow if we add error process `return Promise.reject(err?.error || 'Failed to estimate gas');` we get:

Screenshot 2023-01-09 at 13 03 30

if we remove error process:
Screenshot 2023-01-09 at 13 01 41

I think second case is more helpful for us as provides more information. 

nevertheless, if we want to handle case with error attribute, I suggest:

`return Promise.reject(err?.error || error);`

src/services/gas-price/gas-price.service.ts
alter-eggo: thanks for your help here @agualis , I agree that it might be better to refactor here and created a task for this.

src/services/gas-price/gas-price.service.ts
garethfuller: Ok I see your point, I didn't know exactly how it was working either. Feel free to remove it and let the error be handled higher up the stack.

alter-eggo: ok. fixed

garethfuller (approved):

PR #2588, opened 2022-12-20
Fix approval flow with gnosis safe wallet

alter-eggo:
# Description
Change signature approval to tx approval if user uses gnosis safe wallet

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203639227157457


Matthews3301:
@alter-eggo I'm wondering, maybe it would be better to instead of modifying the array of actions, to modify the logic of the sign action.

E.g. in:
https://github.com/balancer-labs/frontend-v2/blob/3cdf0e26d6717dcd9df002c7cee00e13c38d3bdd/src/composables/useSignRelayerApproval.ts#L43

Make this a conditional logic, and if on gnosis, use `isSignAction: false`.

alter-eggo:
@Matthews3301 thanks for your help here! 
but seems it won't solve the problem, as it still will call the same action. `isSignAction` is used only for specific handling of sign action in `BalActionSteps` component.

Matthews3301:
@alter-eggo Sorry, that's what I mean: changing `isSignAction` parameter and also the action. It's more efficient to have it in `useSignRelayerApproval.ts` as it'll fix it in all places signing is currently used.

alter-eggo:
hm, yeah, makes sense, I will fix it

Matthews3301:
I tested both actions locally and they work 👍 Haven't tested in Gnosis Safe.

Matthews3301 (approved):

src/composables/useSignRelayerApproval.ts
```diff
@@ -49,6 +53,15 @@ export default function useSignRelayerApproval(relayerType: RelayerType) {
     isSignAction: true,
   };
 
+  function getAction() {
+    const provider = lsGet('connectedProvider');
```
garethfuller, line 57: Instead of getting this from local storage, can we get it from the active state in useWeb3?

garethfuller (changes requested):
Ahh I think this is completely wrong. There is some confusion because for the Cowswap interface in the Swap UI you need to approve the gnosis relayer. That has nothing to do with signing the Balancer batch relayer.

The logic should be:
```
if provider === 'gnosis'
    inject approval tx rather than signature (but for the same relayer)
```

src/composables/useSignRelayerApproval.ts
```diff
@@ -23,6 +24,9 @@ export default function useSignRelayerApproval(relayerType: RelayerType) {
   const { account, getSigner } = useWeb3();
   const { networkId } = useNetwork();
   const { t } = useI18n();
+  const { action: gnosisApprovalAction } = useRelayerApproval(
+    RelayerType.GNOSIS
```
Matthews3301, line 28: What if we change this to `Relayer`, i.e. the type of approval which is requested, not just Gnosis (Cowswap). This will include batch relayer.

Matthews3301:
@garethfuller would you mind checking now? I changed the logic.

garethfuller:
@Matthews3301 thanks, I also had a go at a larger refactor of approvals because they were all over the place. I:
- Moved all approval composables under one 'approvals' folder.
- Renamed `useRelayerApproval` > `useRelayerApprvalTx` since it was specifically for handling a tx approval.
- Renamed `useSignRelayerApproval` > `useRelayerApproval` since it actually handles both cases, e.g. signature + fallback.

garethfuller (approved):
I was able to add liquidity to a boosted pool on polygon via a Safe wallet. It asked me to approve the relayer with a tx rather than a signature as expected.

alter-eggo (commented):
thanks for your help here guys. from the code side lgtm

PR #2607, opened 2023-01-04
Rename and refactor deep join/exit handlers

garethfuller:
# Description

- Refactors DeepExitHandler and DeepJoinHandler classes.
- Renames DeepExitHandler > GeneralisedExitHandler and DeepJoinHandler > GeneralisedJoinHandler because it better describes what the handler is handling. The fact that it's used for deep pools is just one use case, it may be used for other pool types at some point.

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test exits that use the generalised exit function, e.g. proportional exits from deep pools.
- [ ] Test joins that use the generalised join function, e.g. proportional joins for deep pools.

## Visual context

n/a

## Checklist:

### DIFF
--- a/src/services/balancer/pools/exits/handlers/generalised-exit.handler.ts
+++ b/src/services/balancer/pools/exits/handlers/generalised-exit.handler.ts
@@ -16,21 +16,13 @@ import { flatTokenTree } from '@/composables/usePool';
 import { getAddress } from '@ethersproject/address';
 import { TransactionBuilder } from '@/services/web3/transactions/transaction.builder';
 
-interface GeneralisedExitResponse {
-  to: string;
-  callData: string;
-  tokensOut: string[];
-  expectedAmountsOut: string[];
-  minAmountsOut: string[];
-  priceImpact: string;
-}
+type ExitResponse = Awaited<ReturnType<typeof balancer.pools.generalisedExit>>;
 
 /**
- * Handles exits for single asset flows where we need to use a BatchSwap to exit
- * the pool.
+ * Handles exits using SDK's generalisedExit function.
  */
-export class DeepExitHandler implements ExitPoolHandler {
-  private lastGeneralisedExitRes?: GeneralisedExitResponse;
+export class GeneralisedExitHandler implements ExitPoolHandler {
+  private lastExitRes?: ExitResponse;
 
   constructor(
     public readonly pool: Ref<Pool>,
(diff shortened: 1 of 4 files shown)


src/services/balancer/pools/exits/handlers/generalised-exit.handler.ts
```diff
@@ -16,21 +16,13 @@ import { flatTokenTree } from '@/composables/usePool';
...
 
-interface GeneralisedExitResponse {
-  to: string;
-  callData: string;
-  tokensOut: string[];
-  expectedAmountsOut: string[];
-  minAmountsOut: string[];
-  priceImpact: string;
-}
+type ExitResponse = Awaited<ReturnType<typeof balancer.pools.generalisedExit>>;
```
timjrobinson, line 19: This is cool! Didn't know you can build types like this. I just had a read through all the Typescript utility types.

src/services/balancer/pools/exits/handlers/generalised-exit.handler.ts
```diff
@@ -55,40 +49,47 @@ export class DeepExitHandler implements ExitPoolHandler {
...
-    const allPoolTokens = flatTokenTree(this.pool.value);
+
+    this.lastExitRes = await balancer.pools.generalisedExit(
+      this.pool.value.id,
+      evmAmountIn.toString(),
+      signerAddress,
+      slippage,
+      relayerSignature
+    );
+    if (!this.lastExitRes) throw new Error('Failed to query exit.');
```
timjrobinson, line 69: It previously said "Not enough liquidity". Was that not true? Is there a way to tell the difference between a failed query and not enough liquidity? Does the user actually see this?

timjrobinson (approved):

src/services/balancer/pools/exits/handlers/generalised-exit.handler.ts
garethfuller: The only time 'Not enough liquidity' is relevant is for swap join/exits which are handled in the swap join/exit handlers. This error message looks like it was copied over without understanding why it was defined.

Yes, the user sees these error messages, so maybe 'Failed to query exit' is not perfect.

sssmi (approved):

PR #2608, opened 2023-01-04
Migrate from vue-query 1.x to @tanstack/vue-query 4.x

agualis:
# Description

`vue-query` v2 was released sometime ago.
Furthermore,  `vue-query` was migrated to be a part of https://github.com/TanStack/query. 

This PR migrates to the latest version, which involves this [breaking changes](https://vue-query-next.vercel.app/#/guides/migrating-to-vue-query-2): 

- `VueQueryDevTools` was removed. You can simply use official vue devtools instead.
- Callback methods like `refetch` or `fetchNextPage` are not refs anymore so no need to call them with `.value()`.
- Typescript types were improved. No more references to `react-query`.
- `idle` state was removed. Our use cases (no offline mode) just need `isLoading`, which simplifies logic. 
- Setup is done through a plugin which will simplify incoming global providers refactor.

## Type of change

- [x] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:


src/composables/queries/useGraphQuery.ts
```diff
@@ -63,5 +62,5 @@ export default function useGraphQuery<T>(
     }
   };
 
-  return useQuery(queryKey, queryFn, options);
+  return useQuery(queryKey, queryFn, options as any);
```
agualis, line 65: I did not find a way to make TS compiling the options type 🤷‍♂️
Maybe you can give it a try 🧙

src/providers/local/staking/userUserStakingData.ts
```diff
@@ -59,63 +57,15 @@ export type PoolGaugesResponse = {
   pools: PoolGaugesInfo[];
 };
 
-export type UserStakingDataResponse = {
```
agualis, line 62: [Using TS type inference with ReturnType](https://github.com/balancer-labs/frontend-v2/pull/2608/files#diff-431f5029dfdaefe2a611e7169e9fb4b62bbe00314e90efb9c8dfde20b5acffd0) avoids manually maintaining this complex type. The only drawback is that I had to move the related comments close to the related declarations. In any case, most of those comments should not be needed in case of proper naming.

src/composables/queries/useGraphQuery.ts
sssmi: How's this:
```suggestion
  return useQuery<T>(queryKey, queryFn, options);
```

agualis: I forgot that useQuery was expecting a generic.  Thanks!

src/providers/local/staking/userUserStakingData.ts
```diff
@@ -307,7 +264,7 @@ export default function useUserStakingData(
     () => isWalletReady.value && userGaugeShares.value.length > 0 && !isL2.value
   );
 
-  const { data: poolBoosts, isLoading: isLoadingBoosts } = useQuery<
+  const { data: poolBoosts, isInitialLoading: isLoadingBoosts } = useQuery<
```
sssmi, line 267: Have to be careful with queries that are not always enabled like this one here. Their `isLoading` state stays `false` until they are enabled. This caused the `StakePreview.vue` to show perpetual loading block.

Screenshot 2023-01-05 at 14 55 35

Checking the `isInitialLoading` instead works here. https://tanstack.com/query/v4/docs/react/guides/disabling-queries

agualis: I didn't catch that braking change in the docs. I think that we will have more cases like this one.

simeonkerkola (approved):
This is a good upgrade. Seems to be working fine 👍

agualis:
@garethfuller @simeonkerkola I fixed conflicts and updated the queries in the staking refactor. I think that we should try to merge this before we add/update more queries but we should: 

1. Double check that we don't need more adjustments like the ones Simeon did in [this commit](https://github.com/balancer-labs/frontend-v2/pull/2608/commits/eb739bfa13c509eecf792ef14b45c7499c78ea94) 
2. Fix `tokens.provider` tests as they were affected by some refactor in the related `dynamicDataLoaded.value` logic.

Could you help me with this 2 points as you are more familiar with this?

garethfuller:
Closing in favour of #2835

PR #2637, opened 2023-01-11
Feat: Gnosis chain

Matthews3301:
# Description

Support for Gnosis Chain

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:


src/services/balancer/pools/weighted-pool.service.ts
```diff
@@ -38,6 +46,12 @@ export interface JoinPoolRequest {
 }
 
 export default class WeightedPoolService {
+  version: FactoryVersion;
+
+  constructor(public readonly _version: FactoryVersion) {
+    this.version = _version;
+  }
```
garethfuller, line 53: When you define as `public readonly` it's automatically available on the instance. It's basically short hand for defining within the constructor.
```suggestion
  constructor(public readonly version: FactoryVersion) {}
```

src/services/balancer/pools/weighted-pool.service.ts
```diff
@@ -48,29 +62,43 @@ export default class WeightedPoolService {
...
+            owner,
+          ]
+        : [
+            name,
+            symbol,
+            tokenAddresses,
+            seedTokens,
+            swapFeeScaled.toString(),
+            owner,
+          ];
```
garethfuller, line 96: Does V3 not expect the same params as V2? Is it not only V1 that doesn't accept the rateProviders?

garethfuller (changes requested):

src/services/balancer/pools/weighted-pool.service.ts
Matthews3301: Updated

Matthews3301:
@garethfuller shall we merge this in? I'm still trying to verify to API, however, this codebase should work regardless.

Maybe removing the Gnosis-chain option from the dropdown but retaining url for testing?

garethfuller:
@Matthews3301 no not yet, I'm still reviewing and I think we need to make some changes to the create a pool stuff.

Matthews3301:
Continued in https://github.com/balancer-labs/frontend-v2/pull/2903/files

PR #2638, opened 2023-01-11
Add filtering to veBal table

alter-eggo:
# Description
This pr adds filtering to veBal table by token, network and expiring of gauges. 

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:


garethfuller:
Can we get the modules on the right to align with the bottom please?
Screenshot 2023-01-13 at 16 31 47

src/components/contextual/pages/vebal/LMVoting/GaugesFilters.vue
```diff
@@ -0,0 +1,105 @@
...
+            {{ $t('gaugeFilter.moreFilters') }}
+          </div>
+        </div>
+      </div>
+    </template>
+    <template #default="{ close }">
+      <div
+        class="flex overflow-hidden flex-col py-4 px-3 w-64 rounded-lg"
+        @click="close"
+      >
```
garethfuller, line 56: Please use BalBtn for this. The style might not be exactly the same as Pon's designs but we have to stay consistent with buttons.

src/components/inputs/TokenSearchTextInput.vue
```diff
@@ -0,0 +1,33 @@
...
+  >
+    <BalIcon name="search" size="lg" class="mr-4" />
+    <div class="flex flex-wrap gap-2 items-center">
+      <input
+        class="dark:bg-gray-800"
+        :value="modelValue"
+        type="text"
+        :placeholder="placeholder"
+        @input="handleInput"
+      />
```
garethfuller, line 29: Please use BalTextInput for this, again we need to keep design consistent, that is why these Bal components exist.

garethfuller (changes requested):

src/components/contextual/pages/vebal/LMVoting/GaugesFilters.vue
garethfuller: If @pkattera want's to change the style of buttons we have to apply that to BalBtn and therefore everywhere.

alter-eggo:
@garethfuller 
Screenshot 2023-01-16 at 12 59 30

it's already aligned to bottom. or you mean it's better to put it in one line with filters?

garethfuller:
> it's already aligned to bottom. or you mean it's better to put it in one line with filters?

Yes please align it with the filters, unless @pkattera thinks otherwise. But aligning with the bottom of the title looks weird to me.

alter-eggo:
@garethfuller I've done the fixes.
I used `BalBtn` and `BalTextInput` components, and deleted unnecessary `TokenSearchTextInput` component. 
info modules are now align with filters, I think looks better now, good catch! thanks for the review

src/components/_global/BalCheckbox/BalCheckbox.vue
```diff
@@ -1,13 +1,13 @@
 <template>
   <div :class="['bal-checkbox', wrapperClasses]">
     <div class="flex">
-      <div class="flex items-start">
+      <div class="flex items-center">
```
garethfuller, line 4: This has implications in other areas when the checkbox label has a large amount of text. For example in high price impact joins, see screenshot:
Screenshot 2023-01-16 at 12 07 01

Please modify BalCheckbox to take a prop like "align" with a default value of center but that allows for the checkbox to align to the top as well.

src/components/contextual/pages/vebal/LMVoting/LMVoting.vue
```diff
@@ -72,6 +83,70 @@ const hasExpiredLock = computed(
...
+});
+
+const debouncedFilterText = computed({
+  get() {
+    return tokenFilter.value;
+  },
+  set: debounce(newValue => {
+    tokenFilter.value = newValue;
+  }, 500),
+});
```
garethfuller, line 108: Not a fan of get/set computed properties, it's kind of an anti-pattern, isn't it? Is there an alternative pattern that achieves the same result without computed properties, if so what does it look like?

garethfuller (changes requested):
Looks great, a few minor things.

src/components/_global/BalCheckbox/BalCheckbox.vue
garethfuller: If the default is center you will have to find all the price impact use cases and add the prop to align with top.

src/components/contextual/pages/vebal/LMVoting/LMVoting.vue
```diff
@@ -7,18 +7,29 @@ import useVotingEscrowLocks from '@/composables/useVotingEscrowLocks';
...
  */
+const tokenFilter = ref('');
+const hideExpiredGauges = ref(true);
+const activeNetworkFilters = ref<Network[]>([]);
 const activeVotingGauge = ref<VotingGaugeWithVotes | null>(null);
+const networks = {
+  [Network.MAINNET]: 'Ethereum',
+  [Network.POLYGON]: 'Polygon',
+  [Network.ARBITRUM]: 'Arbitrum',
+};
```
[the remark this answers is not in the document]
garethfuller: It also needs to include Optimism.

pkattera:
@alter-eggo A few tweaks:
Screenshot 2023-01-16 at 12 22 20 PM

src/components/contextual/pages/vebal/LMVoting/LMVoting.vue
alter-eggo: hmm, it seems that we can use `customRef` https://vuejs.org/api/reactivity-advanced.html#customref
I will try

src/components/_global/BalCheckbox/BalCheckbox.vue
alter-eggo: fixed

src/components/contextual/pages/vebal/LMVoting/LMVoting.vue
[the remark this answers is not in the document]
alter-eggo: moved to useNetwork

src/components/contextual/pages/vebal/LMVoting/LMVoting.vue
alter-eggo: yeah, I used `customRef` and it works in the same way

alter-eggo:
@garethfuller @pkattera fixed all the comments, except one by Pon:
if some filter is active I just change the color of `More filters` text and icon.
the problem is that the filter applying is debounced, so number changes only after some time (500ms in our case). it’s not easy to fix and will lead to a lot of overhead logic

src/components/_global/BalCheckbox/BalCheckbox.vue
```diff
@@ -89,7 +90,7 @@ export default defineComponent({
         case 'lg':
           return 'w-6 h-6';
         default:
-          return 'w-5 h-5 mt-1';
```
garethfuller, line 92: This `mt-1` and the `relative top-2` above in the diff were probably required to make the items-start look right. Please conditionally add these in if the alignment is to the top.

src/composables/useNetwork.ts
```diff
@@ -27,6 +27,13 @@ const NETWORK_ID =
 if (windowAvailable) localStorage.setItem('networkId', NETWORK_ID.toString());
 export const networkSlug = config[NETWORK_ID].slug;
 export const networkConfig = config[NETWORK_ID];
+export const networksMap = {
```
garethfuller, line 30: Since lots of things could be a 'networksMap' let's rename it 'networkLabelMap' to be more specific.

garethfuller (changes requested):
> so number changes only after some time (500ms in our case)

Even if the number takes a while to update I think it would be worth adding. Is it really difficult just to track active filters even if it's not technically accurate for some time will the debounce functions happen in the background?

garethfuller (approved):

PR #2649, opened 2023-01-13
Add frxETH / WETH polygon to allowlist

bogdoslavik:
frxETH / WETH polygon whitelist


garethfuller:
Closing in favour of #2652, had to bump version again and lock file needs to be updated too.

PR #2664, opened 2023-01-17
Hotfix: gnosis zero trade value handle

alter-eggo:
# Description
This pr fixes the bug on swap page, when user tries to input, e.g. `0.01` in `gnosis` trade flow. Now this value will automatically set to 0. 

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer-labs/frontend-v2",
-  "version": "1.85.1",
+  "version": "1.85.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@balancer-labs/frontend-v2",
-      "version": "1.85.1",
+      "version": "1.85.2",
       "license": "MIT",
       "devDependencies": {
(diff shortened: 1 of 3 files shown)


garethfuller (changes requested):
The problem with this fix is that it doesn't capture `0.0` for example, then you get a `ZeroAmount` error message in the UI.

I also don't think the logic was correct in the first place, if it's an `exactIn` case then we shouldn't be force overwriting tokenInAmount to 0 (and vice versa for `exactOut`), because the user has just entered this value.

I think the logic should be something like:
```
if (amountToExchange.isZero() && exactIn.value) {
  tokenOutAmountInput.value = '0';
  return;
} else if (amountToExchange.isZero() && exactOut.value) {
  tokenInAmountInput.value = '0';
  return;
}

alter-eggo:
@garethfuller fixed

garethfuller (approved):

PR #2684, opened 2023-01-20
Hotfix: Reduce the number of Tenderly API calls

simeonkerkola:
# Description

Call queryExit and queryJoin less often to reduce the nmbr of calls to Tenderly api

1. When user is changing input amounts queries are debounced, but first query will not fire immediately (`leading` is off)
2. `refetchOnWindowFocus` for queries is disabled, but queries still refetch in the background if Preview Modal is open.
3. Prevent invest form rerender if user tokens array hasn't changed

## Type of change

- [x] Other

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer-labs/frontend-v2",
-  "version": "1.85.5",
+  "version": "1.85.6",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@balancer-labs/frontend-v2",
-      "version": "1.85.5",
+      "version": "1.85.6",
       "license": "MIT",
       "devDependencies": {
         "@aave/protocol-js": "^4.3.0",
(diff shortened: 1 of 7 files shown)


timjrobinson:
I'm seeing some weird glitch where numbers keep being reset to 0 after a few seconds (maybe after the simulation returns). It only happens when manually entering and not when clicking max.

timjrobinson:
Looks like the glitch is also happening in prod so may be unrelated. The calls to tenderly have reduced so that part LGTM.

timjrobinson:
Yea it seems to reset all the values on re-focus. Could you look into that, I thought it wouldn't be refetching on focus after this change.

simeonkerkola:
> I'm seeing some weird glitch where numbers keep being reset to 0 after a few seconds (maybe after the simulation returns). It only happens when manually entering and not when clicking max.

Thanks. I noticed it too now.
When wallet balances update, that might cause a rerender. It's not related to this, so I'm going to fix that in a separate PR.

timjrobinson (approved):
Sounds good. This reduces the calls so LGTM

PR #2686, opened 2023-01-20
hotfix: cowswap re-enabled, using new endpoint

Matthews3301:
# Description

CoWswap has depreciated `market` endpoint which we used to get gassless trade quotes. This has been replaced with POST `quote` endpoint. Posting orders remains unchanged as far as I can tell.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:


src/services/gnosis/gnosisProtocol.service.ts
```diff
@@ -131,20 +131,19 @@ export default class GnosisProtocolService {
...
-      )}`;
-
-      const response = await axios.get<PriceInformation>(
-        `${this.baseURL}/markets/${market}/${kind}/${amount}`
+      const response = await axios.post<CowSwapQuoteResponse>(
+        `${this.baseURL}/quote`,
+        {
+          sellToken,
+          buyToken,
+          from: sellToken,
```
Matthews3301, line 139: I'm not sure what this parameter is for. I can use user account but we also want quotes when wallet is not connected.

src/composables/trade/useGnosis.ts
```diff
@@ -350,9 +351,15 @@ export default function useGnosis({
             if (
               priceQuote.status === 'fulfilled' &&
               priceQuote.value &&
-              priceQuote.value.amount != null
+              (exactIn.value
```
timjrobinson, line 354: This could be simplified to `const amount = exactIn.value ? priceQuote.value.buyAmount : priceQuote.value.sellAmount` and use that in both places.

timjrobinson (approved):

timjrobinson:
Tested and the quote information in the network requests seems to be correct.

src/composables/trade/useGnosis.ts
Matthews3301: Thanks, refactored this entire block of code.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@balancer-labs/frontend-v2",
-  "version": "1.86.1",
+  "version": "1.86.2",
```
garethfuller, line 3: Need to bump again.

src/composables/trade/useCowswap.ts
```diff
@@ -33,7 +26,7 @@ import { Goals, trackGoal } from '../useFathom';
 
 const HIGH_FEE_THRESHOLD = parseFixed('0.2', 18);
 const APP_DATA =
-  import.meta.env.VITE_GNOSIS_APP_DATA ??
+  process.env.VUE_APP_GNOSIS_APP_DATA ??
```
garethfuller, line 29: This will break Vite. Why convert it back to proces.env?

garethfuller (changes requested):

src/composables/trade/useCowswap.ts
Matthews3301: Sorry, my bad. Fixed.

garethfuller (changes requested):
https://www.loom.com/share/bec9411ad9f14ac586a538a6435dd284

There's an issue when you change input values quickly and the output values lag and then cycle through all the previous input value requests. @simeonkerkola did we use debounce in the join/exit flows to prevent this type of thing for price impact?

Matthews3301:
@garethfuller the reason for this is that we were fetching quote after each digit was typed. I added a small timeout to prevent spamming the endpoints whilst typing.

Matthews3301:
@garethfuller could this work? https://github.com/balancer-labs/frontend-v2/pull/2686/commits/bd632c5fe731759bb2840ec065d5153f7b0d8dce

garethfuller (approved):

PR #2707, opened 2023-01-25
Refactor: Avoid defineComponent in staking provider

agualis:
# Description

Instead of using a `<StakingProvider>` wrapper component, the current PR follows the same pattern that we used in "global providers".

`porfolio.vue` provides the needed provider with` provideStaking()` from its setup. 

Children components can `import { useStaking } from '@/providers/local/staking/staking.provider';` to inject the provided component. 

The new pattern is a bit simpler but also much easier to unit test. 

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:


src/components/contextual/pages/pools/UnstakedPoolsTable.vue
```diff
@@ -155,16 +153,12 @@ function handleModalClose() {
         @trigger-stake="handleStake"
       />
     </BalStack>
-    <StakingProvider
+    <StakePreviewModal
       v-if="stakePool"
-      :poolAddress="getAddressFromPoolId(stakePool.id)"
```
agualis, line 160: If I'm not wrong, `stakePool.id` will be undefined until the `stakePool` is created after calling   `setPoolAddress(pool.address)`:

```ts
 handleStake(pool: Pool) {
  setPoolAddress(pool.address);
  showStakeModal.value = true;
  stakePool.value = pool;
} 
```

That means that we can avoid initialising poolAddress when calling `StakingProvider`. Am I missing something?

garethfuller (commented):
Actually, I don't think there is anything I want to include here. I thought you might suggest some change to using defineComponent in PoolStakingProvider but I don't see it.

agualis:
> Actually, I don't think there is anything I want to include here. I thought you might suggest some change to using defineComponent in PoolStakingProvider but I don't see it.

Yeah, by mistake, I refactored `StackingProvider` instead of `PoolStakingProvider`. I can do it with `PoolStakingProvider` in a new PR or we can do it afterwards. Whatever you prefer.

PR #2709, opened 2023-01-26
Feat: My pool share composition

agualis:
# Description

Adds "My pool share" tab to composition table.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

my-pool-share

## Checklist:


pkattera:
Looks good to me

garethfuller:
To calculate these percentages could you please use the token values combined as the total in the calculation rather than the `pool.totalLiquidity` because sometimes they're slightly different which can cause odd percentages here.
Screenshot 2023-01-27 at 11 13 46

timjrobinson:
It's not working for me for the veBAL pool. Is it supposed to? (this is for timjrobinson.eth)

poolcomposition

src/components/contextual/pages/pool/PoolCompositionCard/components/composables/useTokenBreakdown.ts
```diff
@@ -1,66 +1,138 @@
...
+  const isDeepPool = computed(() => isDeep(rootPool.value));
+  let tokensData: TokensData = {};
+  // Sum of every token's fiat value.
+  // Used to increase token percentage accuracy as pool.totalLiquidity is not completely in sync with balance * coingecko-price calculations).
+  let totalFiat = 0;
+
+  // Recalculates recursive tokensData whenever the pool is re-fetched
+  return computed(() => calculateAllTokensData(rootPool));
+
+  function calculateAllTokensData(pool: Ref<Pool>) {
```
timjrobinson, line 33: This code is good, but I think it could be moved to the SDK (later) as it would be useful to have a `getUserBalances` function on a pool that takes a user address and returns all the details about the investment they have in the pool. @gmbronco

timjrobinson (approved):
LGTM - just the one issue with veBAL pool (not sure if this is fixable becuase the BPT are locked?) and that a lot of the calculation code should be in the SDK instead of frontend.

src/components/contextual/pages/pool/PoolCompositionCard/components/composables/useTokenBreakdown.ts
agualis: > LGTM - just the one issue with veBAL pool (not sure if this is fixable becuase the BPT are locked?) and that a lot of the calculation code should be in the SDK instead of frontend.

It is interesting that it does not show the tab because I use this query:

`  const { stakedShares } = usePoolStaking();
`

https://github.com/balancer-labs/frontend-v2/pull/2709/files#diff-9dd467c6218b8ac7397d3cee3fcddca3e89a1faacceb00ac07e0a4fd6a749075R10

That it's returning 0 for `timjrobinson.eth` in this pool (http://localhost:8080/#/ethereum/pool/0x5c6ee304399dbdb9c8ef030ab642b10820db8f56000200000000000000000014)

@timjrobinson or @garethfuller Do you know why?

garethfuller: For this specific pool you don't stake your shares, you lock them, so they don't feature in `stakedShares`. You need to use `useLock`. And the value you're looking for is `lock.value.lockedAmount` https://github.com/balancer-labs/frontend-v2/blob/develop/src/composables/useLock.ts#L53.

agualis: Ok, that makes sense. I think that we could merge this current branch and creating an independent task for this (Showing user locked shares in composition table).

garethfuller (changes requested):
Why is package-lock.json changing but package.json isn't?

agualis:
> Why is package-lock.json changing but package.json isn't?

Probably a bad merge conflict resolution. I got the version from develop and run `npm i` and everything looks ok now.

garethfuller (approved):

PR #2740, opened 2023-01-31
Hotfix: SOR issues

garethfuller:
# Description

Fixes a variety of issues with the swap UI.
- [fix: Prevent join/exit overwriting sor output if no returnAmount](https://github.com/balancer-labs/frontend-v2/commit/a413819263f741f459e28e3cb9bdbd4c2aa75d8e)
- [fix: Handle No Liquidity errors directly in useSOR](https://github.com/balancer-labs/frontend-v2/commit/6fdddea66bbebb8533d2427d7d6f5abadb4117cd)

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test you don't see 'No liquidity' validation errors when changing amounts quickly.]
- [ ] Test swaps work as expected

## Visual context

n/a

## Checklist:

### DIFF
--- a/src/composables/trade/useJoinExit.ts
+++ b/src/composables/trade/useJoinExit.ts
@@ -163,13 +163,16 @@ export default function useJoinExit({
     const tokenInDecimals = getTokenDecimals(tokenInAddressInput.value);
     const tokenOutDecimals = getTokenDecimals(tokenOutAddressInput.value);
 
-    if (exactIn.value) {
+    const returnAmount = swapInfo.value?.returnAmount || BigNumber.from('0');
+    const hasReturnAmount = swapInfo.value?.returnAmount.gt(0);
+
+    if (exactIn.value && hasReturnAmount) {
       tokenOutAmountInput.value = bnum(
-        formatUnits(swapInfo.value?.returnAmount ?? '0', tokenOutDecimals)
+        formatUnits(returnAmount, tokenOutDecimals)
       ).toFixed(6, OldBigNumber.ROUND_DOWN);
-    } else {
+    } else if (hasReturnAmount) {
       tokenInAmountInput.value = bnum(
-        formatUnits(swapInfo.value?.returnAmount ?? '0', tokenInDecimals)
+        formatUnits(returnAmount, tokenInDecimals)
       ).toFixed(6, OldBigNumber.ROUND_DOWN);
     }
   }
--- a/src/composables/trade/useJoinExit.ts
+++ b/src/composables/trade/useJoinExit.ts
@@ -164,13 +164,14 @@ export default function useJoinExit({
     const tokenOutDecimals = getTokenDecimals(tokenOutAddressInput.value);
 
     const returnAmount = swapInfo.value?.returnAmount || BigNumber.from('0');
-    const hasReturnAmount = !returnAmount.isZero();
 
-    if (exactIn.value && hasReturnAmount) {
+    if (returnAmount.isZero()) return;
+
+    if (exactIn.value) {
       tokenOutAmountInput.value = bnum(
         formatUnits(returnAmount, tokenOutDecimals)
       ).toFixed(6, OldBigNumber.ROUND_DOWN);
-    } else if (hasReturnAmount) {
+    } else {
       tokenInAmountInput.value = bnum(
         formatUnits(returnAmount, tokenInDecimals)
       ).toFixed(6, OldBigNumber.ROUND_DOWN);
--- a/src/composables/trade/useJoinExit.ts
+++ b/src/composables/trade/useJoinExit.ts
@@ -164,7 +164,7 @@ export default function useJoinExit({
     const tokenOutDecimals = getTokenDecimals(tokenOutAddressInput.value);
 
     const returnAmount = swapInfo.value?.returnAmount || BigNumber.from('0');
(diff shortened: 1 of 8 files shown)


src/composables/trade/useJoinExit.ts
```diff
@@ -163,13 +163,16 @@ export default function useJoinExit({
...
+    const returnAmount = swapInfo.value?.returnAmount || BigNumber.from('0');
+    const hasReturnAmount = returnAmount.gt(0);
+
+    if (exactIn.value && hasReturnAmount) {
       tokenOutAmountInput.value = bnum(
-        formatUnits(swapInfo.value?.returnAmount ?? '0', tokenOutDecimals)
+        formatUnits(returnAmount, tokenOutDecimals)
       ).toFixed(6, OldBigNumber.ROUND_DOWN);
-    } else {
+    } else if (hasReturnAmount) {
```
Matthews3301, line 173:
```suggestion
    } else {
```

src/composables/trade/useJoinExit.ts
```diff
@@ -163,13 +163,16 @@ export default function useJoinExit({
     const tokenInDecimals = getTokenDecimals(tokenInAddressInput.value);
     const tokenOutDecimals = getTokenDecimals(tokenOutAddressInput.value);
 
-    if (exactIn.value) {
+    const returnAmount = swapInfo.value?.returnAmount || BigNumber.from('0');
+    const hasReturnAmount = returnAmount.gt(0);
+
+    if (exactIn.value && hasReturnAmount) {
```
Matthews3301, line 169:
```suggestion
    if (exactIn.value) {
```

src/composables/trade/useJoinExit.ts
```diff
@@ -163,13 +163,16 @@ export default function useJoinExit({
     const tokenInDecimals = getTokenDecimals(tokenInAddressInput.value);
     const tokenOutDecimals = getTokenDecimals(tokenOutAddressInput.value);
 
-    if (exactIn.value) {
+    const returnAmount = swapInfo.value?.returnAmount || BigNumber.from('0');
+    const hasReturnAmount = returnAmount.gt(0);
+
```
Matthews3301, line 168:
```suggestion
if (!hasReturnAmount) {
  return;
}
```

src/composables/trade/useSor.ts
```diff
@@ -245,17 +245,28 @@ export default function useSor({
...
             : '';
         }
 
         if (swapType === SwapType.SwapExactIn) {
+          tokenOutAmount = await mutateAmount({
+            amount: tokenOutAmount,
+            address: tokenOutAddressInput.value,
+            isTokenIn: false,
+          });
+
```
garethfuller, line 269: In this function, if the in or out token is stETH the deltas returned the out token is actually the wstETH amount and so needs to be converted using our mutateAmount function.

src/composables/trade/useSor.ts
```diff
@@ -365,11 +376,6 @@ export default function useSor({
         priceImpact.value = 0;
         state.validationErrors.noSwaps = true;
       } else {
-        tokenOutAmount = await adjustedPiAmount(
-          tokenOutAmount,
-          tokenOutAddress
-        );
```
garethfuller, line 367: I believe the `swapReturn.returnAmount` which is `tokenOutAmount` here used to be the wstETH value that then needed to be converted for the price impact calculation. It now seems that the swapReturn amount is actually correct and so the conversion must be being handled in the SDK.

Matthews3301:
@garethfuller hmm, there still seems to be some issue with stETH -> ETH swap.

To reproduce:
1. Set tokens, input exactOut (e.g. 4 ETH)
2. InputIn will be quoted and set.
3. Wait for the next block.
4. InputIn amount will now change by a lot with 10% price impact.

Either the initial or recalculated value is incorrect.

src/composables/trade/useSor.ts
```diff
@@ -245,17 +245,28 @@ export default function useSor({
           tokenOutAddress.toLowerCase()
         );
 
-        const tokenInAmount = BigNumber.from(deltas[tokenInPosition]).abs();
-
-        const tokenOutAmount = BigNumber.from(deltas[tokenOutPosition]).abs();
+        let tokenInAmount = BigNumber.from(deltas[tokenInPosition]).abs();
+        let tokenOutAmount = BigNumber.from(deltas[tokenOutPosition]).abs();
 
         if (swapType === SwapType.SwapExactOut) {
```
mikemcdonald, line 251: I don't think these two if statements are needed. Otherwise a exact in stETH trade amount won't be mutated.

src/composables/trade/useSor.ts
```diff
@@ -245,17 +245,28 @@ export default function useSor({
...
+          tokenInAmount = await mutateAmount({
```
mikemcdonald, line 252: I don't think these two if statements are needed. Otherwise a exact in stETH trade amount won't be mutated.

garethfuller: This part is about mutating the output token only. It's called every block to make sure the display number for the amount you are not specifying is kept up to date. I missed performing the mutations on the values for the priceImpact further down in the `handleAmountChange` function, which catches the case you are talking about.

garethfuller:
@Matthews3301 @mikemcdonald This should be fixed now.

PR #2749, opened 2023-02-01
Handle generalisedJoin update that adds ability to choose different simulation types

brunoguerios:
# Description

Update generalised-join handler to support SDK update that enables different simulation types:
- Tenderly: keeps current behavior of simulating via Tenderly
- VaultModel: simulates using TS math from the SOR (above 99% accuracy, but not exact match)
- Static: simulates via staticCall and provides exact match, but requires vault approval

Current change simply defaults to Tenderly simulation - other options should be supported on a different PR.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] generalisedJoin should keep working as is

## Checklist:


simeonkerkola (approved):
Thanks @brunoguerios ! This seems to work now as it is. Should we merge this now, and then I can start working on with the update to VaultModel in a different PR?

brunoguerios:
Yes, I believe the best approach would be to merge this and create a new PR with the necessary changes to start using the VaultModel + Static simulations.

PR #2776, opened 2023-02-08
feat: Generate TODO file from comments

garethfuller:
# Description

Generates a TODO.md file on every push from TODO comments in code.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

n/a

## Visual context

n/a

## Checklist:


garethfuller:
Not sure if this is a good idea but wanted to test it out in case it was useful so we don't forget parts of the code that need fixed.

simeonkerkola (approved):
I like it.

alter-eggo (approved):
nice 👍 easier to understand where we can improve the codebase now

agualis (commented):
Can we somehow add a check to limit the max number of TODOs? 
I think it can be useful if we keep the TODO list small to make sure that we frequently address them.

garethfuller:
@agualis Do you have any idea how that could work? Just limit the number injected into the file?

simeonkerkola:
Can the `// TODO` regex be modified so it matches both `// TODO` and `// TODO:`? I think a comment with colon is the more common way.

agualis:
> 
Adding this at the beginning of the generated file and making sure that it is included in eslint:

`/* eslint max-lines: ["error", {"max": 50, "skipBlankLines": true}] */
`

garethfuller:
> > 
> 
> Adding this at the beginning of the generated file and making sure that it is included in eslint:
> 
> `/* eslint max-lines: ["error", {"max": 50, "skipBlankLines": true}] */ `

Not sure if this will achieve what you're looking for, right? It will just cause checks to fail in any PR once you cross that max. Which would force someone to remove TODOs to get a PR to pass checks. In reality devs will probably just delete todos...

agualis:
> Not sure if this will achieve what you're looking for, right? It will just cause checks to fail in any PR once you cross that max. Which would force someone to remove TODOs to get a PR to pass checks. In reality devs will probably just delete todos...

If a dev deletes a TODO then another dev will see it in the PR review. If the TODO was not so important it is ok to delete it. IMO The important think is that it reminds the team to review the TODO list from time to time (if they just add new TODOs but they don't clean old ones).

garethfuller:
> 
> If a dev deletes a TODO then another dev will see it in the PR review. If the TODO was not so important it is ok to delete it. IMO The important think is that it reminds the team to review the TODO list from time to time (if they just add new TODOs but they don't clean old ones).

Ok I guess that might work. I can see us just constantly incrementing the max though to make PRs pass. 😂,

garethfuller:
> Can the `// TODO` regex be modified so it matches both `// TODO` and `// TODO:`? I think a comment with colon is the more common way.

@simeonkerkola agreed, I'll try to improve it so it's more robust before I merge.

agualis:
> Ok I guess that might work. I can see us just constantly incrementing the max though to make PRs pass. 😂,

Like when you enforce 80% testing coverage and devs add tests with empty assertions to hack the metric 😄

I think that it can work if we are a bit disciplined. If not, why having TODOs at all if you are not taking care of them 😇

garethfuller:
Not sure if auto-commits are the best approach here. Closing and will reconsider how to approach this.

PR #2789, opened 2023-02-09
Test improvements: MSW, Old vs New Multicaller and fetch error traces 

agualis:
# Description

Includes new tests and several refactors to make the code more testable and debuggable:

- New `handleFetchException` helper to use console.trace instead of console.error to get the real line where the exception is originated: 
console trace
- New eslint rule to enforce using DI for "Old Multicaller" (we will soon do the same with the "new Multicaller")
- Separation of REST and Graphql MSW handlers 
- Added `name` parameter to add names to Grapql queries so that they can be intercepted by MSW 

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:


src/services/balancer/subgraph/entities/poolShares/index.ts
```diff
@@ -14,7 +14,8 @@ export default class PoolShares {
   }
 
   public async get(args = {}, attrs = {}): Promise<PoolShare[]> {
-    const query = this.query(args, attrs);
+    const queryName = 'PoolShares';
+    const query = this.query(args, attrs, queryName);
     const data = await this.service.client.get(query);
```
agualis, line 19: In this type of manual graphql clients` client.get` returns `Promise<any>` and receives `Record<string, any>`, which makes it unsafe and less testable. 

I know that we have some complex queries where we do complex stuff like decorating but, I think that we could improve other queries like this ones by using a TS typed Graphql client like shown in this spike: 
https://github.com/balancer-labs/frontend-v2/pull/2380

 @garethfuller Do we talk about this?

agualis: Also, we have similar logic in balancer-subgraph.client.ts, gauges-subgraph.client.ts, block-subgraph.client.ts.
Duplication is not always bad but in this case they were having different behaviour: 

One was sending query as an object and another one was sending the JSON.stringified object. 

Using a single graphql client could avoid those problems.

src/lib/utils/exceptions.ts
```diff
@@ -0,0 +1,8 @@
+export function handleFetchException(
```
garethfuller, line 1: Should maybe be more explicit with the function name here, e.g.
```suggestion
export function logException(
```

src/composables/queries/useUserGaugeSharesQuery.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+import { mountComposable, waitForQueryData } from '@tests/mount-helpers';
+import {
+  defaultGaugeBalance,
+  defaultPoolId,
+  defaultTotalSupply,
+} from '@tests/msw/graphql-handlers';
+
+initDependenciesWithDefaultMocks();
+
+test('Returns token prices from balancer SDK', async () => {
```
garethfuller, line 12: Should this description be changed?

src/composables/queries/useUserPoolSharesQuery.spec.ts
```diff
@@ -0,0 +1,14 @@
+import useUserPoolSharesQuery from '@/composables/queries/useUserPoolSharesQuery';
+import { mountComposable, waitForQueryData } from '@tests/mount-helpers';
+import { defaultPoolBalance } from '@tests/msw/graphql-handlers';
+
+test('Returns token prices from balancer SDK', async () => {
```
garethfuller, line 5: And this one?

src/services/balancer/gauges/gauges-subgraph.client.ts
```diff
@@ -21,7 +21,7 @@ export class GaugesSubgraphClient {
   }
 
   public payloadFor(query) {
-    return JSON.stringify({ query: jsonToGraphQLQuery({ query }) });
```
garethfuller, line 24: I assume this is just redundant? If not will it not break the request?

src/services/balancer/subgraph/entities/poolShares/index.ts
garethfuller: Yeah, this pattern has grown arms and legs from the initial implementation. We should use a generator like in your spike, but will require quite a big refactor.

garethfuller (approved):

src/lib/utils/exceptions.ts
agualis: I wanted to explicitly use this helper for fetch exceptions so what about `logFetchException`?

The current implementation can be valid for any other kind of exception but that could change if we decide to give more fetch related info in the future. Maybe a bit YAGNI but not too much effort.

src/services/balancer/gauges/gauges-subgraph.client.ts
agualis: I think so, also it was preventing MSW from detecting this requests as Graphql ones.

PR #2794, opened 2023-02-10
Display negative values in fees chart

alter-eggo:
# Description
There may be a case when there are negative swap fees e.g. in FX pools.
See: https://github.com/balancer-labs/balancer-subgraph-v2/issues/366
         https://docs.xave.co/product-overview-1/amm#mechanism
 
Example pool: `0x726e324c29a1e49309672b244bdc4ff62a270407000200000000000000000702`

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:

### DIFF
--- a/src/components/contextual/pages/pool/PoolChart.vue
+++ b/src/components/contextual/pages/pool/PoolChart.vue
@@ -13,6 +13,7 @@ import useTailwind from '@/composables/useTailwind';
 import { HistoricalPrices } from '@/services/coingecko/api/price.service';
 import { PoolSnapshot, PoolSnapshots, PoolType } from '@/services/pool/types';
 import { twentyFourHoursInSecs } from '@/composables/useTime';
+import { isFx } from '@/composables/usePool';
 
 /**
  * TYPES
@@ -93,7 +94,7 @@ const tabs = [
 ];
 const activeTab = ref(tabs[0].value);
 
-const currentChartValue = ref('');
+const currentChartValue = ref({ num: '', isNegative: false });
 const currentChartDate = ref('');
 const isFocusedOnChart = ref(false);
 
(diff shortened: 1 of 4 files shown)


src/constants/pools.ts
```diff
@@ -388,7 +388,6 @@ const POOLS_POLYGON: Pools = {
     'Gyro2',
     'Gyro3',
     'GyroE',
-    'FX',
```
alter-eggo, line 391: somehow FX pool type was not removed from excluded list, but worked anyway, probably api doesn't handle this

src/components/contextual/pages/pool/PoolChart.vue
```diff
@@ -381,6 +382,11 @@ const defaultChartData = computed(() => {
   return { title, value: chartData.value.defaultHeaderStateValue };
 });
 
+const showFxPoolWarning = computed(() => {
```
garethfuller, line 385: Since this warning is Xave specific we should probably limit this conditional to only Xave pools too.

garethfuller (approved):
Also, could you see if Jarek can get the total liquidity calc in the SDK to handle these pool types.

src/components/contextual/pages/pool/PoolChart.vue
garethfuller: No, I don't think so. I think the pool factories are permissionless so anyone could technically create an FX pool. Is that right @markusbkoch?

src/constants/pools.ts
alter-eggo: in current pool request to api, we don't send excluded pools list, https://github.com/balancer-labs/frontend-v2/blob/33ee35e5fc7ea412e060fa81ec0c218426bff6e1/src/services/balancer/api/entities/pool/index.ts#L37

so it seems we need to change the whole request flow

I remember you wanted to refactor/combine usePoolsQuery and usePoolQuery, I think its better to do in the scope of this refactoring as it will fix this issue

src/components/contextual/pages/pool/PoolChart.vue
garethfuller: Talked to Markus and I think it's fine to leave this conditional as it is.

PR #2796, opened 2023-02-10
fix: Catch all user rejected tx errors

garethfuller:
# Description

Currently we are showing an error message after a user cancels a transaction in any BalActionSteps (e.g. join/withdraw/stake/unstak/etc). This PR ensures we parse all error types so that we don't render errors or submit them to Sentry in the UI in this case.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Try a join transaction and cancel it, you should not see any error message. Try this on production and you will.

## Visual context

n/a

## Checklist:

### DIFF
--- a/src/composables/useTransactionErrors.ts
+++ b/src/composables/useTransactionErrors.ts
@@ -33,11 +33,20 @@ export default function useTranasactionErrors() {
     };
   }
 
+  function isUserRejected(error): boolean {
+    return (
+      (error?.code && error.code === 4001) ||
+      (error?.message && error.message.includes('user rejected transaction')) ||
+      (error?.cause &&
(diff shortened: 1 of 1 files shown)


alter-eggo:
hm, I got such error 

Screenshot 2023-02-10 at 15 42 30

garethfuller:
@alter-eggo Yeah logging to the console is fine and what we want, before we were actually displaying an error in the UI to the user which is not what we want. Just test on production and you'll see what I mean.

alter-eggo:
@garethfuller  ah, yeah, I see now

alter-eggo (approved):

PR #2800, opened 2023-02-10
feature: Use Static or VaultModel simulations for generalised joins and exits

simeonkerkola:
# Description

### Use VaultModel (TS Math) and Static (onchain calls) simulations to calculate token amounts for Generalised joins/exits.

Static mode is more accurate and is used if user has:
1. Approved relayer (joins and exits)
2. Approved all tokens they are joining with (joins only)
3. Has enough token balance (joins and exits)

Otherwise, VaultModel simulation is used.

### Other changes:
- Make `useTokenApprovalActions` `tokenAddresses` parameter reactive
- join pool composable returns both relayer approval action and token approval actions (if any)

### TODO:
- Waiting for an SDK update to fix the incorrect scaling for some tokens when simulating Exit with VaultModel. (Euler pool is affected) 
  - Screenshot 2023-02-20 at 7 59 31

## Type of change

## How should this be tested?

You can test manually with pool: http://localhost:8080/#/polygon/pool/0x48e6b98ef6329f8f0a30ebb8c7c960330d64808500000000000000000000075b/

Use https://revoke.cash/ to revoke some token approvals to see the the simulation type is correct.

**Simulation types:**
- Static (=2), Should only run when relayer and tokens in are approved (will throw an error otherwise), and if preview modal is open.
- VaultModel (=1), When the condition for the Static simulation is not met.
- Tenderly (=0), Is not used at all any longer

| Static Simulation | VaultModel Simulation |
| ----------------- | ---------------------- |
|  Screenshot 2023-02-13 at 18 45 33 | Screenshot 2023-02-13 at 18 46 32 |

## Visual context

## Checklist:


src/components/cards/CreatePool/CreateActions.vue
```diff
@@ -70,10 +70,8 @@ const {
 } = usePoolCreation();
 const { networkSlug } = useNetwork();
 
-/**
- * COMPUTED
- */
 const actions = ref<TransactionActionInfo[]>([
+  ...tokenApprovalActions.value,
```
garethfuller, line 74: I'm pretty sure this won't work because I just changed it from this to fix a bug. This is a ref not a computed property, so it's only set on component setup. If the approvals have not been fetched via the tokens.provider allowances query then it returns an incorrect tokenApprovalActions array. That is why below in the `onBeforeMount` hook I had to use the explicit fetch method.

src/components/forms/pool_actions/InvestForm/components/InvestPreviewModal/components/InvestActionsV2.vue
```diff
@@ -59,23 +57,11 @@ const {
...
-  tokensToApprove.value,
-  amountsToApprove
-);
-
 /**
  * COMPUTED
  */
 const actions = computed((): TransactionActionInfo[] => [
   ...approvalActions.value,
-  ...tokenApprovalActions,
```
garethfuller, line 78: Why would you remove this from here? approvalActions and tokenApprovalActions are different things. Where are the token approvals loaded now?

garethfuller (changes requested):

src/components/forms/pool_actions/InvestForm/components/InvestPreviewModal/components/InvestActionsV2.vue
simeonkerkola: Token approvals and relayer approval both are now part of the `approvalActions` that comes from  [join-pool-provider]( https://github.com/balancer-labs/frontend-v2/pull/2800/files#diff-c5466dc8169d416c1cd54c2bc0e3e8d6e64dae65d71bdc35675bd48b3c3dca87R211).

Provider needs to know about token approvals, so it knows when to use right simulation type.

src/components/cards/CreatePool/CreateActions.vue
simeonkerkola: You are right

src/providers/local/join-pool.provider.ts
```diff
@@ -264,13 +289,16 @@ const provider = (props: Props) => {
...
+      const simulationType = getSimulationType();
 
       const output = await joinPoolService.queryJoin({
         amountsIn: amountsInWithValue.value,
         tokensIn: tokensIn.value,
         prices: prices.value,
         signer: getSigner(),
         slippageBsp: slippageBsp.value,
+        relayerSignature: relayerSignature.value,
+        simulationType: simulationType,
```
garethfuller, line 301: At a high level, I don't think we should be specifying the simulationType and passing it into the service from the provider. It's something specific to a particular handler and should be handled there. I think it makes more sense that approval state becomes part of the JoinParams and then the handler decides which simulation type to use.

garethfuller (changes requested):

src/providers/local/join-pool.provider.ts
garethfuller: There may also be other handlers in the future that need access to the approval state to decide how to handle the join but it wouldn't require simulation type so then you'd have to pass both anyway.

simeonkerkola: Done. Provider decides the simulation type based on token approvals, and account's token balances.

src/components/cards/CreatePool/CreateActions.vue
simeonkerkola: Rolled back

src/providers/local/exit-pool.provider.ts
```diff
@@ -397,7 +400,9 @@ const provider = (props: Props) => {
         slippageBsp: slippageBsp.value,
         tokenInfo: exitTokenInfo.value,
         prices: prices.value,
-        relayerSignature: '',
+        approvalActions: approvalActions.value,
+        bptInValid: bptInValid.value,
```
garethfuller, line 404: Looking at the use if this value `bptInValid` I don't think it's relevant. Is it ever invalid? How can the slider be in an invalid state?

src/providers/local/join-pool.provider.ts
```diff
@@ -71,6 +72,7 @@ const provider = (props: Props) => {
   const priceImpact = ref<number>(0);
   const highPriceImpactAccepted = ref<boolean>(false);
   const txError = ref<string>('');
+  const approvalActions = ref<TransactionActionInfo[]>([]);
```
garethfuller, line 75: Should we not use these in the join flow? So we don't define/fetch them twice?

src/services/balancer/pools/exits/handlers/exit-pool.handler.ts
```diff
@@ -24,6 +25,8 @@ export type ExitParams = {
   signer: JsonRpcSigner;
   slippageBsp: number;
   relayerSignature?: string;
+  bptInValid: boolean;
+  approvalActions: TransactionActionInfo[];
```
garethfuller, line 29: Doesn't look like we're using this in exits?

src/services/balancer/pools/exits/handlers/generalised-exit.handler.ts
```diff
@@ -60,15 +62,30 @@ export class GeneralisedExitHandler implements ExitPoolHandler {
...
-      signerAddress,
-      slippage,
-      signer,
-      SimulationType.Tenderly, // TODO: update to use VaultModel + Static (see SDK example for more details)
-      relayerSignature
-    );
+    // Static call simulation is more accurate than VaultModel, but requires relayer approval and
+    // account to have enough BPT balance
+    const simulationType: SimulationType =
+      bptInValid && !approvalActions.length
```
garethfuller, line 68: Seems like this `!approvalActions.length` will always be true. So should we just be using the VaultModel for all exits?

garethfuller (changes requested):

src/providers/local/exit-pool.provider.ts
simeonkerkola: Slider can't go over account balance, but manually typing the input can.

src/services/balancer/pools/exits/handlers/generalised-exit.handler.ts
simeonkerkola: I'm not sure I understand. Only approval action for exits is currently the relayer approval, so we need to check for that.

VaultModel = TS math, are you suggesting it is accurate enough to use always?

src/providers/local/join-pool.provider.ts
simeonkerkola: Yeah, `approvalActions` are exported here, and used in `InvestActionsV2.vue`

garethfuller (approved):

PR #2805, opened 2023-02-13
Hotfix: handle swap errors

garethfuller:
# Description

We were logging error contents to the UI in the swap card, even for user cancelled transactions, which results in a huge error message. This PR makes the error message shown to the user more friendly and prevents errors being shown in the case where a user rejects the transaction.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Try cancelling a swap tx in production and then in this branch, you will notice that the error message isn't shown.

## Visual context

n/a

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer-labs/frontend-v2",
-  "version": "1.89.0",
+  "version": "1.89.1",
   "lockfileVersion": 2,
   "requires": true,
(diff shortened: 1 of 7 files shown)


Matthews3301 (approved):
Looks good, maybe we don't need to do `console.trace(error);` if it's a rejected transaction?

PR #2816, opened 2023-02-14
Hotfix: multi choice custom input

garethfuller:
# Description

The custom input in the multic choice forms were broken. When you input a custom value the onCustomInput method wasn't being triggered and therefore the value wasn't being used and the input didn't change style to suggest it had been used. This PR fixes that even trigger so that custom inputs are respected. This affects to places:|
- The slippage form in the navbar settings dropdown
- The fee selection in the pool creation UI.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test the slippage form in the app settings dropdown and try to set a custom slippage amount.
- [ ] Test the fee form in the pool creation UI and try to set a custom fee.

## Visual context

n/a

## Checklist:

### DIFF
--- a/src/components/cards/CreatePool/PoolFees.vue
+++ b/src/components/cards/CreatePool/PoolFees.vue
@@ -88,7 +88,7 @@ function onFixedInput(val: string): void {
 }
 
 function onCustomInput(val: string): void {
-  if (!val) return;
+  if (val === '') return;
 
   initialFee.value = (Number(val) / 100).toString();
   isCustomFee.value = true;
--- a/src/components/cards/CreatePool/PoolFees.vue
+++ b/src/components/cards/CreatePool/PoolFees.vue
@@ -82,12 +82,15 @@ const feeOptions = FIXED_FEE_OPTIONS.map(option => {
  * FUNCTIONS
  */
 function onFixedInput(val: string): void {
+  fee.value = '';
   initialFee.value = '0';
   initialFee.value = val;
   isCustomFee.value = false;
 }
 
 function onCustomInput(val: string): void {
+  if (!val) return;
+
   initialFee.value = (Number(val) / 100).toString();
   isCustomFee.value = true;
 
@@ -133,6 +136,8 @@ async function onChangeFeeController(val: string) {
     height: cardWrapper.value?.offsetHeight || 0,
   });
 }
+
+watch(fee, onCustomInput, { immediate: true });
 </script>
 
 <template>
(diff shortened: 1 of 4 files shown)


src/components/cards/CreatePool/PoolFees.vue
```diff
@@ -82,12 +82,15 @@ const feeOptions = FIXED_FEE_OPTIONS.map(option => {
  * FUNCTIONS
  */
 function onFixedInput(val: string): void {
+  fee.value = '';
   initialFee.value = '0';
   initialFee.value = val;
```
alter-eggo, line 87: hm, don't understand here why initialFee is set to 0 first, and then to value

src/components/forms/AppSlippageForm.vue
```diff
@@ -68,6 +69,7 @@ function onCustomInput(val: string): void {
 watch(
   slippage,
   newSlippage => {
+    console.log('newSlippage', newSlippage);
```
alter-eggo, line 72: necessary log?

src/components/cards/CreatePool/PoolFees.vue
```diff
@@ -82,12 +82,15 @@ const feeOptions = FIXED_FEE_OPTIONS.map(option => {
...
   isCustomFee.value = false;
 }
 
 function onCustomInput(val: string): void {
+  if (!val) return;
```
alter-eggo, line 92: if we check empty string here it's better to check empty string
`if (val === '')`
otherwise, arg type should be changed to `val?: string`

alter-eggo (approved):
some minor comments

src/components/cards/CreatePool/PoolFees.vue
garethfuller: I didn't want to change `initialFee.value = '0';` because it was already like that and didn't want to change the logic, although I agree it doesn't make much sense since it's set to `val` straight after, I'll remove. Setting `fee.value` to nothing here is because `fee.value` actually represents the custom input value, it's just badly named.

src/components/forms/AppSlippageForm.vue
garethfuller: Nope, removed.

src/components/cards/CreatePool/PoolFees.vue
garethfuller: Good point, changed

PR #2867, opened 2023-02-17
fix wrong APR_THRESHOLD import (develop merge)

garethfuller:
Original #2860


simeonkerkola:
We should allow merging only after all jobs in pipeline are completed successfully. This PR added failing typecheck to develop branch

PR #2875, opened 2023-02-17
Add baoUSD stablepool to allowlist

fabiaz84:
# Description

## Type of change

- [X] Other: add stablePool address


garethfuller:
I'm assuming the adding to Goerli was a mistake.

PR #2913, opened 2023-02-23
feat: seo improvement

alter-eggo:
# Description
This pr includes some seo improvements (set pages meta title and description)

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?
Check meta title and description on every page, it should correspond with the document.

## Visual context

## Checklist:


src/services/meta/meta.service.ts
```diff
@@ -0,0 +1,44 @@
...
+interface IMetaService {
+  setMeta(route: RouteLocationNormalized | string): void;
+}
+
+class MetaService implements IMetaService {
+  networkConfig: Config;
+  constructor() {
+    const { networkConfig: _networkConfig } = useNetwork();
+    this.networkConfig = _networkConfig;
+  }
```
garethfuller, line 15: I didn't realise it was to use in a service. For this it's probably best to do it like this since that's how we do it elsewhere. 
```suggestion
  constructor(private readonly config = configService) {}
```
Then within the class you can do:
```
this.config.network.name;
```

garethfuller (changes requested):

src/services/meta/meta.service.ts
alter-eggo: ah, yeah, it's much better. thanks

alter-eggo: ah, yeah, it's much better. thanks

src/pages/pool/_id.vue
```diff
@@ -130,6 +131,7 @@ function addIntersectionObserver(): void {
 }
 onMounted(() => {
   addIntersectionObserver();
+  metaService.setOgImage(poolId);
```
garethfuller, line 134: I don't see how this would work? The HTML structure is already rendered at this point? Would you not need to do this in the router?

alter-eggo: hm, seems problem is deeper, og scraper doesn't run js at all. putting this in router doesn't help

alter-eggo: yeah. seems we need to implement middleware in cloudflare to handle this

src/pages/pool/_id.vue
```diff
@@ -192,6 +193,15 @@ watch(poolQuery.error, () => {
...
 });
+
+watch(
+  () => pool.value,
+  () => {
+    if (pool.value) {
+      metaService.setMeta(route, pool.value);
+    }
+  }
+);
```
garethfuller, line 204: As far as I can tell, this doesn't work. Just run the preview URL for pool page through some OG metadata checkers. Of course it works to look at the page title when you're browsing, but that's about it. Maybe we need a different solution? For example, could we build a script that compiles a static json of poolId -> metadata that we can use in the routes? Will that even work because again it's being dynamically set after page load.

alter-eggo: actually for og this doesn't work, but it works for e.g googlebots and it is one of the tips here: 
https://developers.google.com/search/docs/crawling-indexing/javascript/javascript-seo-basics#titles-and-snippets

garethfuller (changes requested):
Once last change, if we have metatdata for the pool in constants, please use the name of the pool as title instead of the pool symbol.

alter-eggo:
yeah, good suggestion, done

garethfuller (approved):

PR #2952, opened 2023-03-01
Remove vuex dependency

agualis:
# Description

- Refactored 2 modules into composables
- Refactored related client componentes from `<script>` to `<script setup>`
- Added and improved tests

Bye bye vuex 👋

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:


src/components/navs/AppNav/AppNavSettings.vue
```diff
@@ -1,3 +1,78 @@
+
```
alter-eggo, line 1: empty line

src/components/navs/AppNav/AppNavSettings.vue
```diff
@@ -1,3 +1,78 @@
...
+} = useWeb3();
+const { ethereumTxType, setEthereumTxType } = useEthereumTxType();
+
+// DATA
+const data = reactive({
+  copiedAddress: false,
+});
+
+// COMPUTED
+const avatarUri = computed(() => profile.value?.avatar || undefined);
```
alter-eggo, line 40: seems no need in `|| undefined`

src/components/popovers/SwapSettingsPopover.vue
```diff
@@ -1,3 +1,47 @@
+<script lang="ts">
+export enum SwapSettingsContext {
```
alter-eggo, line 2: may be there is some better place for this, `swap.service.ts`?

src/composables/useAppStore.ts
```diff
@@ -0,0 +1,37 @@
...
+  lsSet(LS_KEYS.App.Locale, newLocale);
+  i18n.global.locale.value = newLocale;
+}
+
+function setTransactionDeadline(newTransactionDeadline: number) {
+  transactionDeadline.value = newTransactionDeadline;
+  lsSet(LS_KEYS.App.SwapDeadline, newTransactionDeadline);
+}
+
+export function useAppStore() {
```
alter-eggo, line 27: 👍

alter-eggo (approved):
nice 👍! there is definitely not much need in vuex

src/components/navs/AppNav/AppNavSettings.vue
agualis: It is needed to avoid `null` value, which is not a valid value for iconUri prop in Avatar component.

I could change the type of that prop but then I should also change this api:

`const iconSRC = computed(() => resolve(props.iconURI));
`
And i prefered to avoid that for now.

alter-eggo: ah, I see, ok

src/components/popovers/SwapSettingsPopover.vue
agualis: `SwapSettingsContext` is the type of a prop in `SwapSeetingsPopover` component so I think that it makes sense to leave it there.
 
Look how you only import SwapSettingsContext when using the SwapSeetingsPopover component (high cohesion). If some other code needed this type outside this scope, then I would extract it but, that's not the case here.

alter-eggo: I am just not a big fan of exporting types from .vue files, seems they shouldn't be responsible only for it. but yeah, may be I'm wrong here

agualis: I also don't like that pattern but, in this concrete case I think it is the less worst option. 

Moving the type to a service seems off IMO

Moving it to a new file `SwapSeetingsPopover.d.ts` would be my next option. But I prefer to have the type self-contained within the component that uses it. 

@garethfuller What do you think?

garethfuller: I don't have a strong opinion here. I think exporting types highly relevant to a component from the component file is ok.

src/composables/useAppStore.ts
```diff
@@ -0,0 +1,37 @@
+import LS_KEYS from '@/constants/local-storage.keys';
```
garethfuller, line 1: We already have a composable called `useApp` which was meant to be for this type of app-level state. Please just put this logic in there. Using the `Store` affix is a bit restrictive.

src/services/contracts/vault.service.ts
```diff
@@ -35,8 +35,8 @@ export default class VaultService {
     tokenOutAmount: string,
     options: Record<string, any> = {}
   ): Promise<TransactionResponse> {
-    const storeState = store.state as any;
-    const deadline = calculateValidTo(storeState.app.transactionDeadline);
+    const { transactionDeadline } = useAppStore();
```
garethfuller, line 38: Not a big fan of using composables within services like this. Really `transactionDeadline` should be passed into the service as a param wherever the service is used in the component context.

src/components/navs/AppNav/AppSidebar/SidebarContent.vue
```diff
@@ -120,7 +120,7 @@ watch(blockNumber, async () => {
     <div
       class="flex flex-col justify-center px-4 h-20 border-b border-gray-800"
     >
-      <AppLogo forceDark />
```
garethfuller, line 123: This causes problems, why did it change?
- Open the preview in light mode
- Make the window narrow enough so the sidebar is togglable
- Open the sidebar, you should see that the logo is the wrong color.
Screenshot 2023-03-01 at 10 59 28

garethfuller (changes requested):

src/services/contracts/vault.service.ts
agualis: Super agree, refactored.

agualis (commented):
@garethfuller Refactored and manually tested but please review it cause it is the first time that I deal with this logic.

src/providers/local/exit-pool.provider.ts
```diff
@@ -506,6 +511,7 @@ export const exitPoolProvider = (pool: Ref<Pool>) => {
     exitTokenInfo,
     queryExitQuery,
     approvalActions,
+    transactionDeadline,
```
garethfuller, line 514: Don't think it needs to be exported here?

garethfuller (approved):

src/providers/local/exit-pool.provider.ts
agualis: Yep, deleted.

PR #3006, opened 2023-03-10
Change excluded pool types to included

timjrobinson:
# Description

There are many new Linear pools and different pool types being created (full list here: https://github.com/balancer/balancer-subgraph-v2/blob/master/src/mappings/helpers/pools.ts#L9-L33), almost all of which we don't want to show. Rather than constantly excluding new types it makes more sense to have an included pool types list. 

## Type of change

- [x] Code refactor / cleanup

## How should this be tested?

- Load each pools list on each network. Ensure the list of returned pools is the same as production. 
- Search for different tokens in the pools list, ensure the list of returned pools is the same. 

## Checklist:

### DIFF
--- a/src/constants/pools.ts
+++ b/src/constants/pools.ts
@@ -74,7 +74,7 @@ export type Pools = {
     Gauntlet: string[];
   };
   BlockList: string[];
-  ExcludedPoolTypes: string[];
+  IncludedPoolTypes: string[];
   Stable: {
     AllowList: string[];
   };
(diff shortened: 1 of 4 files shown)


src/constants/pools.ts
```diff
@@ -192,17 +188,14 @@ const POOLS_MAINNET: Pools = {
...
-    'AaveLinear',
-    'EulerLinear',
-    'GearboxLinear',
-    'Linear',
-    'ERC4626Linear',
-    'Gyro2',
-    'Gyro3',
-    'GyroE',
-    'HighAmpComposableStable',
+  IncludedPoolTypes: [
```
timjrobinson, line 191: FX was not excluded on Mainnet, should we add it here?

src/constants/pools.ts
```diff
@@ -510,17 +503,14 @@ const POOLS_POLYGON: Pools = {
...
```
timjrobinson, line 506: FX also not included here

alter-eggo (approved):

PR #3016, opened 2023-03-13
Hotfix: Don't treat pools with joins disabled as deep pools

timjrobinson:
# Description

Changing isDeep to be false when a pool is disabled, so that if a parent pool has disabled joins users can still withdraw all linear tokens correctly. 

## Type of change

## Checklist:


timjrobinson:
Not the right logic, closing.

PR #3042, opened 2023-03-16
Create a protocol risks page

pkattera:
# Description

A page to outline risks for LPs on the Balancer Protocol.

## Type of change

- [x] New feature (non-breaking change which adds functionality): Enabling anchor links
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Documentation or wording changes
- [ ] Other

## How should this be tested?

Go to the new page at /#/risks (also found in the footer) and make sure it displays well across different screen sizes.

View the two links:
- In the footer
- On the Connect Wallet screen.

One important change to test is the scroll behavior across the app. I made a change to the `scrollBehavior` function in `src/plugins/router/index.ts`, in order to enable normal functioning on anchor links which currently don't seem to work across the site. Please review this code.

## Checklist:

### DIFF
--- a/src/plugins/router/index.ts
+++ b/src/plugins/router/index.ts
@@ -214,12 +214,9 @@ const router = createRouter({
   history: createWebHashHistory(),
   routes,
   scrollBehavior(to, from, savedPosition) {
-    if (savedPosition) {
-      return savedPosition;
-    }
-    if (to.hash) {
-      return { el: to.hash };
-    }
+    if (savedPosition) return savedPosition;
+    if (to.hash) return { el: to.hash };
+
     return { x: 0, top: 0 };
   },
 });
--- a/src/plugins/router/index.ts
+++ b/src/plugins/router/index.ts
@@ -40,6 +40,8 @@ const PrivacyPolicyPage = () =>
   );
 const TermsOfUsePage = () =>
   import(/* webpackChunkName: "TermsOfUsePage" */ '@/pages/terms-of-use.vue');
+const RisksPage = () =>
+  import(/* webpackChunkName: "RisksPage" */ '@/pages/risks.vue');
 const SwapPage = () =>
   import(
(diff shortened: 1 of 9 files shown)


src/plugins/router/index.ts
```diff
@@ -205,8 +213,14 @@ if (isGoerli.value) {
 const router = createRouter({
   history: createWebHashHistory(),
   routes,
-  scrollBehavior() {
-    return { top: 0 };
+  scrollBehavior(to, from, savedPosition) {
```
pkattera, line 216: Please review these changes in detail. I did this to ensure that anchor links would work (previously, anchor links would place users at the top of the page). 

One nice advantage is that people return to the position they were previously at. So if a user on the home page loads some extra rows and scrolls down before clicking in to a pool, they will be returned to the same position if they return back to the homepage. 

Could this have any other unintended consequences?

alter-eggo: hm, I think it's ok, but need some testing

src/plugins/router/index.ts
```diff
@@ -205,8 +213,14 @@ if (isGoerli.value) {
...
+    if (savedPosition) {
+      return savedPosition;
+    }
+    if (to.hash) {
+      return { el: to.hash };
+    }
+    return { x: 0, top: 0 };
```
garethfuller, line 223:
```suggestion
    if (savedPosition) return savedPosition;
    if (to.hash) return { el: to.hash };

    return { x: 0, top: 0 };
```

garethfuller (changes requested):
Not sure I like the smooth scroll behaviour here: https://www.loom.com/share/3f8c7df8bb834781a642358a4baa5c57 On clicking back to the home page should it not just be immediately at the previous location rather than smooth scrolling down to it?

garethfuller (approved):

PR #3079, opened 2023-03-23
Hotfix: swap approval states

garethfuller:
# Description

Because actions steps in the swap preview is a computed property based on many other computed properties, it sometimes triggered a watcher of the actions prop in BalActionSteps despite the actions not actually changing. When that watcher was triggered it reset the state of all the BalActionStep actions, so any completed actions looked like they weren't. This PR ensures the action states are only reset if the length of actions passed via props changes. This is only usually the case once, after initial loading of approvals for example.

## Type of change

## How should this be tested?

- [ ] Test swap approvals work as expected.

## Visual context

n/a

## Checklist:

### DIFF
--- a/src/components/modals/SwapPreviewModal.vue
+++ b/src/components/modals/SwapPreviewModal.vue
@@ -35,35 +35,31 @@ type Props = {
 /**
  * PROPS & EMITS
  */
-
 const props = defineProps<Props>();
-
 const emit = defineEmits(['swap', 'close']);
-// COMPOSABLES
+
+/**
+ * COMPOSABLES
+ */
 const { t } = useI18n();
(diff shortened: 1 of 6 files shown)


src/components/modals/SwapPreviewModal.vue
```diff
@@ -351,52 +356,32 @@ const showPriceUpdateError = computed(
...
-            'swapSummary.transactionTypesTooltips.tokenApproval.content'
-          ),
-        },
-      ]
-    : []),
-  {
+const actions = computed((): TransactionActionInfo[] => {
+  const actions: TransactionActionInfo[] = [];
+
+  if (showCowswapRelayerApprovalStep.value) {
```
alter-eggo, line 362: 👍

PR #3084, opened 2023-03-24
Migrate voting gauges to new config system

timjrobinson:
# Description

- Migrates the voting-gauge config to the new config folder. 

Previously we had a voting-gauge constants file AND a stakable pool list which are almost identical (the only difference is there might be pools you can stake in but can't vote for). There is now one stakable pool list with pools that have voting enabled and voting disabled. 

Tested by comparing the old voting-gauges.json file with the newly generated one and they are identical, but the networks in the wrong order. In the testing section you can see how to compare the files in the right order (I didn't want to leave this code in as it would have to be changed when adding new networks). 

## Type of change

- [x] Code refactor / cleanup

## How should this be tested?

If you apply the following diff to `src/lib/scripts/voting-gauge.generator.ts` (copy the diff into `changes.diff` and run `git apply changes.diff`) and generate the gauges, the output list should be identical to the one from the develop branch:

```
diff --git a/src/lib/scripts/voting-gauge.generator.ts b/src/lib/scripts/voting-gauge.generator.ts
index e9fb67fce..ae3455f1e 100644
--- a/src/lib/scripts/voting-gauge.generator.ts
+++ b/src/lib/scripts/voting-gauge.generator.ts
@@ -477,8 +477,18 @@ async function getGaugeInfo(
   console.log('Fetching gauges info...');
   console.time('getGaugeInfo');
 
+  const sortedNetworks: Network[] = [
+    Network.MAINNET,
+    Network.ARBITRUM,
+    Network.POLYGON,
+    Network.GOERLI,
+    Network.OPTIMISM,
+    Network.GNOSIS,
+  ];
+
   const POOLS = flatten(
-    Object.entries(config).map(([network, networkConfig]) => {
+    sortedNetworks.map(network => {
+      const networkConfig = config[network];
       return networkConfig.pools.Stakable.VotingAllowed.map(id => {
         return {
           id,
```

## Checklist:


src/types/pools.ts
```diff
@@ -64,7 +64,8 @@ export type Pools = {
   };
   Factories: Record<string, FactoryType>;
   Stakable: {
-    AllowList: string[];
+    VotingAllowed: string[];
+    VotingDisabled: string[];
```
garethfuller, line 68: This is confusing to me now. Why do we have these two attributes and why are VotingDisabled pools still Stakable? Either we need comments to explain the logic here or the naming could be better.

garethfuller: Yeah, I still find the naming confusing. Maybe instead it could be something like:
```suggestion
    // Pools to be included in the voting gauges list.
    VotingGaugePools: string[];
    // Pools that have additional rewards but are not included in the VotingGaugePools list.
    AllowList: string[];
```

timjrobinson: Sounds good. Changed.

garethfuller (approved):

PR #3097, opened 2023-03-27
Add boosts enabled flag, change l2 check to use it

timjrobinson:
# Description

- Add a new boolean to mark if a network has boosts enabled for its pools (where you can earn more than 1x BAL rewards APR)
- In all places where pool boosts are shown or calculated use this new Boolean instead of `isL2` (which is too ambiguous). 

## Type of change

- [x] Code refactor / cleanup

## How should this be tested?

- Ensure that pools show APR ranges on Mainnet and not other networks
- Ensure boosted APR's show on the portfolio page. 
- Ensure all networks load the pools table and individual pools correctly without stalling waiting for boost information. 

## Checklist:

### DIFF
--- a/src/components/contextual/pages/pools/StakedPoolsTable.vue
+++ b/src/components/contextual/pages/pools/StakedPoolsTable.vue
@@ -42,7 +42,7 @@ const noPoolsLabel = computed(() => {
 
 const hiddenColumns = computed(() => {
   const _hiddenColumns = ['poolVolume', 'migrate', 'lockEndDate'];
-  if (isPoolBoostsEnabled.value) _hiddenColumns.push('myBoost');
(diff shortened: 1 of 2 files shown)


agualis (changes requested):
Maybe we can take this opportunity to add unit tests `useUserBoostsQuery.ts`.

That would require using `getMulticaller` from dependencies to inject a mock of the old multicaller. Another improvement would be  migrating the underlying `old multicaller` calls to use the [new multicaller](https://github.com/balancer/frontend-v2/blob/e16fddac38b3a83d29ec6972cf27e1274a7cf471/src/services/multicalls/multicaller.ts).  

Of course, it can be done in a different PR and I can approve this one if you don't have time now.

timjrobinson:
@agualis I'd rather do that in another PR as this is a simple one just to fix the config.

agualis (approved):

PR #3098, opened 2023-03-27
Add new isVeBalSupported boolean using config data instead of isL2

timjrobinson:
# Description

- Add a new boolean isVeBalSupported that is set based on if a veBal address is set in config.
- Wherever we have a veBAL related page (such as locking veBAL, or seeing rewards) use this boolean instead of `isL2`. 

## Type of change

- [x] Code refactor / cleanup

## How should this be tested?

- Ensure that the veBAL page works correctly on Mainnet and Goerli and redirects correctly on other networks.
- Ensure veBAL locked stats show on portfolio page on Mainnet and Goerli
- Ensure the veBAL locked stats do not show on other networks. 

## Checklist:

### DIFF
--- a/src/composables/useVeBAL.ts
+++ b/src/composables/useVeBAL.ts
@@ -19,8 +19,7 @@ const showRedirectModal = ref(false);
  * COMPUTED
  */
 export const isVeBalSupported = computed<boolean>(
-  () =>
-    configService.network.addresses.veBAL !== '' && POOLS.IdsMap?.veBAL !== ''
+  () => configService.network.addresses.veBAL !== ''
 );
 
 /**
--- a/src/composables/useVeBAL.ts
+++ b/src/composables/useVeBAL.ts
@@ -1,7 +1,6 @@
 import { differenceInSeconds, formatDistanceToNow, sub } from 'date-fns';
 import { computed, ref } from 'vue';
 
-import { isGoerli, isMainnet } from '@/composables/useNetwork';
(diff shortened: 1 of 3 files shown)


src/composables/useVeBAL.ts
```diff
@@ -18,8 +18,9 @@ const showRedirectModal = ref(false);
 /**
  * COMPUTED
  */
-export const isVeBalSupported = computed(
-  () => isMainnet.value || isGoerli.value
+export const isVeBalSupported = computed<boolean>(
+  () =>
+    configService.network.addresses.veBAL !== '' && POOLS.IdsMap?.veBAL !== ''
```
garethfuller, line 23: I think we can simplify this, the IdsMap is just a helper map so you can access a poolId by name.
```suggestion
    configService.network.addresses.veBAL !== ''
```
Maybe we should even make the `veBAL` key on addresses optional? I don't like that we have a bunch of keys in the config which are just empty strings. If the contract doesn't exist on a network then the key shouldn't really be accessible.

garethfuller (changes requested):

src/composables/useVeBAL.ts
timjrobinson: Actually it seems like it might be more trouble than it's worth as that string is used in many places that don't accept undefined, so they'd all have to be changed to `address.veBAL || ''` or have an if around them anyway.

garethfuller (approved):

PR #3101, opened 2023-03-28
Chore: Make AppNavMenu dynamic and based on config

timjrobinson:
# Description

- Change the AppNavMenu to be built based on configuration settings instead of hard coded. 

Question: The order of these buttons is now different. Do we want them in a specific order? 
 
## Type of change

- [x] Code refactor / cleanup

## How should this be tested?

- Ensure that only Ethereum, Polygon, Arbitrum, and Gnosis Chain appear in the navigation menu
- Ensure that Goerli appears in the menu in dev mode
- Ensure that all button names look correct and clicking the buttons takes you to the right network. 

## Checklist:

### DIFF
--- a/src/lib/config/types.ts
+++ b/src/lib/config/types.ts
@@ -72,6 +72,8 @@ export interface Config {
   slug: string;
   network: string;
   unknown: boolean;
+  inNetworkSelector: boolean;
+  testNetwork: boolean;
   rpc: string;
   publicRpc?: string;
   ws: string;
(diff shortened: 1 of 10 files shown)


src/lib/config/types.ts
```diff
@@ -72,6 +72,8 @@ export interface Config {
   slug: string;
   network: string;
   unknown: boolean;
+  inNetworkSelector: boolean;
```
garethfuller, line 75: How about?
```suggestion
  uiSupported: boolean;
```

timjrobinson: Makes sense. I've made it `visibleInUI` to be more explicit that enabling this makes it show in the menu's etc.

garethfuller (approved):

PR #3116, opened 2023-03-29
Fix Price Impact calculations

timjrobinson:
# Description

Our price impact calculation function is currently quite broken (it was made years ago and never had tests and no one really knew how it worked). It previously only worked if the sell token was 18 decimals AND the swap type was SwapExactIn.

This PR adds many tests and documentation for this function so we can fix it once and for all. 

## Type of change

## How should this be tested?

- Ensure that price impact looks correct for large swaps of USDC/USDT -> other tokens, or SwapExactOut swaps of any tokens. 

## Checklist:


timjrobinson:
Turned into hotfix in #3119

PR #3140, opened 2023-04-04
fix: Swap decimal inputs

garethfuller:
# Description

There was a bug where by if you typed inputs like .1, .2, .3 etc into the swap UI the amount would be parsed and sent to the SOR as 1, 2, 3, etc. So you would have a 10x output returned for your input and the tx would obviously fail. This PR fixes this bug by parsing the original input number correctly and formating it for the SOR.

## Type of change

## How should this be tested?

- [ ] Input various amounts into the Swap UI like .1, .2, .03 etc, and check the outputs make sense.

## Checklist:


timjrobinson (approved):
I can't reproduce this in prod (using decimals like 0.1 always works for me in Chrome and Firefox) but the PR seems good and works fine too.

PR #3224, opened 2023-04-25
Fix polygon native asset address

timjrobinson:
# Description

The native asset address on Polygon has always been wrong. This didn't matter previously as when fetching prices the SDK had a check for if it's a native asset and to pull from a different Coingecko API, and the balances code just calls `this.provider.getBalance` for the native asset. The default Coingecko API doesn't work with the 0xEee address:

```
> curl "https://api.coingecko.com/api/v3/simple/token_price/polygon-pos?contract_addresses=0x0000000000000000000000000000000000001010&vs_currencies=usd"
{"0x0000000000000000000000000000000000001010":{"usd":0.971639}}
> curl "https://api.coingecko.com/api/v3/simple/token_price/polygon-pos?contract_addresses=0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE&vs_currencies=usd"
{}                                       
```

This is needed for switching to the Beets API.

## Type of change

## How should this be tested?

- Ensure prices in pools and on claim page for MATIC related things look correct. 
- Ensure useTokenPricesQuery is returning the correct price for this address. 

## Checklist:

### DIFF
--- a/src/lib/config/polygon/index.ts
+++ b/src/lib/config/polygon/index.ts
@@ -41,7 +41,7 @@ const config: Config = {
   blockTime: 4,
   nativeAsset: {
     name: 'Matic',
-    address: '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE',
+    address: '0x0000000000000000000000000000000000001010',
     symbol: 'MATIC',
     decimals: 18,
     deeplinkId: 'matic',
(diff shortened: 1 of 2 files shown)


garethfuller (commented):
Are you sure this doesn't break other functionality? Native asset addresses are used/referenced a lot int the join/exit and swap UIs.

garethfuller:
For most other networks we reference the native asset address as `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` so I'm against changing this actually, we should do the conversion at the source of the issue.

timjrobinson:
The source of the issue is that this address is wrong though. Other networks use the EEE address because they use Ether as their native asset (and Gnosis Chain doesn't have a native asset address).

I'll try joins and swaps, it seems most things work because the address is self referential.

timjrobinson:
I tried a few different ways of joining/exiting/swapping and all seem to work correctly.

garethfuller (approved):

PR #3232, opened 2023-04-25
feat: Allow user to specify signature support 

garethfuller:
# Description

Allow the user to specify whether they want to use signatures or not. Many wallet connection types don't support eth_signTypedData_v4 so it's common but not particularly easy to detect and turn off automatically.

## Type of change

## How should this be tested?

- [ ] Test a signature flow such as joining a deep pool with a wallet that supports signatures (MetaMask), check that initially it requests a relayer signature, then change the setting to off and check that the flow now requests a relayer approval tx.

## Checklist:


alter-eggo (approved):
how users will know about this feature? maybe we can notify them about it somehow

garethfuller:
@alter-eggo I'm not sure it's something we really want to highlight, but potentially we could direct users to this setting when they experience the rejection error that happens for wallets that don't support `eth_signTypedData_v4`.

timjrobinson (approved):
Tested with my Gnosis safe and it worked for me joining a deep pool on polygon. It never showed a sign relayer prompt though no matter what the toggle was. 

The toggle worked correctly with my normal Metamask wallet, when I turned it off it did a relayer transaction instead of signature as expected.

timjrobinson:
I also couldn't get the sign relayer button to appear for my Argent wallet, even though when I put breakpoints in all the spots it  checks it seems like it should be requesting a signature, not sure what's going on there.

PR #3243, opened 2023-04-27
Improve token fetching speed - make all token addresses lower case

timjrobinson:
# Description

The tokens list started to become very slow since adding many new prices from the new API because it's doing a O^2 search for token prices and balances looping over two arrays. You can see when filtering by tokens in beta it takes 2 - 5 seconds to load the list.

This O^2 search was done because we compare upper and lower case token addresses.

Instead of doing this, change all addresses to lowercase by default and use a map to search for prices / balances.

This makes the token list open instantly again.

I tested to ensure there weren't upper case addresses by adding an error if any were found. I've left the commented out code in there for testing purposes so you guys can also see if you can find any upper case addresses that I missed. 

## Other Changes

- Switched the main token list to the new one, this is so that when tokens are injected from a pool they are found in the list and ignored. Without the new list the frontend thinks they're new tokens and tries to re-fetch info about them. 
- Removed injecting the pool address into the token list - not sure why this was even in there in the first place, most pools don't have a price for their address so injected their token and waiting for a price just makes things much slower. 
- The above two changes + the general improvements makes loading an individual pool page much faster because it's no longer injecting + waiting for tokens to be loaded, and all token searches through maps instead of arrays is much faster. 

## Type of change

## How should this be tested?

- Ensure prices and balances look correct on all pages
- Run locally and uncomment the code that errors on uppercase tokens to see if you encounter any errors.

## Checklist:

### DIFF
--- a/src/composables/queries/usePoolQuery.ts
+++ b/src/composables/queries/usePoolQuery.ts
@@ -82,7 +82,6 @@ export default function usePoolQuery(
     await injectTokens([
       ...tokensListExclBpt(pool),
       ...tokenTreeLeafs(pool.tokens),
-      pool.address,
     ]);
 
     return pool;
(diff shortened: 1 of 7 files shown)


timjrobinson:
Superceded by #3258

PR #3247, opened 2023-04-28
Improve token fetch speed

timjrobinson:
# Description

This is an alternate variation on #3243 where I've created a custom case insensitive map so that we don't have to worry about the case tokens are in. 

It doesn't currently work because the iterator is implemented incorrectly and so the page doesn't load. I haven't yet figured out how to implement the iterator type. This is needed so when adding objects via the spread operator it adds them correctly. 

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


timjrobinson:
Superceded by #3258, I didn't like the look of this new class and it added a lot of complexity.

PR #3256, opened 2023-05-02
Use new mainnet tokenlist

timjrobinson:
# Description

Switching the main token list over to the new one that the API is using + fetching prices for. 

This fixes an issue where pools are injecting tokens that the frontend thinks are new and so the frontend is attempting to fetch prices + data for them which slows down single pool loading. 

## Type of change

## How should this be tested?

- Ensure that pools are loading as expected. 
- When swapping make sure the tokens list looks correct and tokens on the new list can be selected. 

## Checklist:

### DIFF
--- a/src/lib/config/mainnet/tokenlists.ts
+++ b/src/lib/config/mainnet/tokenlists.ts
@@ -3,7 +3,7 @@ import { TokenListURLMap } from '@/types/TokenList';
 const tokenlists: TokenListURLMap = {
   Balancer: {
     Default:
-      'https://raw.githubusercontent.com/balancer/tokenlists/main/generated/listed-old.tokenlist.json',
+      'https://raw.githubusercontent.com/balancer/tokenlists/main/generated/balancer.tokenlist.json',
     Vetted:
       'https://raw.githubusercontent.com/balancer/tokenlists/main/generated/balancer.tokenlist.json',
   },


garethfuller:
@timjrobinson the problem with this is that the token search modal becomes huge and slow. The original purpose of the `listed` list was to be a subset of tokens that we want to be easily searchable for. The alternative solution would be to ensure the `vetted` list is used for all the price/balance fetching.

timjrobinson:
@garethfuller The tokens provided by the API is the same as the default list.

Do we want the main tokens from tokens.provider to be all tokens that Balancer has whitelisted (which is the default list, also what the API returns)? Or only the tokens that appear in the search list? That seems to be the main confusion here. It's not clear what it should be. 

Also #3247 fixes the slow loading of the token search.

garethfuller:
The tokens provided by the API are actually equivalent to the old `vetted` list which is now just balance.tokenlist.json in the tokenlists repo. The `listed` list is supposed to be a smaller subset for the token search. I think the original `tokens` from `tokens.provider` was suppose to start off small using the subset list and grow as the user interacted with more pools/tokens. I think originally this was probably done to reduce the size of the coingecko request, so it's probably ok to use the full list for the `tokens`. However, this change will also change the list used for the token search as far as I'm aware which is not what we want.

PR #3258, opened 2023-05-02
Improve Token Search Speed - Change tokens to checksum'd case

timjrobinson:
# Description

This is a replacement for #3243 but making the tokens checksum'd instead of lowercase because it was a much simpler change as most addresses are in this case already.

The tokens list started to become very slow since adding many new prices from the new API because it's doing a O^2 search for token prices and balances looping over two arrays. You can see when filtering by tokens in beta it takes 2 - 5 seconds to load the list.

This O^2 search was done because we compare upper and lower case token addresses.

Instead of doing this, change all addresses to checksum'd case (using `getAddress`) by default and use a map to search for prices / balances. Most tokens are already in this case, we just need to ensure when fetching from external sources (such as fetching prices from the API) that the addresses are sanitized correctly.

This makes the token list open instantly again.

## Other Changes

- Added the new tokens list (the all tokens list) to `tokens` in `tokens.provider`, this is so that when tokens are injected from a pool they are found in the list and ignored. Without the new list the frontend thinks they're new tokens and tries to re-fetch info about them.
- Removed injecting the pool address into the token list - not sure why this was even in there in the first place, most pools don't have a price for their address so injected their token and waiting for a price just makes things much slower. 

The above two changes + the general improvements makes loading an individual pool page much faster because it's no longer injecting + waiting for tokens to be loaded, and all token searches through maps instead of arrays is much faster. 

## Type of change

## How should this be tested?

- Ensure prices and balances look correct on all pages
- Run locally and uncomment the code that errors on uppercase tokens to see if you encounter any errors.

## Checklist:


src/composables/queries/usePoolQuery.ts
```diff
@@ -82,7 +82,6 @@ export default function usePoolQuery(
     await injectTokens([
       ...tokensListExclBpt(pool),
       ...tokenTreeLeafs(pool.tokens),
-      pool.address,
```
timjrobinson, line 85: I went back in history and injecting the pool address has been in the code forever but I couldn't figure out why we'd need it, as most pools don't have a price so it's making it do unnecessary calls to coingecko and slowing things done.

src/lib/utils/index.ts
```diff
@@ -159,12 +159,7 @@ export function selectByAddress<T>(
   map: Record<string, T>,
   address: string
 ): T | undefined {
-  const foundAddress = Object.keys(map).find(itemAddress => {
-    if (isSameAddress(itemAddress, address)) {
-      return true;
-    }
-  });
-  if (foundAddress) return map[foundAddress];
```
garethfuller, line 167: The point of this function, and the other address helper functions, is that you can use them without worrying about the address casing. I understand why this is slow here, but maybe where the map is large we shouldn't be using this function rather than changing it?

timjrobinson: I wanted to ensure everything is fast, but I can see this function is used in a bunch of places where only a handful of tokens are in the map, so I've made a new fast version and am using that for the price/balance fetching stuff that goes over hundreds of tokens.

src/providers/tokens.provider.ts
```diff
@@ -249,10 +249,13 @@ export const tokensProvider = (
...
 
     const existingAddresses = Object.keys(tokens.value);
+    const existingAddressesMap = Object.fromEntries(
+      existingAddresses.map((address: string) => [getAddress(address), true])
+    );
 
     // Only inject tokens that aren't already in tokens
     const injectable = addresses.filter(
-      address => !includesAddress(existingAddresses, address)
+      address => existingAddressesMap[address] !== true
```
alter-eggo, line 258: prob just `!existingAddressesMap[address]` would be better?
I get a little bit of brain lag when see`!== true` in filter callback 😁

alter-eggo (approved):
lgtm, and yeah, seems pool page loading is much faster now. great work @timjrobinson 💪

src/providers/tokens.provider.ts
```diff
@@ -411,9 +417,10 @@ export const tokensProvider = (
   /**
    * Get single token from state
    */
-  function getToken(address: string): TokenInfo {
+  function getToken(address: string): TokenInfo | undefined {
```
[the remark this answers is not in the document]
timjrobinson: I only use void when a function is supposed to be pure and never return anything, so don't think it makes sense here.

agualis:
@timjrobinson I fixed some tests that were affected by the speed change. There are now some test warnings that I will fix in another PR.

PR #3265, opened 2023-05-02
Hotfix: Fix swap page crashing with invalid assets

timjrobinson:
# Description

If an invalid asset or address is passed to the swap page it errors and stops working. This also caused the page to become permanently unavailable if a bad asset gets stored in  local storage.  

This PR fixes that bug ensuring that the swap assets are valid token addresses.

Example URL that currently crashes the swap page: http://app.balancer.fi/#/polygon/swap/eth/matic

Fixes: https://balancer-labs.sentry.io/issues/4147784240/?project=5725878

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- Try adding different words or tokens to the URL of the swap page, or selecting bad assets, and ensure it doesn't throw errors in the console or fail to load. 

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer/frontend-v2",
-  "version": "1.98.0",
+  "version": "1.98.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@balancer/frontend-v2",
-      "version": "1.98.0",
+      "version": "1.98.1",
       "license": "MIT",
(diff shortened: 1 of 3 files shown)


agualis (approved):
Good catch!

Outside the scope ot this bugfix: we have ~200 lines of complex swap logic in this component. I would extract them to a composable so that we can easily unit test all the edge cases like this one.

garethfuller (approved):

PR #3296, opened 2023-05-05
add BaoUSD-LUSD StablePool mainnet

fabiaz84:
Description

Type of change
[ X ] Other; add stablepool on mainnet


garethfuller:
Replacing with #3313 so I could bump the version.

PR #3326, opened 2023-05-10
Hotfix: Fix useSignatures user setting

garethfuller:
# Description

Currently, the useSignature user setting is being set as a string but treated as a boolean, we need to add some parsing to ensure it's set as a boolean in state. Also, in the join flow, because approval actions are set with a function call, we need to trigger a recall of that function if the userSignature setting changes.

## Type of change

## How should this be tested?

- [ ] Test that toggling the "Use signatures" setting in the join flow changes the preview flow back and forth between signatures and transactions for approval.

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer/frontend-v2",
-  "version": "1.100.9",
+  "version": "1.100.10",
   "lockfileVersion": 2,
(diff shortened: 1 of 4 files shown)


alter-eggo (approved):

timjrobinson (approved):
Works for me.

PR #3338, opened 2023-05-11
Add original error info for Sentry

timjrobinson:
# Description

In some Sentry error logs (e.g. https://balancer-labs.sentry.io/issues/4173315674/?project=5725878) there is a `data.originalError` value (which I believe is the EVM error) that is not being parse correctly because it's too deep. We should surface this in the additional data. 

I also found https://github.com/enzoferey/ethers-error-parser which could be useful but I worry what happens if we attempt to parse a non ethers error with it. 

## Type of change

## How should this be tested?

- Create a transaction that fails (can do this via disconnecting your hardware device) and debug the code to ensure originalError is received and passed. 
- Ensure this does not interfere with other error types that don't have `data.originalError` (not sure how)

## Visual context

## Checklist:


src/services/web3/transactions/concerns/contract.concern.ts
```diff
@@ -131,6 +131,7 @@ export class ContractConcern extends TransactionConcern {
         action,
         sender,
         simulate,
+        originalError: error.data?.originalError,
```
garethfuller, line 134: Data is also optional, should it be:
```suggestion
        originalError: error?.data?.originalError,
```

garethfuller (changes requested):

PR #3359, opened 2023-05-12
feat: pool risks section in pool page

agualis:
# Description

Adds risks section to pool detail page:

- It contains the specific pool risk logic but we won't use it for now ([see comment](https://github.com/balancer/frontend-v2/pull/3359#issuecomment-1546910891) in this PR).
- It does not use i18n to simplify the logic (reduce bundle size) until we decide to translate (given that the original `Risks.vue` was not translated neither).
- We could extract hash IDs from `Risk.vue` to enforce hash correctness but we avoided it as we won't probably change them too much. 

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

https://github.com/balancer/frontend-v2/assets/1316240/8ceaa953-c01a-4ce5-9896-68a84e3fe948

## Checklist:


agualis:
@pkattera Do we already have specific risks for a some concrete pool?

We should probably add then in a new pool metadata field @gareth?

pkattera:
@agualis We do not have any specific risks for any pools as far as I know. I thought it may be a good idea to design for this possibility which may arise. We could just start with the category risks and add the ability to add specific risks later if the need arises if it is difficult to set up.

src/pages/pool/_id.vue
```diff
@@ -268,6 +268,7 @@ watch(
           <template v-if="isSentinelIntersected && pool">
             <PoolTransactionsCard :pool="pool" :loading="loadingPool" />
             <PoolContractDetails :pool="pool" />
+            <PoolRisks :pool="pool"></PoolRisks>
```
garethfuller, line 271:
```suggestion
            <PoolRisks :pool="pool" />
```

garethfuller (changes requested):
Because the pool risk page is so large, should we avoid using smooth scroll here? It feels a bit clumsy to click a link to a section on a page and have to wait a non-trivial amount of time for the page to scroll to the right position. wdyt @pkattera?

garethfuller:
@pkattera I think we should get this out as is, dependent on review updates above. However, would be nice to make the risk section a little more engaging. @pkattera wdyt about having a bordered box with a relevant icon for each type of risk?

pkattera:
> Because the pool risk page is so large, should we avoid using smooth scroll here? It feels a bit clumsy to click a link to a section on a page and have to wait a non-trivial amount of time for the page to scroll to the right position. wdyt @pkattera?

I agree that smooth scroll feels clunky when arriving from outside of the page (e.g. clicking from a pool page onto the risks page and then scrolling downwards). Ideally, the user would arrive directly at the anchor when they come from a different page. 

I do like the smooth scroll when clicking anchor links within the risks page. 

Is there an easy fix to allow both behaviors?

garethfuller (changes requested):
Doesn't work on first load for me: https://www.loom.com/share/5bde7fb3128f4b65a48946dd3b6f4afc

agualis:
> Doesn't work on first load for me: https://www.loom.com/share/5bde7fb3128f4b65a48946dd3b6f4afc

@ It should work now that I had a delay to wait for the route transition.

garethfuller (approved):

PR #3388, opened 2023-05-18
Fix: avoid "unable to fetch token prices" errors in pool page

agualis:
# Description

Before this change, we were getting many `Unable to fetch token prices` errors when calling coingecko in `useHistoricalPricesQuery`. This means lots of sentry errors and annoying console.errors.

https://github.com/balancer/frontend-v2/assets/1316240/74e1c442-2eeb-4d37-a55b-6abf4667d310

The problem was that we were trying to get prices for "pool tokens" such as ` bb-a-USDT` or `bb-ag-WETH`, that are not defined in coingecko. 

The solution is excluding the addresses of those "pool tokens" from the query.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:


garethfuller (changes requested):
I believe this might mess up badly where these historical prices are actually used. See here: https://github.com/balancer/frontend-v2/blob/develop/src/components/contextual/pages/pool/PoolChart.vue#L187-L191

There is something about changing the amounts array based on the preMintedBptIndex so array lengths matter. I believe these prices are only used as a fallback but I'm not sure. @alter-eggo could you review this and let us know if it might break things? Can we remove the dependency on historical prices?

garethfuller:
@alter-eggo could you review this please?

alter-eggo:
historical prices are used to calculate properly pool's TVL. @markusbkoch wonder if we can remove this calculation and just use `snapshot.liquidity` to draw TVL chart?

markusbkoch:
> historical prices are used to calculate properly pool's TVL. @markusbkoch wonder if we can remove this calculation and just use `snapshot.liquidity` to draw TVL chart?

Yes. There've been a few improvements to TVL computation on the subgraph since we first implemented this logic that computes TVL based on coingecko prices. So I think we can give `snapshot.liquidity` a shot. Eventually maybe the coingecko logic can be moved into the API if accuracy turns out to be an issue again?

garethfuller (approved):
I think this is ok to merge for now then. I'll add a ticket to Asana to remove the Coingecko dependency in the charts.

alter-eggo (approved):

PR #3389, opened 2023-05-18
Hotfix: BPT balance fetching

garethfuller:
# Description

We need to inject pool addresses into the token registry so we include it in balance fetching. This is how we know if a user has an unstaked balance in a pool.

## Type of change

## How should this be tested?

- [ ] Test a user balance is displayed on the pool page when you refresh the page. If you try this on production it will always be zero.

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer/frontend-v2",
-  "version": "1.101.4",
+  "version": "1.101.5",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


timjrobinson (approved):
LGTM. Was removed in https://github.com/balancer/frontend-v2/pull/3258/files#diff-c3b5456416ba261147783ebe53f3998fa1c980e67040a7aa3850b373d105438a as I had no idea why it was there. 

This may cause failed coingecko requests but that may be fixed by https://github.com/balancer/frontend-v2/pull/3388

PR #3399, opened 2023-05-19
chore: Ensure only proportional recovery exits are available for ComposableStableV1

garethfuller:
# Description

We were showing the option to single asset withdraw from shallow ComposableStableV1 pools. This works if you withdraw everything, but not if you try to withdraw a specific amount, because the pool type doesn't support the required exit type. This PR effectively enforces proportional recovery withdrawals for ComposableStableV1 pools.

## Type of change

## How should this be tested?

- [ ] Test withdrawals from this pool polygon pool `0x216690738aac4aa0c4770253ca26a28f0115c595000000000000000000000b2c` you may need to unblock add-liquidity locally in order to test withdrawals.

## Checklist:

### DIFF
--- a/src/components/contextual/pages/pool/withdraw/WithdrawPage.vue
+++ b/src/components/contextual/pages/pool/withdraw/WithdrawPage.vue
@@ -8,6 +8,7 @@ import useWithdrawPageTabs from '@/composables/pools/useWithdrawPageTabs';
 import WithdrawPageTabs from './WithdrawPageTabs.vue';
 import { provideExitPool } from '@/providers/local/exit-pool.provider';
 import { Pool } from '@/services/pool/types';
+import { isComposableStableV1, isDeep } from '@/composables/usePoolHelpers';
 
 type Props = {
   pool: Pool;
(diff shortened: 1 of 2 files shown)


src/components/contextual/pages/pool/withdraw/WithdrawPage.vue
```diff
@@ -32,6 +33,18 @@ const { network } = configService;
 const { resetTabs } = useWithdrawPageTabs();
 
 onMounted(() => resetTabs());
+
```
agualis, line 36: You could extract this to a composable and add some unit tests.

agualis (approved):
Approved with testing suggestion.

src/components/contextual/pages/pool/withdraw/WithdrawPage.vue
agualis: No sorry,  I mean the `singleAssetExitsAvailable` computed. 

Something like: 

```ts
export function useWithDraw(poo:Pool) {
...
returns { singleAssetExitsAvailable };
}
```

garethfuller:
I've changed my mind about this change. Max single-asset exits are possible so we should leave this open. The majority of cases will be withdrawing 100% so this PR was solving a small edge case but limiting the primary use of single asset withdrawals.

PR #3418, opened 2023-05-23
Move token addresses to the tokens config file

timjrobinson:
# Description

Moves all the token addresses from the contracts file to the tokens file. This is a more correct place for them and they aren't statically defined so the list can be custom to the network instead of having so many empty strings on networks that don't support LSD tokens. 

I also changed a few places referencing "Weth" when really they meant "Wrapped Native Asset". 

## Type of change

- [x] Code refactor / cleanup

## How should this be tested?

- Ensure places involving stETH / stMATIC work as expected. 
- Ensure SOR and other things that might be affected by this work correctly. 

## Checklist:


garethfuller (changes requested):
Thanks for doing this, looks good but a few type/test errors to resolve.

garethfuller (commented):
Tests failing.

agualis:
> Tests failing.

@timjrobinson 

It was the lido mock. I pushed a commit to fix it. We should replace this untyped vi.mock by a better solution if it fails again.

garethfuller (approved):

PR #3426, opened 2023-05-23
Bao pools new

fabiaz84:
Description

Type of change
[ X ] Other; add stablepool and weighted pool on mainnet

Weighted pool
[Bao / baoETH-ETH (50/50)](https://app.balancer.fi/#/ethereum/pool/0x3b9fb87f7d081ceddb1289258fa5660d955317b6000200000000000000000544)

Stable pool
[bETH/baoETH-ETH](https://app.balancer.fi/#/ethereum/pool/0x505a889b08b4295ef5412c7b2f9ef8c6fc60e53300000000000000000000054d)


solarcurvey:
I think you need to use a rate provider for baoETH, if it's an LST like wstETH. can see our docs here https://docs.balancer.fi/reference/contracts/rate-providers.html#rate-providers

but also please feel free to reach out to me on telegram @solarcurve or discord to discuss further. for now we can't approve this since I believe the pools are made incorrectly

PR #3449, opened 2023-05-26
Hotfix - Dont allow joining pools with native asset

timjrobinson:
# Description

Joining a pool with a native asset is currently broken. It always gives the error "Unable to construct join". This PR temporarily disables joining with a native asset until this is fixed.

## Type of change

## How should this be tested?

- Try joining a pool that contains WETH/WMATIC
- You shouldn't be able to select ETH or MATIC to join with. 

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer/frontend-v2",
-  "version": "1.102.15",
+  "version": "1.102.16",
   "lockfileVersion": 2,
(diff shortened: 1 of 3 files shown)


garethfuller (changes requested):
This is only relevant for deep pools that are using the generalizedJoin function in the SDK. Not all pools. =

garethfuller (approved):

PR #3480, opened 2023-06-01
Re-add bump tag

timjrobinson:
This fixes and re-adds the BumpTag action removed in #3477 

It failed because the master branch is protected. I changed it to use a personal access token that I created to get around this restriction. The other changes are making it checkout without default Github credentials, and re-checkout at the end to push to the deploy branch. 

The permissions on the personal access token are:

Screenshot from 2023-06-01 16-49-01

I'm not sure if all of these are necessary. We could create a balancer bot with these permissions instead of using my token too. 

Tested in my personal fork by protecting my master branch in the same way this repo is protected, then making sure this worked.

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -1,5 +1,7 @@
+# Bumps the package version with a patch if the version wasn't already updated
+# - This is skipped if there is already a version bump in the merged PR
 # Tags each push to master with version in package.json if not already a tag.
-name: Tag
+name: BumpTag
 
 on:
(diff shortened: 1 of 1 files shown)


agualis:
@timjrobinson Version workflow is failing. 

Ideally we should use a general purpose organization PAT for this but we can add it later.

timjrobinson:
@agualis Yea the version isn't needed for this, it will bump itself after its merged

agualis (approved):

PR #3486, opened 2023-06-01
Hotfix: Add missing + to deploy script check. 

timjrobinson:
There was a plus missing which made pushing to deploy branch break when the patch version is double digits, which just happened.


timjrobinson:
Merged this fix into #3485

PR #3513, opened 2023-06-09
refactor: avoid capturing queryExit errors until the last retry attempt

agualis:
# Description

Goal: we want to reduce false positives in sentry issues

In the exit flow, we use a watcher to repeatedly call `queryExit` to simulate the exit transaction from the "preview modal". 
If some of this queries goes wrong (due to different causes) vue-query will retry 3 times. 
We have some random errors that happen just once but they don't in the second/third retry so they are not actually affecting the user, however we currently log this errors increasing fatal-error/transaction ratio in sentry. 

I reproduced this scenario in the generalized exit flow, where, I get random 'out-of-bounds' exceptions during queryExit simulations that do not happen after the first retry. 

Solution: In the simulation queries, instead of calling `logExitException` after the first error, we wait for the 3 vue-query retries to be attempted, so that, we will avoid this sentry issues when they don't actually affect the user flow.

It also renames `queryExit` to `simulateExit` because we had `queryExit` function in 4 layers (provider, service, handler and sdk) so it can be a bit difficult to reason about the flow and error handling for a new developer. Now `simulateExit` function shows the upper level where we handle the errors. 

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:

### DIFF
--- a/src/providers/local/exit-pool.provider.ts
+++ b/src/providers/local/exit-pool.provider.ts
@@ -1,4 +1,3 @@
-import useRelayerApprovalTx from '@/composables/approvals/useRelayerApprovalTx';
 import useNumbers from '@/composables/useNumbers';
 import {
   fiatValueOf,
@@ -39,19 +38,6 @@ import useWeb3 from '@/services/web3/useWeb3';
 import { TokenInfoMap } from '@/types/TokenList';
 import { TransactionActionInfo } from '@/types/transactions';
 import { TransactionResponse } from '@ethersproject/abstract-provider';
-import {
-  computed,
-  InjectionKey,
-  onBeforeMount,
-  provide,
-  ref,
-  reactive,
-  toRef,
-  watch,
-  onMounted,
-  Ref,
-  readonly,
-} from 'vue';
 import { useQuery } from '@tanstack/vue-query';
 import debounce from 'debounce-promise';
 import { captureException } from '@sentry/browser';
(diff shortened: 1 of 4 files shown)


src/providers/local/exit-pool.provider.ts
```diff
@@ -385,7 +397,7 @@ export const exitPoolProvider = (
       isTxPayloadReady.value = output.txReady;
       return output;
     } catch (error) {
-      logExitException(error as Error);
+      // We will log this error (via logExitException) only after the last retry attempt (inside logOnlyAfterLastRetryAttempt)
```
garethfuller, line 400: The problem here may be that you don't get the root error sent to Sentry, you get the error thrown below.

src/providers/local/exit-pool.provider.ts
```diff
@@ -348,7 +360,7 @@ export const exitPoolProvider = (
   /**
    * Simulate exit transaction to get expected output and calculate price impact.
    */
-  async function queryExit() {
+  async function simulateExit() {
```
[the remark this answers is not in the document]
agualis: Ok, I see your point.

src/providers/local/exit-pool.provider.ts
agualis: Arrggg, true. I see that vue-query also exposes `failureCount`. Probably that's the way to go.

agualis:
Closed by #3514

PR #3525, opened 2023-06-13
Make veBAL gauge filter run via config

timjrobinson:
# Description

- Refactor the filter list on the veBAL gauge page to use config settings so that it automatically adds new networks that are not test networks and have pools users can vote for.
- This will show zkEVM once we add gauges people can vote for to the list. 

## Type of change

- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- Load the veBAL page
- Attempt to filter by a network
- See same networks are present as before with same names

## Checklist:


alter-eggo (approved):
lgtm (except type error)

PR #3595, opened 2023-06-30
Fix allowance validation when staking in gauges

timjrobinson:
This adds the post allowance validation from #3515 back into BalActionSteps. 

It had a bug previously where no one was able to stake their BPT because we didn't think they had allowed it to be spent. This happened because the allowances query only checks spenders of the vault, wstETH and veBAL. It doesn't check if a gauge can spend a users BPT. 

The bug was fixed by allowing injecting of new spenders, and injecting the gauge into the spenders list before the user stakes their BPT. 

I renamed `allowanceContracts` to `spenders` as that's how the ERC20 standard refers to them, it made it less confusing to understand, and they might not be contracts (you can allow normal EOA's to be a spender of your tokens). 

This also fixes a minor bug we didn't notice before: Users had to re-allow their BPT every time they staked previously, now they don't have to do that. 

This could be made faster by not re-fetching all allowances after the injection and only fetching the allowances of that gauge, but that could come in a future PR. 

## Testing

- Invest in a pool 
- Try and stake those BPT. It should work correctly. 
- Unstake and restake the BPT, if you allowed max previously it should not ask for an approval again. 
- Check normal swaps allowances work correctly.

### DIFF
--- a/src/composables/approvals/useTokenApprovalActions.ts
+++ b/src/composables/approvals/useTokenApprovalActions.ts
@@ -37,13 +37,8 @@ export default function useTokenApprovalActions() {
   /**
    * COMPOSABLES
    */
-  const {
-    refetchAllowances,
-    approvalsRequired,
-    approvalRequired,
-    getToken,
-    injectSpenders,
-  } = useTokens();
+  const { approvalsRequired, approvalRequired, getToken, injectSpenders } =
+    useTokens();
   const { t } = useI18n();
   const { getSigner } = useWeb3();
   const { addTransaction } = useTransactions();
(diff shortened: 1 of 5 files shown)


src/composables/approvals/useTokenApprovalActions.ts
```diff
@@ -89,6 +94,7 @@ export default function useTokenApprovalActions() {
     amountsToApprove: AmountToApprove[],
     spender: string
   ): Promise<AmountToApprove[]> {
+    await injectSpenders([spender]);
```
garethfuller, line 97: Since `injectSpenders` triggers the allowance query and waits for it, could we not remove the `refetchAllowances` call after it? If we do, I would add a comment above saying that this is what we expect to happen, in case someone tries to remove that from the `injectSpenders` function in the future.

timjrobinson: True, will fix.

PR #3631, opened 2023-07-05
fix: Missing tokens join call

garethfuller:
# Description

Might fix: https://balancer-labs.sentry.io/share/issue/4078ca967f574832828e73caf59b44b4/

- Ensures setAmountsIn is never called with an empty array on mounting components.
- Adds extra guard for join query after approval action promise.

## Type of change

## How should this be tested?

It's difficult to replicate this issue. I only replicated by keeping the add liquidity page open for a long time, leaving it un focused and then returning. My guess is that some kind of re-render caused the setAmountsIn to be reset as empty during a call to query the join.

## Checklist:

### DIFF
--- a/src/components/forms/pool_actions/AddLiquidityForm/AddLiquidityForm.vue
+++ b/src/components/forms/pool_actions/AddLiquidityForm/AddLiquidityForm.vue
@@ -61,8 +61,7 @@ const {
   hasValidInputs,
   hasAmountsIn,
   queryError,
-  setAmountsIn,
-  addTokensIn,
+  setTokensIn,
 } = useJoinPool();
 
 const { tokensWithBalance } = useUserTokens();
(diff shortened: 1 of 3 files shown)


src/components/forms/pool_actions/AddLiquidityForm/AddLiquidityForm.vue
```diff
@@ -101,14 +100,13 @@ const joinTokensWithoutBalance = computed<string[]>(() =>
 );
 
 async function initializeTokensForm(isSingleAssetJoin: boolean) {
-  setAmountsIn([]);
```
garethfuller, line 104: I think this might have caused it, but can't see how that would also get past the existing guard in the query function.

src/providers/local/join-pool.provider.ts
```diff
@@ -284,22 +284,31 @@ export const joinPoolProvider = (
...
-      return null;
-    }
+    if (!validateAmountsIn()) return null;
 
     try {
       joinPoolService.setJoinHandler(joinHandlerType.value);
       await setApprovalActions();
 
       console.log('joinHandler:', joinHandlerType.value);
+      if (!validateAmountsIn()) return null;
```
garethfuller, line 311: So I've added this additional guard right before the query call, it's possible that the amounts were being reset to empty whilst we waited on the `setApprovalActions` function.

agualis (approved):

PR #3642, opened 2023-07-06
refactor: Fetch allowances for vault only once in join flow

garethfuller:
# Description

Previously we were refetching vault allowances everytime the the user changed amounts in the join form, this was a side-effect of re-constructing approval actions required in the preview modal. This PR makes that allowance fetching optional for the get actions function and adds a separate function to ensure allowances are updated for a given spender. Then in the join pool provider, onBeforeMount, we just fetch allowances for the vault and skip the fetching for reconstructing the approval actions as token amounts are changed.

## Type of change

- [x] Code refactor / cleanup

## How should this be tested?

- [ ] Test approvals work in join flows and there is no longer excessive allowance fetching.

## Checklist:

### DIFF
--- a/src/composables/approvals/useTokenApprovalActions.ts
+++ b/src/composables/approvals/useTokenApprovalActions.ts
@@ -23,6 +23,7 @@ interface Params {
   spender: string;
   actionType: ApprovalAction;
   forceMax?: boolean;
+  skipAllowanceCheck?: boolean;
 }
 
 interface ApproveTokenParams {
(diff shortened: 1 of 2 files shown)


src/composables/approvals/useTokenApprovalActions.ts
```diff
@@ -90,12 +91,19 @@ export default function useTokenApprovalActions() {
     }
   }
 
+  async function updateAllowancesFor(spender: string): Promise<void> {
+    await injectSpenders([spender]);
+    await refetchAllowances();
```
alter-eggo, line 96: cannot be in Promise.all?

alter-eggo (approved):

src/composables/approvals/useTokenApprovalActions.ts
alter-eggo: ah, seems not

PR #3676, opened 2023-07-10
fix: zkevm tx confirmation callback

garethfuller:
# Description

It's possible that zkevm rpc state mismatches/caching are causing problems with submitting transactions before approvals are finalized. Like we had already done for Polygon PoS, this PR introduces a 10 confirmation wait before we consider a tx finalized on zkevm.

Additionally, txListener was being used at two levels in several flows, we only need to listen in BalActionSteps and then emit the events to be handled at a higher level. This has been refactored where relevant.

## Type of change

## How should this be tested?

- [ ] Test zkevm transactions

## Checklist:

### DIFF
--- a/src/components/_global/BalActionSteps/BalActionSteps.vue
+++ b/src/components/_global/BalActionSteps/BalActionSteps.vue
@@ -51,7 +51,8 @@ const props = withDefaults(defineProps<Props>(), {
 });
 
 const emit = defineEmits<{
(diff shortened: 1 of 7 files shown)


alter-eggo (approved):

timjrobinson (approved):

garethfuller:
There is some additional logic that needs to be fixed here.

timjrobinson (approved):

PR #3678, opened 2023-07-11
Remove default tokens list, change vetted to allowlisted

timjrobinson:
# Description

- Removed the old default tokens list as it's no longer used in the UI. In the future we may add a special list.
- Renamed the vetted list to allowlisted to be consistent with tokenlists repo and other places that we allow tokens. As vetted was ambiguous if it represents all allowlisted tokens or a hand picked "vetted" set of them.

This fixes the token list toggle in the token search modal, it now has the balancer allowlist permanently on and you can toggle other lists. Previously the vetted list was hidden and the listed-old toggle was displayed which didn't really do anything. 

### Future Considerations:

- Maybe we should move pool tokens to another list and allow them to be toggled on and off? As they look weird being in the normal swap list even though they are swappable tokens
- I'm working on generating a list of top 100 tokens by volume that we could have as the default list and then users can toggle on the full allowlist if they like. 

## Type of change

- [x] Code refactor / cleanup

## How should this be tested?

- Load the token list modal on different networks, ensure it has all tokens and the token list is permanently enabled. 

## Checklist:


src/composables/useDisabledJoinPool.ts
```diff
@@ -35,10 +35,10 @@ function doesRequireAllowListing(pool: Pool, account: string): boolean {
 
 export function useDisabledJoinPool(pool: Pool) {
   const { account } = useWeb3();
-  const { vettedTokenList } = useTokenLists();
+  const { allowlistedTokenList } = useTokenLists();
```
garethfuller, line 38: Maybe we could just name this `balancerTokenlist` instead? That's what it's called in the tokenlist repo.

src/lib/config/arbitrum/tokenlists.ts
```diff
@@ -2,9 +2,7 @@ import { TokenListURLMap } from '@/types/TokenList';
 
 const tokenlists: TokenListURLMap = {
   Balancer: {
-    Default:
-      'https://raw.githubusercontent.com/balancer/tokenlists/main/generated/listed-old.tokenlist.json',
-    Vetted:
+    Allowlisted:
```
garethfuller, line 5: Since we no longer have more than one, why do we need the nested attribute?

garethfuller (changes requested):
Looks like there is a type issue to resolve as well.

src/lib/config/arbitrum/tokenlists.ts
timjrobinson: I was thinking we'd want the top 100 tokens by volume here. Do we want that list generated on tokenlists update or in the frontend?

src/composables/useDisabledJoinPool.ts
timjrobinson: That makes sense

src/lib/config/arbitrum/tokenlists.ts
garethfuller: I don't think we know how we'll fetch that list yet, so maybe don't assume we'll list it here yet? Alternatively, it's fine to leave this like this for now just in case we do need it.

garethfuller (approved):

PR #3683, opened 2023-07-12
Use JSON RPC provider for initial block number (DM)

timjrobinson:
Original: #3682


timjrobinson:
@garethfuller It's complaining about the readme file?? Seems like a bug with the linter as the README hasn't changed. I merged develop into this branch to see if it fixes it.

PR #3687, opened 2023-07-13
Show Branded Redirect per Pool Type instead of individual pool ID's

timjrobinson:
# Description

There are a number of Xave / Gyro pools on Polygon that aren't showing the branded redirect banner because we forgot to add their ID's to the list. As these pools are always of a certain type it's easier to make the banner show based on pool type rather than ID's so they all show the correct banner.

## Type of change

## How should this be tested?

Visit some of the Xave and Gyroscope pools and ensure the branded banner shows. Some are:
- Polygon: 0xfa9ee04a5545d8e0a26b30f5ca5cbecd75ea645f000200000000000000000798
- Polygon: 0x7ec3e47ca85602351ee252c0bfae14e1e2c90bd500020000000000000000077f
- Polygon: 0xdac42eeb17758daa38caf9a3540c808247527ae3000200000000000000000a2b

Visit other pools and ensure the branded banner doesn't show. 

## Checklist:


garethfuller (changes requested):
Looks like there are some issues with tests/typecheck.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@balancer/frontend-v2",
-  "version": "1.112.8",
+  "version": "1.112.9",
```
garethfuller, line 3: Why is the version bumped?

timjrobinson: oh whoops I merged deploy into this branch instead of develop -_-

timjrobinson: Will fix

garethfuller (approved):

PR #3697, opened 2023-07-14
Add extra action step to unapprove tokens that require it

timjrobinson:
# Description

Some tokens such as USDT and Enjin Coin don't allow changing the token approval amount without first setting it to 0. This is at the top of their approve function:

```
 /**
    * @dev Approve the passed address to spend the specified amount of tokens on behalf of msg.sender.
    * @param _spender The address which will spend the funds.
    * @param _value The amount of tokens to be spent.
    */
    function approve(address _spender, uint _value) public onlyPayloadSize(2 * 32) {

        // To change the approve amount you first have to reduce the addresses`
        //  allowance to zero by calling `approve(_spender, 0)` if it is not
        //  already 0 to mitigate the race condition described here:
        //  https://github.com/ethereum/EIPs/issues/20#issuecomment-263524729
        require(!((_value != 0) && (allowed[msg.sender][_spender] != 0)));
```

This means if a user is updating their allowance in the normal flow while they already have an allowance, their transaction will fail. 

This PR adds an additional actionstep to set the allowance to 0 before setting the normal allowance if the user already has an existing allowance. 

Fixes https://balancer-labs.sentry.io/issues/4312899573/events/6f69b166f8de4732983309e24799a1cc/

## Type of change

## How should this be tested?

I haven't tested this yet. Ideally It'd be good if we made our USDT on Goerli function the same as this so we can test on there. I haven't found any tokens like this on Polygon or cheaper networks than mainnet. 

## Checklist:


src/composables/approvals/useTokenApprovalActions.ts
```diff
@@ -198,29 +201,64 @@ export default function useTokenApprovalActions() {
...
+            }),
+          postActionValidation: () => isApprovalValid(amountToApprove, spender),
+          actionInvalidReason: {
+            title: t('actionSteps.approve.invalidReason.title'),
+            description: t('actionSteps.approve.invalidReason.description'),
+          },
+        });
+        return actions;
+      })
+    );
```
garethfuller, line 261: Could you clean this up so it's easier to parse, something like:
```suggestion
    return flatten(
      approvalsRequired.map(amountToApprove => {
        const token = getToken(amountToApprove.address);
        const actions: TransactionActionInfo[] = [];

        if (isDoubleApprovalRequired(token.address, spender)) {
          actions.push(zeroApprovalAction(...params));
        }

        actions.push(approvalAction(...params));
     
        return actions;
      })
    );
```
You'll need to extract the action construction into function etc.

garethfuller (changes requested):

src/composables/approvals/useTokenApprovalActions.ts
timjrobinson: Fixed

garethfuller (changes requested):
Screenshot 2023-07-18 at 10 46 51
Testing this out on polygon (I added USDT to the list), I end up with just the unapprove action followed by the swap action, for some reason it fails to include the second approval action after the unapprove.

timjrobinson:
@garethfuller fixed. Also committed your suggestion.

PR #3718, opened 2023-07-17
Hotfix: Re-trigger exit queries on wallet change

garethfuller:
# Description

We should re-trigger exit queries, in particular, the single exit max query, when the user's wallet changes.

## Type of change

## How should this be tested?

- [ ] Switch wallets in the single asset exit flow and check if the max amount to withdraw updates.

## Checklist:

### DIFF
--- a/src/providers/local/exit-pool.provider.ts
+++ b/src/providers/local/exit-pool.provider.ts
@@ -97,17 +97,14 @@ export const exitPoolProvider = (
   const { txState, txInProgress } = useTxState();
   const { transactionDeadline } = useApp();
   const { slippageBsp } = useUserSettings();
-  const { getSigner } = useWeb3();
+  const { account, getSigner } = useWeb3();
   const { relayerSignature, relayerApprovalAction, relayerApprovalTx } =
     useRelayerApproval(RelayerType.BATCH);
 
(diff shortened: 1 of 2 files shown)


src/providers/local/exit-pool.provider.ts
```diff
@@ -97,17 +97,14 @@ export const exitPoolProvider = (
...
+  const { account, getSigner } = useWeb3();
   const { relayerSignature, relayerApprovalAction, relayerApprovalTx } =
     useRelayerApproval(RelayerType.BATCH);
 
   const debounceQueryExit = debounce(queryExit, debounceQueryExitMillis);
   const debounceGetSingleAssetMax = debounce(
     getSingleAssetMax,
-    debounceGetSingleAssetMaxMillis,
-    {
-      leading: true,
```
timjrobinson, line 109: Why was this needed before but isn't needed now?

timjrobinson (approved):

src/providers/local/exit-pool.provider.ts
garethfuller: I couldn't figure that out, I removed it to see if it would fix the problem. Let me double-check if this needs to be removed.

garethfuller: Yeah, this wasn't related to the issue, so I've reverted.

PR #3742, opened 2023-07-18
hotfix: Don't query exit if invalid amounts out

garethfuller:
# Description

We were allowing exit queries for invalid amounts out like 0 or amounts greater than the user's balance and therefore greater than the pool balance for that token. This allowed failed to construct exit errors. This PR prevents the queryExit function from running if single asset exits have invalid amounts and provides a mechanism for setting the price impact to `-` instead of 0 unless a value is explicitly returned from a queryExit.

## Type of change

## How should this be tested?

- [ ] Try to single asset exit and input a value greater than your balance, no errors should be thrown in the console.

## Checklist:

### DIFF
--- a/src/components/forms/pool_actions/WithdrawForm/components/WithdrawTotals.vue
+++ b/src/components/forms/pool_actions/WithdrawForm/components/WithdrawTotals.vue
@@ -28,6 +28,7 @@ const priceImpactClasses = computed(() => ({
       <div class="data-table-number-col">
         <div class="flex items-center">
           <BalLoadingBlock v-if="isLoadingQuery" class="w-10 h-6" />
+          <span v-else-if="priceImpact === -1">-</span>
           <span v-else>{{ fNum(priceImpact, FNumFormats.percent) }}</span>
 
           <BalTooltip :text="$t('withdraw.tooltips.priceImpact')">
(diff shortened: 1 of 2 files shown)


src/components/forms/pool_actions/WithdrawForm/components/WithdrawTotals.vue
```diff
@@ -28,6 +28,7 @@ const priceImpactClasses = computed(() => ({
       <div class="data-table-number-col">
         <div class="flex items-center">
           <BalLoadingBlock v-if="isLoadingQuery" class="w-10 h-6" />
+          <span v-else-if="priceImpact === -1">-</span>
```
agualis, line 31: I would extract the -1 to a semantic constant (i.e. `undefinedPriceImpact`) to make this logic clearer.

agualis (changes requested):

timjrobinson (approved):

src/components/forms/pool_actions/WithdrawForm/components/WithdrawTotals.vue
garethfuller: Yeah fair enough. I was being lazy... 😄

agualis (approved):

PR #3743, opened 2023-07-18
Fix: use stack instead of captureStackTrace

agualis:
# Description

After multiple tests I found that using `error.stack` instead of `captureStackTrace` works better to keep the stack trace in our sentry errors. 

Tested with chrome, firefox, edge and safari with a real vite build.

This is the issue stacktrace where I manually triggered an ether's `division by zero` inside `contract.concern.ts`:

error-with-stack

It adds the original's error stack trace but excludes the `captureBalancerException` which will be common to all `SentryError's` so IMO is not worthy no be included. 

I would merge this PR and start checking if it groups issues better than the old system (I just tested it with some trivial examples). If not, maybe we can try to fine tune them in sentry setup. 

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

## Checklist:

### DIFF
--- a/src/lib/utils/errors.ts
+++ b/src/lib/utils/errors.ts
@@ -134,8 +134,8 @@ class SentryError extends Error {
   constructor(name: string, message: string, cause: Error | unknown) {
     super(message);
 
-    if (Error.captureStackTrace && cause instanceof Error) {
-      Error.captureStackTrace(cause, SentryError);
+    if (cause instanceof Error && cause.stack) {
+      this.stack = cause.stack;
     }
 
     this.name = name;


garethfuller:
> excludes the `captureBalancerException`

@agualis what do you mean by this?

garethfuller (approved):

agualis:
> > excludes the `captureBalancerException`
> 
> @agualis what do you mean by this?

I mean that the stack trace in sentry will not include the code in `errors.ts` where the exception is being prepared (captureBalancerException, constructError and so on...) which is good.

PR #3752, opened 2023-07-19
Hotfix: Block all access to withdraw page if BPT 0

garethfuller:
# Description

I believe some errors are due to users accessing the withdraw form when they have a 0 BPT balance.

- Before mounting withdraw page, check if user has BPT, if not redirect to pool page
- After a successful withdrawal, if the user tries to close the modal, check if they have any remaining BPT, if not rediect to pool page.

## Type of change

## Checklist:

### DIFF
--- a/src/composables/useEthers.ts
+++ b/src/composables/useEthers.ts
@@ -11,13 +11,12 @@ import {
 } from '@/lib/utils/promise';
 import { rpcProviderService } from '@/services/rpc-provider/rpc-provider.service';
 
-import useBlocknative from './useBlocknative';
 import { toJsTimestamp } from './useTime';
 import { useTokens } from '@/providers/tokens.provider';
 import useTransactions from './useTransactions';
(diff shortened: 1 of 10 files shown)


src/composables/useEthers.ts
```diff
@@ -106,9 +104,9 @@ export default function useEthers() {
       if (receipt != null) {
         finalizeTransaction(txHash, 'tx', receipt);
       }
-      callbacks.onTxConfirmed(receipt);
-      if (shouldRefetchBalances && !supportsBlocknative.value) {
```
garethfuller, line 110: `!supportsBlocknative` is a check to prevent refetching of the balances, we really need to do it here too to ensure we know the balance after the post-action delay we have on networks like Polygon. It's possible we can just remove blocknative in the future, and do all the fetching here instead.

timjrobinson (approved):

PR #3783, opened 2023-07-23
Added composable stablepool STARD-TUSD-wUSDR - Update pools.ts

dakotajakiser:
Added a StarSeeds Protocol Stablecoin Pool with STARD, Tetu USD BP, and wUSDR.


dakotajakiser:
Yes sir. Will do shortly.

dakotajakiser:
> @dakotajakiser This should also be added to the [deep list here](https://github.com/balancer/frontend-v2/blob/f069d1b986af64e45b2178870c3861ac79aee716/src/lib/config/polygon/pools.ts#L313) because it contains `bb-t-USD` balancer pool.

Sir. It is done.

PR #3788, opened 2023-07-24
fix: Selected token in add liquidity form

garethfuller:
# Description

Single asset joins in the add-liquidity flow are only ever available when the pool is a composable stable and you have the option to swap join with any tokens. The options passed into TokenInput are only relevant when it's the pool token input flow, where you want to give the user the option to toggle between native asset and wrapped. If you pass these options when it's the single asset flow, it messes with the logic and prevents the token icon from displaying if the default asset is the wrapped native asset.

## Type of change

## How should this be tested?

- [ ] On production go to the single asset join flow of the top pool on polygon, you should see the "Select token" button in blue, but the balance suggests the current selected token in WMATIC.
- [ ] In the preview it should render as expected, with WMATIC selected.

## Visual context

Before: 
Screenshot 2023-07-24 at 11 53 38

After:
Screenshot 2023-07-24 at 11 53 28

## Checklist:

### DIFF
--- a/src/components/forms/pool_actions/AddLiquidityForm/AddLiquidityForm.vue
+++ b/src/components/forms/pool_actions/AddLiquidityForm/AddLiquidityForm.vue
@@ -120,6 +120,8 @@ function getTokenInputLabel(address: string): string | undefined {
  * the native asset instead.
  */
 function tokenOptions(address: string): string[] {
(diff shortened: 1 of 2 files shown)


timjrobinson (approved):
Works for me

agualis (approved):
Tested in polygon 👍

PR #3811, opened 2023-07-27
feat: Adds pool filtering by type

garethfuller:
# Description

Adds pool filters by type.

## Type of change

## How should this be tested?

- [ ] Test pool filtering on the home page

## Visual context

Screenshot 2023-08-24 at 20 42 37

## Checklist:


garethfuller:
@timjrobinson it's still a bit buggy, if you try filtering by stable pool, then load more a few times, you get to a pool value of around $1.7M and then load more doesn't work anymore.

alter-eggo (approved):

timjrobinson:
I discovered a very strange bug on the portfolio page where when you didn't have a wallet connected, all pools were showing in the unstaked list. I traced it back to usePoolsQuery always runs even when it should be disabled, but previously it was running with the option `id { in: [] }` by default which made it always return no pools so we didn't notice it. 

I fixed it by checking if it's enabled in `queryFn` but I was surprised that vue query doesn't do this automatically.

timjrobinson (approved):
Fixed some small issues, otherwise working well

PR #3909, opened 2023-08-08
fix: Set lower min tx buffers on L2s (DM)

garethfuller:
Original #3906


agualis (approved):

agualis:
@garethfuller I will fix this conflicts

garethfuller:
There are still conflicts.

PR #3968, opened 2023-08-15
Hotfix: allow all tokens on testnet

timjrobinson:
# Description

We shouldn't show the "Tokens not on allowlist" message on testnets, so that other teams can quickly test pools with their own test tokens without waiting on us to add them to the tokenlist. 

## Type of change

## How should this be tested?

- Visit a sepolia pool like `0x2ea0b1475f2864de2c6de25fc93ed588b99a625700000000000000000000002a`. It should not give any warnings about missing tokens. 
- Pools with unverified tokens on normal networks should still show the warning. 

## Checklist:


agualis (approved):
@timjrobinson Instead of destructuring BoostedPoolMock constant (legacy thing), I imported `aBoostedPool` function, which creates a new object.

garethfuller (approved):

PR #3986, opened 2023-08-16
WIP: SDK errors

gmbronco:
Adding tags for SDK specific errors.


garethfuller (commented):
Looks ok. It doesn't affect the existing error throwing in the exit handlers does it?

src/services/balancer/pools/exits/handlers/exact-in-exit.handler.ts
```diff
@@ -167,4 +172,92 @@ export class ExactInExitHandler implements ExitPoolHandler {
...
+      captureBalancerException({
+        error,
+        context: {
+          tags: {
+            dependency: 'Balancer SDK',
+          },
+        },
+      });
+    }
+  }
```
garethfuller, line 230: I'm not sure we want to do this. We don't capture errors in the handlers, we throw and let the providers catch them so we can render an error for the user.

garethfuller (changes requested):

PR #4063, opened 2023-08-23
whitelists new Gnosis pools

daveai:
whitelist following pools:
wstETH-WETH/staBAL3
wstETH-WETH/GNO
wstETH-WETH/COW
wstETH-WETH/BAL/AURA


Tritium-VLK:
Discussed with Karpatkey.  They will relaunch these pools pairing with wstETH and maintain decent wstETH/ETH depth.

They have asked us not to whitelist these pools for now.

PR #4094, opened 2023-08-25
Fix/unique key voting list

agualis:
# Description

There was a rendering problem in the vebal voting list because `BalTable` uses `data.id` as row `key` by default but, in this concrete case, after migrating to the API, we can have 2 pools (with different gauges) with the same id, which leads to weird row duplications when vue re-renders rows based on `key` attribute.

This PR fixes the problem by allowing `BalTable` consumers to pass a prop to calculate the unique key.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Visual context

Before:

before

After:
after

## Checklist:

### DIFF
--- a/src/components/_global/BalTable/BalTable.vue
+++ b/src/components/_global/BalTable/BalTable.vue
@@ -30,6 +30,8 @@ const emit = defineEmits<{
 type Props = {
   columns: ColumnDefinition[];
   data: DataProp[];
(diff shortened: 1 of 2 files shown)


alter-eggo (approved):
👍

PR #4136, opened 2023-09-01
chore: minor optimisations

garethfuller:
# Description

Minor optimisations for pool list loading.
- Removes await for token injection (not sure if this actually makes a difference)
- Prevents token injection happening twice.

## Type of change

- [x] Code refactor / cleanup

## Checklist:

### DIFF
--- a/src/composables/queries/usePoolsQuery.ts
+++ b/src/composables/queries/usePoolsQuery.ts
@@ -27,7 +27,8 @@ type PoolsQueryResponse = {
 
 export default function usePoolsQuery(
   filterOptions: PoolFilterOptions,
-  options: UseInfiniteQueryOptions<PoolsQueryResponse> = { enabled: true }
+  options: UseInfiniteQueryOptions<PoolsQueryResponse> = { enabled: true },
+  shouldInjectTokens = true
 ) {
   /**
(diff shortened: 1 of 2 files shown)


src/composables/queries/usePoolsQuery.ts
```diff
@@ -77,14 +80,16 @@ export default function usePoolsQuery(
...
-        await injectTokens(tokens);
+        if (shouldInjectTokens) {
+          const tokens = flatten(
+            pools.map(pool => [
+              ...pool.tokensList,
+              ...tokenTreeLeafs(pool.tokens),
+              pool.address,
+            ])
+          );
+          await injectTokens(tokens);
```
alter-eggo, line 91: shouldn't `await` be removed?

alter-eggo (approved):

src/composables/queries/usePoolsQuery.ts
garethfuller: We don't really use the subgraph case anymore and we'll be completely removing it soon. But I think when we fetch pools from the subgraph we do have a need to await the injection. Could be wrong though. Don't think it really matters right now.

PR #4138, opened 2023-09-01
refactor: Fix dark mode execution time

garethfuller:
# Description

For some reason the useDarkMode composable has a significantly higher execution time than everything else. My first attempt to fix that was to move the composable into a global provider so it's only instantiated once. This doesn't seem to have fixed the problem.

Screenshot 2023-09-01 at 11 58 08

## Type of change

## How should this be tested?

- [ ] Run lighthouse on the preview and check the execution time stats shown in the image above.

## Checklist:


garethfuller:
@agualis I broke the tests here. Do I need to add a provider somewhere to the test setup?

garethfuller:
Seems this is more to do with SDK dependency than dark mode specifically.

PR #4158, opened 2023-09-04
hotfix: Fix zkEVM multicall issue

garethfuller:
# Description

Bumps SDK to include fix for zkEVM multicall issue.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?

- [ ] Test swaps and generalized joins/exits on zkevm

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have requested at least 1 review (If the PR is significant enough, use best judgement here)
- [x] I have commented my code where relevant, particularly in hard-to-understand areas
- [x] If package-lock.json has changes, it was intentional.
- [x] The base of this PR is `master` if hotfix, `develop` if not

PR #4162, opened 2023-09-05
hotfix: Fix zkEVM multicall issue with new key

timjrobinson:
This is the same as #4158 but with a new zkEVM key that may work differently to the current prod key.


garethfuller:
Closing since we won't be using this key

PR #4291, opened 2023-10-06
Hotfix swap input scaling

johngrantuk:
# Description

`getSwapInfo` function is using 18 decimals for scaling amount used in SOR/Swaps which causes issues when using something like USDC. Changed to use correct token decimals. This should also fixes token decimals for exact token out case which also seems to be broken.
I'm not 100% clear on how the swap logic is working on FE so should be double checked this doesn't lead to any knock-on effects.

## Type of change

- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## How should this be tested?

Check USDC > wstEth on mainnet with a gas fee: Should now return a proper result compared to production.
Try a BAL > weth on mainnet with a gas fee and an amount for weth (e.g. exact out). Should now return a proper result compared to production.

## Visual context

## Checklist:

- [X] I have requested at least 1 review (If the PR is significant enough, use best judgement here)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@balancer/frontend-v2",
-  "version": "1.127.9",
+  "version": "1.127.10",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@balancer/frontend-v2",
-      "version": "1.127.9",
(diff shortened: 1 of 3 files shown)


alter-eggo (approved):
tested and now works ok 👍

PR #5884, opened 2024-07-10
hf: vebal boost value on portfolio page

alter-eggo:
# Description
This pr fixes vebal boost value on portfolio page


garethfuller (commented):
But why is boost NaN in this case?

alter-eggo:
feel like bc vebal total supply is 0

garethfuller (approved):

PR #5897, opened 2024-07-30
chore: Pool page redirects to new UI

garethfuller:


agualis (approved):
Although we have "Try zen" banner on top of the page, someone could get confused when being redirected. I will tell Cosme Fulanito about this change.

garethfuller:
@agualis @alter-eggo I should have put this in the description, this is a prep PR, not going live right now. When it does, there won't be a pool list on v2 either, so no one will be able to click on pools. It just introduces the redirects for all the external links, linking to our v2 pool pages. We can't do this at the domain level because of the hash router.

PR #5899, opened 2024-09-02
fix: avoid edit votes reloading 

agualis:
Fixes [this edge case bug](https://discord.com/channels/638460494168064021/1279826315163205686/1279826320968388628) reported in discord.


garethfuller (approved):
🙌 thanks for fixing this, can't have been easy to find.

PR #5916, opened 2024-10-25
chore: Remove all dependency on old API

garethfuller:


garethfuller:
@alter-eggo could you takeover on this PR and check the console log errors to see if anything needs to be fixed. Only pages that are important are the vebal and claim pages. I think I have removed all dependency on the old API but there are still errors and I'm not sure what the cause is.

